PR #219, opened 2015-01-06
Improves vector display debug

stephomi:


cedricpinson:
Can you fix member variables to _xxxx at the same time ?

stephomi:
done

PR #233, opened 2015-01-29
Access to program._uniformsCache Map, using static strings. It fixes the...

jtorresfabra:
... optimized too many times warning in Chrome Profiler

### DIFF
--- a/sources/osg/RenderLeaf.js
+++ b/sources/osg/RenderLeaf.js
@@ -54,11 +54,11 @@ define( [
 
                 program = state.getLastProgramApplied();
 
-                modelViewUniform = program._uniformsCache[ state.modelViewMatrix.name ];
-                modelWorldUniform = program._uniformsCache[ state.modelWorldMatrix.name ];
-                viewUniform = program._uniformsCache[ state.viewMatrix.name ];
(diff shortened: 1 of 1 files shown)


jtorresfabra:
Draw time drops from 14-16 ms to 9-11 ms in performance example with the fix applied, it is also noticeable in chrome profiler with 51% of time spent in RenderLeaf.drawGeometry to 27.47%. That's all in my computer. It could be good if others could test this.

Kuranes:
Tested here, solid gain confirmed,  for instance "performance" cpu profile now first top only GL func instead of osgjs func. (gl uniform/drawElement instead of osgjs drawGeom/etc)
Very nicely spotted, good work !

cedricpinson:
Thank you very much for digging and finding this. great work.

cedricpinson:
Could you check / review this branch ? https://github.com/cedricpinson/osgjs/pull/235 this should do the same thing except that a specific function is built with only needed uniform call/update for each program

jtorresfabra:
Thanks, I was just profiling the DatabasePager and get into this. 

@cedricpinson I just tested your PR and it seems to work well. I'm adding a couple of comments...

PR #235, opened 2015-02-02
Jtorresfabra/fix optimizedtoomanytimes

cedricpinson:


sources/osg/RenderLeaf.js
```diff
@@ -4,6 +4,84 @@ define( [
...
+
+        this.apply = undefined;
+        this.tempMatrice = Matrix.create();
+        this.Matrix = Matrix;
+        this.generateMethod();
+    };
+
+    CacheUniformApply.prototype = {
+
+        generateMethod: function () {
```
jtorresfabra: Isn't too generic this function name?

sources/osg/RenderLeaf.js
```diff
@@ -42,60 +120,22 @@ define( [
...
-                program = state.getLastProgramApplied();
 
-                modelViewUniform = program._uniformsCache[ state.modelViewMatrix.name ];
-                modelWorldUniform = program._uniformsCache[ state.modelWorldMatrix.name ];
-                viewUniform = program._uniformsCache[ state.viewMatrix.name ];
-                projectionUniform = program._uniformsCache[ state.projectionMatrix.name ];
-                normalUniform = program._uniformsCache[ state.normalMatrix.name ];
+                var program = state.getLastProgramApplied();
+                var programInstanceID = program.getInstanceID();
+                var cache = state._programCommonUniformsCache;
```
jtorresfabra: Maybe we should use a getter to access to the cache.

sources/osg/State.js
```diff
@@ -55,6 +55,9 @@ define( [
         // texture manager is referenced here because it's associated with gl object
         // of the gl context intialized with State
         this._textureManager = new TextureManager();
+
+
+        this._programCommonUniformsCache = {};
```
jtorresfabra: Did you try using Map from es6? I don't know if it could improve anything, just wondering...

sources/osg/RenderLeaf.js
```diff
@@ -4,6 +4,84 @@ define( [
...
+        this._program = program;
+
+        this.modelViewUniform = program._uniformsCache[ state.modelViewMatrix.name ];
+        this.modelWorldUniform = program._uniformsCache[ state.modelWorldMatrix.name ];
+        this.viewUniform = program._uniformsCache[ state.viewMatrix.name ];
+        this.projectionUniform = program._uniformsCache[ state.projectionMatrix.name ];
+        this.normalUniform = program._uniformsCache[ state.normalMatrix.name ];
+
+        this.apply = undefined;
+        this.tempMatrice = Matrix.create();
```
jtorresfabra: Matrice is too french, it should be tempMatrix  ;)

sources/osg/State.js
cedricpinson: yeah, I tried Map but it seemed a bit slower

sources/osg/RenderLeaf.js
cedricpinson: agree

sources/osg/RenderLeaf.js
cedricpinson: yes, any idea ?

sources/osg/RenderLeaf.js
cedricpinson: completely :)

sources/osg/RenderLeaf.js
jtorresfabra: Maybe generateUniformsApplyMethods or generateApplyMethods?

sources/osg/State.js
stephomi: @jtorresfabra @cedricpinson  It looks like the es6shim vendors might override the native Map.
But indeed, we should use native Map when possible.

cedricpinson: retested but yeah in this context Map is a bit slower. But it's a special case where we dont have a lot of entry in the cache

PR #237, opened 2015-02-02
Feature/multifrustum

Kuranes:


sources/osg/CullSettings.js
```diff
@@ -9,12 +9,17 @@ define( [], function () {
         this.bbCornerFar = ( lookVector[ 0 ] >= 0 ? 1 : 0 ) | ( lookVector[ 1 ] >= 0 ? 2 : 0 ) | ( lookVector[ 2 ] >= 0 ? 4 : 0 );
         this.bbCornerNear = ( ~this.bbCornerFar ) & 7;
         /*jshint bitwise: true */
+        this._enableFrustumCulling = false;
+
+        this._settingsSource = this;
```
cedricpinson: what is this ?

sources/osg/CullSettings.js
```diff
@@ -27,8 +32,20 @@ define( [], function () {
...
+        },
+
+        setEnableFrustumCulling: function ( value ) {
+            this._enableFrustumCulling = value;
+        },
+        getEnableFrustumCulling: function () {
+            return this._enableFrustumCulling;
+        },
+
+        getSettingSource: function () {
```
cedricpinson: what is this ?

sources/osg/CullVisitor.js
```diff
@@ -438,23 +254,36 @@ define( [
...
-            this.pushViewport( camera.getViewport() );
-        }
 
         // save current state of the camera
         var previousZnear = this._computedNear;
         var previousZfar = this._computedFar;
+
+        // save cullSettings
+        // TODO Perf: why it's not a stack
+        // and is pollutin  GC ?
```
cedricpinson: Adds an issue, we could fix it later

sources/osg/Light.js
```diff
@@ -199,6 +199,43 @@ define( [
...
+            // formula playground
+            // https://www.desmos.com/calculator/nmnaud1hrw
+            var constant = this._attenuation[ 0 ];
+            var linear = this._attenuation[ 1 ] * dist;
+            var quadratic = this._attenuation[ 2 ] * dist * dist;
+            return ( 1.0 / ( constant + linear + quadratic ) );
+        },
+        // First step to light intersection
+        // for culling
+        checkForNearerLightFar: function ( currentFar, epsilon ) {
```
cedricpinson: I would prefer to have this code somewhere else

sources/osg/Matrix.js
```diff
@@ -358,21 +369,17 @@ define( [
             return r;
         },
 
-        makeLookAt: ( function () {
-            var f = [ 0.0, 0.0, 0.0 ];
+        makeLook: ( function () {
```
cedricpinson: could comment why the split and how it's different from previous implementation

sources/osg/Matrix.js
```diff
@@ -405,6 +412,39 @@ define( [
...
+                }
+
+
+                Vec3.sub( center, eye, f );
+                Vec3.normalize( f, f );
+                this.makeLook( eye, f, up, result );
+                return result;
+            };
+        } )(),
+        makeInfiniteOrtho: function ( left, right, bottom, top, zNear, zFar, result ) {
```
cedricpinson: how it is different of ortho ?

sources/osg/Matrix.js
```diff
@@ -651,12 +692,16 @@ define( [
...
             return function ( matrix, vector, result ) {
-                var d = 1.0 / ( matrix[ 3 ] * vector[ 0 ] + matrix[ 7 ] * vector[ 1 ] + matrix[ 11 ] * vector[ 2 ] + matrix[ 15 ] );
-
                 if ( result === undefined ) {
-                    Notify.warn( 'no matrix destination !' );
-                    result = Matrix.create();
+                    Notify.warn( 'no vec3 destination !' );
+                    result = Vec3.create();
                 }
+                if ( Matrix.isIdentity( matrix ) ) {
```
cedricpinson: could you bench this, not sure statistically it's better to test if a matrix is identity before doing the operation. I mean in js.

sources/osg/Matrix.js
```diff
@@ -681,8 +726,12 @@ define( [
             return function ( matrix, vector, result ) {
 
                 if ( result === undefined ) {
-                    Notify.warn( 'no matrix destination !' );
-                    result = Matrix.create();
+                    Notify.warn( 'no Vec4 destination !' );
+                    result = Vec4.create();
+                }
+                if ( Matrix.isIdentity( matrix ) ) {
```
cedricpinson: same thing here, bench it

sources/osg/Matrix.js
```diff
@@ -702,14 +751,70 @@ define( [
             };
         } )(),
 
+        // could be 6 mult
+        // http://dev.theomader.com/transform-bounding-boxes/
+        // https://github.com/erich666/GraphicsGems/blob/master/gems/TransBox.c
+        transformBoundingBox: ( function () {
```
cedricpinson: not sure this code should be here, could you check in osg ?

sources/osg/Matrix.js
```diff
@@ -702,14 +751,70 @@ define( [
...
 
             return function ( matrix, vector, result ) {
 
                 if ( result === undefined ) {
-                    Notify.warn( 'no matrix destination !' );
-                    result = Matrix.create();
+                    Notify.warn( 'no Vec4 destination !' );
+                    result = Vec4.create();
+                }
+                if ( Matrix.isIdentity( matrix ) ) {
```
cedricpinson: bench

sources/osg/Matrix.js
```diff
@@ -757,6 +862,9 @@ define( [
 
             return function ( matrix, result ) {
                 if ( result === matrix ) {
+                    if ( Matrix.isIdentity( matrix ) ) {
```
cedricpinson: bench

sources/osg/Matrix.js
```diff
@@ -1170,8 +1341,94 @@ define( [
             };
         } )(),
 
-        // compute the 4 corners vector of the frustrum
-        computeFrustrumCornersVectors: function ( projectionMatrix, vectorsArray ) {
+        clampProjectionMatrix: function ( projection, znear, zfar, nearFarRatio, resultNearFar ) {
```
cedricpinson: Did you move the code from CullVisitor here because you use it in another context than CullVisitor ? Could you show me where ?

sources/osg/Matrix.js
```diff
@@ -1184,6 +1441,29 @@ define( [
             return vectorsArray;
         },
 
+        // better precison
+        // no far clipping artifacts.
+        // no reason not to use.
+        // Tightening the Precision of Perspective Rendering
+        //http://www.geometry.caltech.edu/pubs/UD12.pdf
+        // drop-in, just remove the one below, and rename this one
+        makeInfiniteFrustum: function ( left, right, bottom, top, znear, zfar, result ) {
```
cedricpinson: I would prefer to keep method start the same way, like makeFrustumInfinite or something like that, it's more convenient.

sources/osg/Matrix.js
```diff
@@ -1277,5 +1557,7 @@ define( [
         }
     };
 
+    Matrix.IDENTITY = [ 1.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 1.0 ];
```
cedricpinson: why not Matrix.Identity ?

sources/osg/MatrixTransform.js
```diff
@@ -18,13 +18,21 @@ define( [
...
         },
+
         setMatrix: function ( m ) {
             this.matrix = m;
+            this.dirtyBound();
         },
+
+        // local to "local world" (not Global World)
         computeLocalToWorldMatrix: function ( matrix /*, nodeVisitor */ ) {
+            if ( Matrix.isIdentity( this.matrix ) )
```
cedricpinson: bench

sources/osg/Node.js
```diff
@@ -227,29 +227,39 @@ define( [
             return this.boundingSphere;
         },
 
-        computeBound: function ( bsphere ) {
+        // TODO: PERF: Heavy GC impact
+        // find why making a closure var for bb breaks (v8&&ffx)
+        // frustum culling sample
+        //var  bbTemp = new BoundingBox();
```
cedricpinson: did you try as a member of the class ? just to test if it has an impact

sources/osg/Node.js
```diff
@@ -227,29 +227,39 @@ define( [
...
-                return bsphere;
+                bSphere.init();
+                return bSphere;
             }
-            bsphere._center = bb.center();
-            bsphere._radius = 0.0;
-            for ( var j = 0, l2 = this.children.length; j < l2; j++ ) {
-                var cc = this.children[ j ];
+            bSphere.set( bb.center( bSphere.center() ), 0.0 );
+            // not to do that because bigger results
```
cedricpinson: I dont understand, could you explain it ?

sources/osg/Plane.js
```diff
@@ -0,0 +1,137 @@
...
+    'osg/Matrix',
+    'osg/Utils',
+    'osg/Vec4',
+    'osg/Vec3',
+], function ( Matrix, MACROUTILS, Vec4, Vec3 ) {
+
+    'use strict';
+
+    /** @class Plane Operations */
+    var Plane = MACROUTILS.objectInehrit( Vec4, {
```
cedricpinson: did you check osg implementation, just to follow the api when possible

sources/osg/Polytope.js
```diff
@@ -0,0 +1,356 @@
...
+    /**
+     * Polytope class for representing convex clipping volumes made up of a set of planes.
+     * When adding planes, their normals should point inwards (into the volume)
+     * @class Polytope
+     */
+    var Polytope = function () {
+
+        this._clippingMask = 0x0;
+
+        this._planeList = [ Plane.create(), Plane.create(), Plane.create(), Plane.create(), Plane.create(), Plane.create() ];
```
cedricpinson: why not using new ? I remember that .create of those class is deprecated ( not sure )

sources/osg/Polytope.js
```diff
@@ -0,0 +1,356 @@
...
+            for ( i = 0; i !== pt._planeList.length; ++i ) {
+                if ( resultMask & selectorMask ) {
+                    this._planeList[ index ] = pt._planeList[ i ];
+                    Plane.transformProvidingInverse( this._planeList[ index++ ], matrix );
+                    this._resultMask = ( this._resultMask << 1 ) | 1;
+                }
+                selectorMask <<= 1;
+            }
+
+            this._maskStack[ this._maskStack.length - 1 ] = this._resultMask;
```
cedricpinson: we need a general stack we could add a TODO and an issue to fix this at one time

sources/osgShadow/ShadowFrustumIntersection.js
```diff
@@ -0,0 +1,204 @@
...
+                var bs = this._bs;
+                var matrix = this._matrixStack[ this._matrixStack.length - 1 ];
+                Matrix.transformBoundingSphere( matrix, node.getBound(), bs );
+
+                // camera cull
+                if ( this._cameraFrustum.getCurrentMask() !== 0 ) {
+                    // father bounding sphere is not totally inside
+                    // now test this one
+                    if ( !this._cameraFrustum.containsBoundingSphere( bs ) )
+                        return; // culled
```
cedricpinson: could we talk about tihs logic. It looks like if light or camera is culled we stop. I guess it's good but that would be great to add extra comments to explain the overall logic about it

sources/osgShadow/ShadowMap.js
```diff
@@ -475,131 +632,203 @@ define( [
...
-                centerDistance = Vec3.distance( worldLightPos, center );
 
-                zNear = centerDistance - radius;
-                zFar = centerDistance + radius;
+            Matrix.copy( projection, camera.getProjectionMatrix() );
+            Matrix.copy( view, camera.getViewMatrix() );
+            this.setShadowUniformsDepthValue( zNear, zFar, view, projection );
 
+            myBool = true;
+            if ( myBool )
```
cedricpinson: excuse me ? :)

sources/osgShadow/ShadowMap.js
```diff
@@ -475,131 +632,203 @@ define( [
...
+
+            // view and projection are now tightened.
+            //zFar = this._farCaster;
+            //zNear = this._nearCaster;
+
+            var lightSource = this._lightSource;
+            var light = lightSource.getLight();
+
+            var doBasicTrim = false;
+            if ( doBasicTrim ) {
```
cedricpinson: excuse me ? :) add comment to explain

sources/osgShadow/ShadowMap.js
```diff
@@ -475,131 +632,203 @@ define( [
...
+                    zFar = 1;
+                    zNear = 0.001;
+                }
+
+                var top, right;
+                var up = this._lightUp;
+
+                if ( Math.abs( Vec3.dot( up, worldLightDir ) ) >= 1.0 ) {
+                    // another camera up
+                    up = [ 1.0, 0.0, 0.0 ];
```
cedricpinson: camera up on X ?

sources/osgShadow/ShadowMap.js
```diff
@@ -608,47 +837,189 @@ define( [
...
+            if ( ortho )
+                Matrix.makeOrtho( l, r, b, t, n, f, projNew );
+            else
+                Matrix.makeFrustum( l, r, b, t, n, f, projNew );
+
+            Matrix.postMult( projNew, projection );
+        },
+
+        // Tighten projection matrices upon a bounding box
+        trimProjection2: function ( projection, view, bb, trimMask, ortho ) {
```
cedricpinson: is it for debug ?

sources/osgShadow/osgShadow.js
```diff
@@ -1,21 +1,23 @@
...
-    'osgShadow/ShadowMap'
-], function ( ShadowAttribute, ShadowTexture, ShadowedScene, ShadowSettings, ShadowTechnique, ShadowMap ) {
+    'osgShadow/ShadowTexture',
+    'osgShadow/ShadowedScene'
+], function ( ShadowAttribute, ShadowFrustumIntersection, ShadowMap, ShadowSettings, ShadowTechnique, ShadowTexture, ShadowedScene ) {
     'use strict';
 
     var osgShadow = {};
 
     osgShadow.ShadowAtribute = ShadowAttribute;
```
cedricpinson: ShadowAtribute, you forget a t a attribute

sources/osgUtil/osgUtil.js
```diff
@@ -50,4 +50,4 @@ define( [
 
     return osgUtil;
 
-} );
+} );
```
cedricpinson: could you configure you editor to stop to remove last blank line.

cedricpinson:
ok there are some point we need to discuss:
- performance improvement about using IDENTITY matrix test before doing the real operation
- some code that moved from a place to another, I think there are reason but we should be sure it makes sense and having a logic to move code that transform a shape by matrix. Could you check in osg first
- There are a lot of code to improve shadow, so as we talk does it worth it related to performance vs bigger texture. I know it's bad but in js it could be better use more texture resource instead of improving tighnest.

sources/osg/CullSettings.js
Kuranes: It's a way to handle the enableFrustumCulling at global scale.
(as in url parameter, overriding any further cullsettings which are per camera.)

The override check is here; https://github.com/cedricpinson/osgjs/pull/237/files#diff-ece7c42cf3b8e9cbd8392b2b82e1e6ebR274

Will add codesplaining

sources/osg/CullSettings.js
Kuranes: see previous answer (enablefrustumculling override)

sources/osg/CullVisitor.js
Kuranes: https://github.com/cedricpinson/osgjs/issues/239

sources/osg/Light.js
Kuranes: Could be either in osgShadow/shadowMap  (already a lot full) or a osgShadow/shadowFarCompute ?

(or we could separate shadowMap into shadowMap and a ShadowFrustumAlgo, allowing for multiple algo basic bbox, enhanced bbox, atlas CSM, etc... and incremental update without breaking existing)

sources/osg/Matrix.js
Kuranes: ok.
(it's about not recompute direction when we arleady have it, and don't have a target, like in the light case)

sources/osg/Matrix.js
Kuranes: Arg, to be erased. it's a Fake. an error..

sources/osg/Matrix.js
Kuranes: Ouch, debug code. Ought to be a lot of them.

*VERY* useful/interesting when debugging matrix (and transform/stack/visitor/etc...) so that you gain time by only stepping into matrix compute when it does something....

Doesn't slow thing from what I noticed but not benched, Could be that we make an issue to check at it when we have a bench Utilitary thing. Could even be faster as what it does finally is a kind of prefetch. (wish there was a native array compare that would do a memcmp)

Could bench that, with a special case for where a scene is "flattened", therefore all matrice but view is an identity...

Only place where I would let it is in the NodePath matrix building, where getting a nodepath full of identity matrix ought to be slower not matter what you do. 
( https://github.com/cedricpinson/osgjs/blob/491237d01fa4f2dbc9914089adc16aef30fdcb47/sources/osg/Node.js#L257 etc.)

So I remove them all but in nodepath building and make an issue for further investigation ?
(for when we have bench and a FlattenVisitor )

sources/osg/Matrix.js
Kuranes: no bbox transform in osg.
Usually there in math/3d libs.

sources/osg/Matrix.js
Kuranes: In shadowMap, as a simple projection matrix clamp:
https://github.com/cedricpinson/osgjs/blob/feature/multifrustum/sources/osgShadow/ShadowMap.js#L701
(comparing with what I get using the optimisation, could be that we can letting it as an alternative ShadowFrustumAlgo )

sources/osg/Matrix.js
Kuranes: ok, will do.

sources/osg/Matrix.js
Kuranes: I tought that was the norm ?
(from other class static const in other obj like Texture.FILTER, etc.)

sources/osg/Node.js
Kuranes: Tried, but had a set/copy problem with "center", could try again now that you can provide a result to bbox.center()

sources/osg/Plane.js
Kuranes: It's directly translated from c++ OSG (like polytope.js)

sources/osg/Polytope.js
Kuranes: https://github.com/cedricpinson/osgjs/issues/240

sources/osgShadow/ShadowFrustumIntersection.js
Kuranes: Not sure it stops. (but I should have tought of that, at least warning/log for lights and maybe stop for camera/projection).
Adding comments

sources/osgShadow/ShadowMap.js
Kuranes: Arg, some *private* debug code to clean in there.
Will make a cleaning pass.

sources/osgShadow/ShadowMap.js
Kuranes: debug code again... sry

sources/osgShadow/ShadowMap.js
Kuranes: yes, sry again...

sources/osgShadow/osgShadow.js
Kuranes: will fix

sources/osgUtil/osgUtil.js
Kuranes: done.

PR #243, opened 2015-02-03
Feature/multi frustum and shadow bounds

Kuranes:
- adds Polytope, Plane (didn't change polytope intersector though)
- per Camera Frustum (but kept the global enableFrustum override)
- VFC hirearchical optimisation (and CullSetting to specify what you want to cull)
- InfiniteMatrix Frustum (make it less useful/needed to compute/cull far and yet getting more precision)
- Shadow Frustum tightening to shadow Caster scene

### DIFF
--- a/sources/osg/CullStack.js
+++ b/sources/osg/CullStack.js
@@ -14,12 +14,12 @@ define( [
         this._bbCornerNear = 0;
 
 
-
         // keep a matrix in memory to avoid to create matrix
         this._reserveMatrixStack = [
             Matrix.create()
         ];
         this._reserveMatrixStack.current = 0;
+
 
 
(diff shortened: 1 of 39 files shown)


sources/osg/CullStack.js
```diff
@@ -140,7 +251,7 @@ define( [
             // we track camera node when using pushModelViewMatrix
             // To detect a camera, we check on the nodepath the type of the node and if the
             // camera is relatif or absolute.
-            // When we detect an absolute camera we keep it's index to get it when needed to
+            // When we det                        ect an absolute camera we keep it's index to get it when needed to
```
cedricpinson: fix

PR #249, opened 2015-02-11
Fixes light typo

stephomi:


cedricpinson:
@stephomi @Kuranes It looks like hemi are in the wrong side

PR #251, opened 2015-02-13
Fixes shadow Poisson adding better random

Kuranes:
Fixes shadow Poisson adding better random
Removes glsl for loop (chrome angle bug)
Fix bad copy/paste of position transform between directional and spot (good bug spot by @stephomi again)


sources/osgShadow/shaderLib.js
```diff
@@ -19,6 +21,8 @@ define( [
         'shadowLinearSoft.glsl': shadowLinearSoft,
         'bandPCF.glsl': bandPCF,
         'tapPCF.glsl': tapPCF,
+        'hash.glsl': hash,
+        'arrayPOISSON.glsl': arrayPOISSON,
```
cedricpinson: Poisson, acronyme uppercase, name camel case

Kuranes: ha, misunderstood then, it's even for private name (name of the guy)

sources/osgShadow/shaders/hash.glsl
```diff
@@ -0,0 +1,34 @@
+//// hash glsl
+
+//note: value noise
+float nHash11( const in float n )
```
cedricpinson: normalizedHash11 make function with explicit name or hash11Normalized

sources/osgShadow/shaders/vsm.glsl
```diff
@@ -16,7 +16,7 @@ float chebychevInequality (vec2 moments, float t)
     return variance / (variance + d * d);
 }
 
-float chebyshevUpperBound(vec2 moments, float mean, float bias, float minVariance)
+float chebyshevUpperBound(vec2 moments, float mean, float minVariance)
```
cedricpinson: is there some links ressources to explain what is chebyshev ?

Kuranes: https://github.com/cedricpinson/osgjs/blob/bugfix/shadowPoisson/sources/osgShadow/ShadowSettings.js#L72 ?

sources/osgShadow/shaders/hash.glsl
Kuranes: fixed

cedricpinson:
It's very sluggish with poisson8 like 2 fps, do you have the pb ?

Kuranes:
no. ( win7/nvidia gpu chrome/ffx )

cedricpinson:
for reference did you explore the following ?:
http://http.developer.nvidia.com/GPUGems2/gpugems2_chapter17.html http://http.developer.nvidia.com/GPUGems3/gpugems3_ch08.html - 8.6 Percentage-Closer Soft Shadows

About the slow down on poisson it's maybe that the shader explode

Kuranes:
1st one uses a 3D texture and dependant texture reads
2nd: PCSS is a 2nd step on shadow algo, not really for filtering (penumbra scale with distance to blocker, which means no penumbra when blocker and receiver are near.) note that is very slow and is intended for area light (you need a light size to estimate penumbra)

For Top algo on jittering/pcf, the reference everyone use is:
- http://the-witness.net/news/2013/09/shadow-mapping-summary-part-1/ (5x5 jitted pcf, which is 100texfect for use.)
- CSM . would even try for a "1 pass render into 4 viewport" (might need derivatives for clipspace culling ourself though)
- Even better if shadow moves but scene static: http://advances.realtimerendering.com/s2012/insomniac/Acton-CSM_Scrolling(Siggraph2012).pdf

Now what I would do to get really nice shadows:

- "deffered Shadows": http://www.gamedev.net/page/resources/_/technical/graphics-programming-and-theory/soft-edged-shadows-r2193 or http://www.realtimeshadows.com/sites/default/files/Playing%20with%20Real-Time%20Shadows_0.pdf p8  ( need 1 screen sized buffer more)

- Add temporal Enhancing of "deffered Shadows" which is really the best tool against shadow aliasing: http://www.cg.tuwien.ac.at/research/publications/2007/Scherzer-2007-PCS/#Additional Files and Images (need another screen sized buffer more for ping pong, and any buffer needed for  [https://github.com/cedricpinson/osgjs/tree/experiment/temporalpostfx](temporal algorithms) 

Reporting to Angle:
- we should report that and the android chrome slowdown on depth buffer pass, but need a demo to help them reproduce ?
(once fixed the model load thing, we could release a new version on github and they could check both there. Note that gsoc 2015 application started this week and end next week-end, would be nice to try again, but with updated osgjs.org)

cedricpinson:
Could you sort the list of proposition by time needed to implement it ?

Kuranes:
* Deferred shadows = it's about rendering all shadows in its own buffer, then blur it. (~ days maybe one, mostly "pipe work" in osgjs "shadowmap", and speed check experiment.)
* Temporal shadow = once we have that screenspace shadow buffer, ping-pong, reuse with temporal. (~week at most maybe less, using existing [temporal code/experiment] (rebased & fixed for tests on TSSA), and validating those osgjs internal changes about "history node matrix")

cedricpinson:
Ok good, I had the same feeling. About the blur in screen space how do you solve the blur that would bleed in the air ? for example when you do your blur it just blur everything. When you use this blur rendering on the final pass, how do you avoid to blur part of the scene that should not ?
Not sure if my question is clear ?

Kuranes:
- fast path: you don't prevent that, just gauss blur. (works well in practice from all papers and demo)
- slower but better path, you store depth+shadow in that screen space buffer, and use depth for bilateral edge aware blur Separable, or  Cross-Bilateral Gaussian Filter, or event, Cross-Bilateral Poisson Filter (whatever that works on webgl and use huge depth sudden change as blur "stopper"
 I think gpu pro1 talks about those, and , and sspcss papers gives solid maths and explanation on those (but they store distance to blocker, camera depth, light depth, etc...)

PR #257, opened 2015-02-19
Static Shadow support & better shadow traversal mask support

Kuranes:
Adds shadowMap filledOnce property that allows ensuring static shadowMap is built once at least after a dirty/init
Splits Shadow caster Traversal Maks in two, allowing for bounds & drawselection (you want to draw shadows, but not taking it in account in our
bounds, etc.)

### DIFF
--- a/sources/osgShadow/ShadowMap.js
+++ b/sources/osgShadow/ShadowMap.js
@@ -97,6 +97,7 @@ define( [
         this._casterStateSet = new StateSet();
         this._casterStateSet.addUniform( Uniform.createFloat1( 0, 'exponent0' ) );
         this._casterStateSet.addUniform( Uniform.createFloat1( 0, 'exponent1' ) );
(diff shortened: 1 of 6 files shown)


sources/osgShadow/ShadowMap.js
```diff
@@ -147,6 +153,14 @@ define( [
             return this._dirty;
         },
 
+        isFilledOnce: function () {
+            return this._filledOnce;
+        },
+
+        getEnable: function () {
```
cedricpinson: what means enable here ?

Kuranes: Enable shadow computing.
(!== shadow receiving)

Kuranes: Fixed by upping the filledOnce feature to shadowTechnique

PR #261, opened 2015-02-24
Fix for typos

jtorresfabra:


cedricpinson:
is it to be sync with osg convention right ? if yes could you keep the old accessors but displaying a deprecated message. I will remove old setEnable a bit later

jtorresfabra:
Ok, also I will change the setEnableFrustumCulling to setFrustumCullingEnabled, that seems the way it is done throughout the OSG code.

cedricpinson:
please add fallback temporary that display a warning message, 'deprecated' ...

jtorresfabra:
Ok, will do asap

PR #263, opened 2015-02-27
Bugfix/shadow transparency

Kuranes:
- Adds BLEND.DISABLE
- Adds no shadow casting for alpha blended/masked material
- Fix potential shadow shader compilation error
- Fix shadow texture binding for reading when writing in it

### DIFF
--- a/sources/osgShadow/ShadowMap.js
+++ b/sources/osgShadow/ShadowMap.js
@@ -335,9 +335,42 @@ define( [
             this._receivingStateset.setAttributeAndModes( this._shadowAttribute, StateAttribute.ON | StateAttribute.OVERRIDE );
 
 
+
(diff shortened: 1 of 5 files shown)


sources/osgShadow/ShadowMap.js
```diff
@@ -335,9 +386,42 @@ define( [
             this._receivingStateset.setAttributeAndModes( this._shadowAttribute, StateAttribute.ON | StateAttribute.OVERRIDE );
 
 
+
+            // prevent unnecessary texture bindings
+
+            if ( !ShadowMap.BlankTexture ) {
+                ShadowMap.BlankTexture = new Texture();
+                ShadowMap.BlankTexture.defaultType = true;
```
cedricpinson: no we dont do that now check Texture.js

PR #264, opened 2015-02-28
Bugfix/shadowTransparencyAndShaderCompilerOptim

Kuranes:
- Fix empty shadowmap case (could spout NaN and correctness depended on broswer handling
- Optimizes empty shadowmap case with early detection in the shader (depthRange.x == depthRange.y) and early out of shadow shader. (along with pushing computation after the early out)
- Makes CreateLighting more modular for shared vars between lights ( allow for optimisation when many lights and all use non light dependant precomputations.)
- Adds Extensions support to shaderprocessor and node shader compiler, …
following the GLSL spec (extension should be define before any non
processor instructions in shader compilation unit)

### DIFF
--- a/sources/osgShadow/ShadowMap.js
+++ b/sources/osgShadow/ShadowMap.js
@@ -30,6 +30,7 @@ define( [
 
     var TransparentRemoveVisitor = function ( mask ) {
         NodeVisitor.call( this );
+        // mask setting to avoid casting shadows
         this._noCastMask = mask;
         this._nodeList = [];
     };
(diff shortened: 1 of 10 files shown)


sources/osgShadow/ShadowMap.js
```diff
@@ -48,8 +48,8 @@ define( [
                     if ( nm === ~0x0 ) {
                         nm = this._noCastMask;
                         node.setNodeMask( nm );
-                    }
-                    if ( ( nm & ~( this._noCastMask ) ) !== 0 ) {
+                        this._nodeList.push( node );
+                    } else if ( ( nm & ~( this._noCastMask ) ) !== 0 ) {
```
cedricpinson: comments would help to explains the mask

Kuranes: added comments

PR #268, opened 2015-03-04
Fix code crash when state is passed as undefined to Texture module

hishamabdelrahman:
when using PagedLod Model somtimes the state is passed undefined to
relaseGLObjects function so it needs to be handeled so code wont crash

### DIFF
--- a/sources/osg/Texture.js
+++ b/sources/osg/Texture.js
@@ -91,7 +91,7 @@ define( [
     Texture.FLOAT = 0x1406;
     Texture.HALF_FLOAT_OES = Texture.HALF_FLOAT = 0x8D61;
 
-    Texture.getEnumFromString = function( v ) {
+    Texture.getEnumFromString = function ( v ) {
         var value = v;
         if ( typeof ( value ) === 'string' ) {
             value = checkAndFixEnum( value, v );
@@ -214,12 +214,12 @@ define( [
         },
 
         releaseGLObjects: function ( state ) {
-		 if (state !==undefined){
-            if ( this._textureObject !== undefined && this._textureObject !== null ) {
-                state.getTextureManager().releaseTextureObject( this._textureObject );
-                this._textureObject = undefined;
+            if ( state !== undefined ) {
+                if ( this._textureObject !== undefined && this._textureObject !== null ) {
+                    state.getTextureManager().releaseTextureObject( this._textureObject );
+                    this._textureObject = undefined;
+                }
             }
-		 }
         },
 
 
--- a/sources/osg/Texture.js
+++ b/sources/osg/Texture.js
@@ -214,10 +214,12 @@ define( [
         },
 
         releaseGLObjects: function ( state ) {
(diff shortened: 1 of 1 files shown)


jtorresfabra:
Hi Hisham, 

Check the identation and spacing to be concordant with the rest of osgjs code. Also, do you have some test data to reproduce the issue?

hishamabdelrahman:
Hi Jordi,
Actually indenting was ok on my nodepad++ , just when i sync that happend
yes i do have data ...but its too big as its pagedLod model so it involves lots of child models files to reproduce.what do u suggest?

jtorresfabra:
Can you share it in a downloadable URL(maybe zip compressed)? or if you have it public I can debug myself in your URL. Also there are options like weTransfer, google drive or dropbox. 

About the identation  jsbeautify is used for the sources. There is a .jsbeautifyrc in the root folder of osgjs.

hishamabdelrahman:
ok i will zip it and put it on my Google drive and will send the url
and thanks for the info about jsbeautify ... i will use it from now on
best regards

hishamabdelrahman:
Hi Jordi,
sorry for my late response
I had uploaded the data on google drive ,
https://drive.google.com/file/d/0B9Z19aal418PM04xbUhuRG5ONHc/view?usp=sharing

also i get another  error please see the image below 
image

jtorresfabra:
Hi @hishamabdelrahman, 
I just ask you for permission to download the file from my google account. It seems you need to give me rights to download. 

Thanks.

hishamabdelrahman:
Sorry Ur email comes in French from google drive ...which I did not
understand :) I think I shared it now with you ...plz confirm

jtorresfabra:
ok, I have it now. 
By the way I'm spanish ;).

jtorresfabra:
Hi @hishamabdelrahman 

I have reproduced the error but I feel your modification is not a "real" solution for the problem. An older commit broke the texture release stuff from the databasepager, and we need to address it in a different way. 
Anyway, as your fix avoids the crash we can safely include it till the "real" solution is implemented. ( (@cedricpinson ) 

About the other problem is better to open a different issue, to not merge topics. Also I would recommend to add more information in that issue, not just a Visual Studio image :).

Thanks.

hishamabdelrahman:
i Fully agree with you that its not a real solution...i still new on osgjs so its like a work around for now
the second error
I really don't know what is happening as all of a sudden it crashes at that line shown in the snapshot ...
with undefined error message...were you able to reproduce it? because its an intermittent issue where the code crashes

jtorresfabra:
Please open a different issue. We can discuss it there.

hishamabdelrahman:
hi , 
Related to the undefined  State problem , i found out that Code in node Class for this.stateset.releaseGLObjects(); never pass a state that's why the state comes as  undefined

 releaseGLObjects: function ( /*gl*/) {
            if ( this.stateset !== undefined ) this.stateset.releaseGLObjects();
        }

PR #270, opened 2015-03-06
Bugfix/shadowBias

Kuranes:
- Add multiple bias code, most useful being the derivatives model. 
- Fixes bug on incorrect frustum optimisations
- Optimizes Out uniform not needed
- Add more compiler test for shader node Type unicity

### DIFF
--- a/sources/osgShader/node/shadows.js
+++ b/sources/osgShader/node/shadows.js
@@ -25,10 +25,10 @@ define( [
 
         // must return an array of defines
         // because it will be passed to the ShaderGenerator
-        defines: function () {
+        getDefines: function () {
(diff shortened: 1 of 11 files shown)


sources/osgShader/node/shadows.js
```diff
@@ -28,9 +28,13 @@ define( [
         defines: function () {
             return this._shadow.getDefines();
         },
-
+        extensions: function () {
```
cedricpinson: we should be consistent and use get when it does get, so extensions should be getExtensions, define getDefines.
If you can adjust, for the new field and make it backward compatible for other with a warning

PR #278, opened 2015-03-12
Uses projection matrix for computeHomePosition

stephomi:


sources/osgGA/SwitchManipulator.js
```diff
@@ -67,6 +76,24 @@ define( [
             if ( manipulator !== undefined ) {
                 return manipulator.getInverseMatrix();
             }
+        },
+        setEyePosition: function ( arg ) {
```
cedricpinson: Those functions are more application dependent:
 - getEyePosition
 - setEyePosition
 - setTarget
 - getTarget
 - setDistance
 - getDistance

See https://github.com/openscenegraph/osg/blob/master/include/osgGA/CameraManipulator

stephomi:
Updated

PR #289, opened 2015-03-27
Fixes spot light frustum optimizatin when inside scene bbox

Kuranes:
- better near/far bounding, use of those for the optimized case
- more robust transformation handling for spot
- DRY code (near/far, lightup, single epsilon)

### DIFF
--- a/sources/osgShadow/ShadowMap.js
+++ b/sources/osgShadow/ShadowMap.js
@@ -101,9 +101,8 @@ define( [
                 var m = nv.getCurrentProjectionMatrix();
 
                 //Matrix.clampProjectionMatrix( m, nv.getComputedNear(), nv.getComputedFar(), nv.getNearFarRatio() );
-                // we should use accessors when touching external object !!!
-                this._shadowTechnique._nearCaster = nv.getComputedNear();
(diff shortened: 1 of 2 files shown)


sources/osgShadow/ShadowMap.js
```diff
@@ -102,8 +102,8 @@ define( [
 
                 //Matrix.clampProjectionMatrix( m, nv.getComputedNear(), nv.getComputedFar(), nv.getNearFarRatio() );
                 // we should use accessors when touching external object !!!
-                this._shadowTechnique._nearCaster = nv.getComputedNear();
-                this._shadowTechnique._farCaster = nv.getComputedFar();
+                this._shadowTechnique._depthRange[ 0 ] = nv.getComputedNear();
```
cedricpinson: use accessors when field are +1 indirection

cedricpinson:
is it possible to test this issue, to lock the code. I am sure what was the problem and how it is solved

Kuranes:
Wasn't something that broke, but one use case not tested in the demo.
Could add pure "data" test from use case into unit tests to prevent regression ?

cedricpinson:
I the data are small yes, but it must be a unittest, not something loading a scene ...

PR #296, opened 2015-04-08
[#91685744] Fixes Absolute Ref Node above light Node for shadow

Kuranes:
Fixes Absolute Ref Node above light Node for shadow computation

### DIFF
--- a/sources/osg/Light.js
+++ b/sources/osg/Light.js
@@ -305,6 +305,15 @@ define( [
             return this._ground[ 3 ] >= 0.0;
         },
 
+        getPositionnedMatrix: function () {
+            var uniformMap = this.getOrCreateUniforms();
+            return uniformMap.matrix.get();
+        },
+
+        getPositionnedInvMatrix: function () {
+            var uniformMap = this.getOrCreateUniforms();
+            return uniformMap.invMatrix.get();
+        },
         applyPositionedUniform: function ( matrix /*, state*/ ) {
 
             var uniformMap = this.getOrCreateUniforms();
--- a/sources/osgShadow/ShadowMap.js
+++ b/sources/osgShadow/ShadowMap.js
@@ -185,6 +185,7 @@ define( [
         this._tmpVec = Vec3.create();
         this._tmpVecBis = Vec3.create();
         this._tmpVecTercio = Vec3.create();
+        this._tmpMatrix = Matrix.create();
         if ( settings )
             this.setShadowSettings( settings );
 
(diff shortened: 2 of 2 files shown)


cedricpinson:
Did you check that you dont have the information during the culling ?

Kuranes:
Problem is complex:
- need light world and eye matrix, and/or modelview"
- need scene world and eye matrix  and/or modelview

But Light can be anywhere (meaning not under the same camera or absolute Ref Node), so culling doesn't give you much.

You might have a light under camera A (or under none?) used for casting shadow on a scene under camera B, but shadow computation of matrix must reflect light computation of uniform matrix (even if *wrong*) so that they're consistent one with another

As it is that new code use the Cullvisitor cull info, using current modelview And light matrix computed from latest "light::positionnedUpdate". 
Which means light shadowing NEEDS to be culled before shadow scene otherwise you end up with a matrix one frame late.

sources/osg/Light.js
```diff
@@ -305,6 +305,15 @@ define( [
             return this._ground[ 3 ] >= 0.0;
         },
 
+        getPositionnedMatrix: function () {
```
cedricpinson: those methods should be removed, there are not consistent in other usecase than shadow

sources/osgShadow/ShadowMap.js
```diff
@@ -780,37 +781,61 @@ define( [
...
+
             Matrix.copy( camera.getProjectionMatrix(), this._projectionMatrix );
             Matrix.copy( camera.getViewMatrix(), this._viewMatrix );
             var projection = this._projectionMatrix;
             var view = this._viewMatrix;
 
             // inject camera world matrix.
-            // from light current world/pos
+            // from light current world/pos and camera eye pos.
+            var lightMatrix = light.getPositionnedMatrix();
```
cedricpinson: I guess here we need to get LightSource information in the cull visitor. I guess it can be extracted from current renderstage positionned attribute. or looking into the stack node.

Kuranes: isn't light::getPositionnedMatrix exaclty that ?

cedricpinson: Yes but you added methods in light that only make sense in your context. So
it s more appropriate to take this from cull visitor.
But in general what is computed in cull visitor Computed data about
transform from cull visitor should stay in cull / draw visitor

Kuranes: ok, understood.

cedricpinson: but if there is something not clear to dig in cull visitor we can do it together

Kuranes:
https://github.com/cedricpinson/osgjs/pull/299

Issue #291, opened 2015-03-31
PagedLOD is not freeing memory anymore


hishamabdelrahman:
i believe that the issue is coming from the following

In Node Class
the function releaseGLObjects doesnot pass state object as well as when calling the stateset.releaseGLObjects we dont pass the expected state object

  releaseGLObjects: function ( /*gl*/) {

           
            if ( this.stateset !== undefined ) this.stateset.releaseGLObjects();
        }

then in stateset Class the Function releaseGLObjects  which Expects state object 

  releaseGLObjects: function ( state ) {
            // TODO: We should release Program/Shader attributes too
            for ( var i = 0, j = this.textureAttributeMapList.length; i < j; i++ ) {
                this.getTextureAttribute( i, 'Texture' ).releaseGLObjects( state );
            }
        },
as state is passed as null this  will result to  fail to release the texture when the releaseGLObjects is called on the texture Class as state would always be passed as undefined

 releaseGLObjects: function ( state ) {
            if ( state !== undefined ) {
                if ( this._textureObject !== undefined && this._textureObject !== null ) {
                    state.getTextureManager().releaseTextureObject( this._textureObject );
                    this._textureObject = undefined;
                }
            }
        },

The Intial Code was working for object.releaseGLObjects as the texture as it was not expecting a state object

releaseGLObjects: function ( ) {
            if ( this._textureObject !== undefined && this._textureObject !== null ) {
                Texture.textureManager.releaseTextureObject( this._textureObject );
                this._textureObject = undefined;
                this._image = undefined;
            }
        },

jtorresfabra:
Hi @hishamabdelrahman 
There is more work to be done, but can you try https://github.com/cedricpinson/osgjs/pull/297 to see if it fixes the issues for you? thanks.

hishamabdelrahman:
Thanks jordi ..I saw your pull request .u believe it should fix my issue
...many thanks for Ur support

jtorresfabra:
it can be closed, fixed in #297

hishamabdelrahman:
i verified the latest code with my data , yes its working well now

Issue #222, opened 2015-01-07
GL object tagging "to release"


Kuranes:
When without the "state" (outside of draw) it's not possible to release GL Objects as we don't have gl context. 
But we can tag the GL ressource "toRelease", and its ressource manager will handle that release asynchronously.

jtorresfabra:
It can be closed, fixed in #297

PR #297, opened 2015-04-09
Added a TextureManager Map to store a TextureManager for each GraphicCon...

jtorresfabra:
...text. This way we can emulate the behaviour of OSG when releasing textures. 

It is not a complete fix for all the releaseGL stuff, it's only for Textures so DatabasePager frees Texture memory again.  
We can discuss it further, but I think this way we can release other GLObjects with and without state, as OSG does. In OSG a static variable keeps track of GLObjects or each class with respect a contextID.

### DIFF
--- a/sources/osg/StateSet.js
+++ b/sources/osg/StateSet.js
@@ -217,6 +217,12 @@ define( [
             for ( var i = 0, j = this.textureAttributeMapList.length; i < j; i++ ) {
                 this.getTextureAttribute( i, 'Texture' ).releaseGLObjects( state );
             }
+            var list = this.getAttributeList();
+                for ( i = 0, j = list.length; i < j; i++ ) {
+                    // Remove only if we have releaseGLObject method. 
+                    if ( list[ i ]._object.releaseGLObjects )
+                        list[ i ]._object.releaseGLObjects( state );
+            }
         },
         _getUniformMap: function () {
             return this.uniforms;
--- a/sources/osg/StateSet.js
+++ b/sources/osg/StateSet.js
@@ -218,10 +218,11 @@ define( [
                 this.getTextureAttribute( i, 'Texture' ).releaseGLObjects( state );
             }
             var list = this.getAttributeList();
-                for ( i = 0, j = list.length; i < j; i++ ) {
-                    // Remove only if we have releaseGLObject method. 
-                    if ( list[ i ]._object.releaseGLObjects )
-                        list[ i ]._object.releaseGLObjects( state );
+            for ( i = 0, j = list.length; i < j; i++ ) {
+                // Remove only if we have releaseGLObject method. 
+                if ( list[ i ]._object.releaseGLObjects ) {
+                    list[ i ]._object.releaseGLObjects( state );
+                }
             }
         },
         _getUniformMap: function () {
--- a/sources/osg/StateSet.js
+++ b/sources/osg/StateSet.js
@@ -212,16 +212,15 @@ define( [
         getShaderGeneratorName: function () {
             return this._shaderGeneratorName;
         },
-        releaseGLObjects: function ( state ) {
-            // TODO: We should release Program/Shader attributes too
+        releaseGLObjects: function () {
             for ( var i = 0, j = this.textureAttributeMapList.length; i < j; i++ ) {
-                this.getTextureAttribute( i, 'Texture' ).releaseGLObjects( state );
+                this.getTextureAttribute( i, 'Texture' ).releaseGLObjects();
             }
             var list = this.getAttributeList();
             for ( i = 0, j = list.length; i < j; i++ ) {
                 // Remove only if we have releaseGLObject method. 
                 if ( list[ i ]._object.releaseGLObjects ) {
-                    list[ i ]._object.releaseGLObjects( state );
+                    list[ i ]._object.releaseGLObjects();
                 }
             }
         },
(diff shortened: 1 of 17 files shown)


jtorresfabra:
Should fix #291 #222.

sources/osg/StateSet.js
```diff
@@ -217,6 +217,12 @@ define( [
             for ( var i = 0, j = this.textureAttributeMapList.length; i < j; i++ ) {
                 this.getTextureAttribute( i, 'Texture' ).releaseGLObjects( state );
             }
+            var list = this.getAttributeList();
+                for ( i = 0, j = list.length; i < j; i++ ) {
```
cedricpinson: indentation seems bad

cedricpinson:
I dont see method to mark data to release, I expected to see something like:
texture.markToRelease() or someting like that. Just to be sure it 's something to add in the futur right ?

sources/osg/StateSet.js
jtorresfabra: Fixed

jtorresfabra:
Nope, you simply need to call texture.releaseGLObjects() to get it deleted asynchronously in the next frame. If you want to delete it from a specific Context you pass the state as a parameter, if you don't pass the state then it will be deleted from all contexts. 

Ping me if you need a better explanation, maybe some comments in the code could be good.

cedricpinson:
I understand and maybe we dont need more, but in c++ you have destructor to tell a object to release, but not in js. I thought that it would be helpful to have a special function to say release this texture, but without more arguments. Because the texture is created from a specific context the texture is able to release from the same context. I guess that seems we dont have shared resources between context, the no arguments that delete from all context does the job.
So maybe we dont need more function to do that for now

jtorresfabra:
>I understand and maybe we dont need more, but in c++ you have destructor to tell a object to release, but not in js. I thought that it would be helpful to have a special function to say release this texture, but without more arguments. Because the texture is created from a specific context the texture is able to release from the same context.

Yes, but a texture can be used in several contexts, ( maybe not now in osgjs, but it will/should do) so you need to specify from which context you want to release or release from all contexts (as it works in OSG for Textures and other GLObjects).

>I guess that seems we dont have shared resources between context, the no arguments that delete from all context does the job.

Yes that's the point. 

>For example if you want to release all resource of Geometry you can't call the function from from geometry, you need state. So it does not emulate a destructor.

Now BufferArray is missing a BufferObjectManager (as Texture has its own TextureManager) to be able to release from all contexts. In fact OSG has a BufferObjectManager. I implemented the BufferArray static to delete cache the easy way, not using a manager, I tried to say it in the IRC. I wanted to show the idea, and how it works.
 
Once we have this manager you will be able to call releaseGLObjects on Geometry safely without passing the state. The most internal classes (TextureManager and BufferArrayManager) will be responsibles of managing the "no-state" situation, as TextureManager does now. So for Geometry class it will be trasnparent.  

For Shaders and Programs it is not the same, in OSG you can only release them using a context ID, not as BufferObejcts and Textures, which makes sense.

jtorresfabra:
I've been having a look to the "Sharing WebGLresources" state. It seems is a proposal/draft https://www.khronos.org/webgl/wiki/SharedResouces of extension. Unfortunately it seems latest chrome does not support the extension (at least mine).
https://www.khronos.org/registry/webgl/sdk/tests/conformance/context/resource-sharing-test.html?webglVersion=1

Said that, TextureManager is prepared in case we need that support. It won't fail trying to delete a texture from a different context from where it was created. 
For Buffer Objects/Programs/Shaders we can simply keep a reference to the glcontext when the resource is binded/created and store it as a class variable. Then you can call releaseGLObjects() without the state as a paramenter. The real thing we need to delete resources is the GL context, not the state. Implement that based on this PR would be pretty straight forward. 

Thoughts?

jtorresfabra:
@cedricpinson:

After looking at http://webglstats.com to check the availability of WebGL Shared Resources extensions. I've just gone ahead with the idea of having a reference to the graphic context in each glresource and it seems to work well. Now you don't need to call releaseGLResources with a parameter. So at the end calling releaseGLResources() is like doing that markToBeReleased we were talking.

cedricpinson:
Ok,
the only thing missing is the manager for other GL resources to have the same way to delete textures, right ?

jtorresfabra:
Really we don't need it. The textures stuff is different because it is using a TextureManager to reuse the texture objects.

cedricpinson:
ok but it means you dont control the time / delay to delete those resource because it's not delayed by a manager, and applied per frame. Or maybe I miss something.

jtorresfabra:
Indeed it IS controlled. Each GLResource has its own static to delete cache, which is flushed every frame.
For example (For buffer array):
When ther is a call to releaseGLObejcts() the buffer is pushed to the cache.
https://github.com/cedricpinson/osgjs/pull/297/files?diff=unified#diff-e466d62e2e95574ba588e47efa37270aR50
Then the next frame the viewer flushes that cache:
https://github.com/cedricpinson/osgjs/pull/297/files?diff=unified#diff-50f9ea5b8e120acc5d8e00a8ae9cb389L275

jtorresfabra:
Maybe it could be easier to understand loading the PagedLOD example a putting a breakpoint in  the above lines. Then play a little with the example and it will be stopped at the breakpoints deleting buffers.

jtorresfabra:
Updated the PR as discussed in IRC. It's ready for review ( I cannot change the PR labels ).

PR #298, opened 2015-04-10
Adjusts the panning speed depending of the fov value

stephomi:


sources/osgGA/SwitchManipulator.js
```diff
@@ -77,8 +76,11 @@ define( [], function () {
                 return manipulator.getInverseMatrix();
             }
         },
-        setByMatrix: function ( matrix ) {
```
cedricpinson: Why is it removed ?

stephomi: Because it's not used.
I added it when I did the compute home position projection stuffs.
But I didn't implemented the function in the orbit and fps manipulator and I forgot to remove it.

PR #299, opened 2015-04-16
[#91685744] Fixes Absolute Ref Node above light Node for shadow

Kuranes:
- Add new constraint on shadowing ligh : the light node must be under shadowedScene Node in the hierarchy scene graph

### DIFF
--- a/sources/osg/RenderStage.js
+++ b/sources/osg/RenderStage.js
@@ -71,6 +71,9 @@ define( [
         getCamera: function () {
             return this.camera;
         },
+        getPositionedAttribute: function () {
+            return this.positionedAttribute;
+        },
         addPreRenderStage: function ( rs, order ) {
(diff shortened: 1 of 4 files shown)


sources/osg/RenderStage.js
```diff
@@ -71,6 +71,17 @@ define( [
         getCamera: function () {
             return this.camera;
         },
+        getPositionnedAttribute: function ( node ) {
```
cedricpinson: Seems better, just add the accessors to return the positioned attribute. We will put the find inside shadowMap for now.

jtorresfabra: Sorry to bother, but @Kuranes shouldn't it be 'positioned' instead of 'positionned'?

Kuranes: Fixed & repushed
@cedricpinson moved it to shadowMap
@jtorresfabra Good spot, fixed to positioned.

PR #301, opened 2015-04-24
Adds triangle sphere intersection support

stephomi:
Maybe it requires a test and an example.
But it's here for code review.

### DIFF
--- a/examples/picking/main.js
+++ b/examples/picking/main.js
@@ -1,6 +1,13 @@
 'use strict';
 
-function getShader() {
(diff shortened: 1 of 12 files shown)


stephomi:
I added a test.

cedricpinson:
@jtorresfabra do you want to make a test on this PR before I merge it ?

jtorresfabra:
The tests are passing, so thumbs up from my side. I will give a try with my scenes when I have the time. It's a cool PR @stephomi !

PR #302, opened 2015-04-28
Improvements to DatabasePager: new methods to start/stop requests(update...

jtorresfabra:
...d pagedLOD example) and to set the dabatase path


jtorresfabra:
I need to check some things

PR #304, opened 2015-04-30
Adds serializer and class for animation

gfaucill:


sources/osgWrappers/serializers/osgAnimation.js
```diff
@@ -238,5 +264,112 @@ define( [
...
+            var promise = input.setJSON( jsonAttribute ).readBufferArray();
+            Q.when( promise ).then( function ( buffer ) {
+                if ( buffer !== undefined ) {
+                    bezierChannel[ name ] = buffer._elements;
+                }
+                defer.resolve( buffer );
+            } );
+        };
+
+        for ( var k in jsonObj.KeyFrames ) {
```
cedricpinson: the way to iterate on keys of an object is:
var keys = Objects.keys(jsonObj.KeyFrames);
for ( var i = 0; i < keys.lenght; i++ ) {
  var key = keys[i];
  var value = jsonObj.KeyFrames[key];
  // do something with value
}

PR #307, opened 2015-05-12
Adds light/lightSource and Camera with RTT as non shadow caster.

Kuranes:
Fixes PCF bug
Fixes Shader error logging when parsing misses first error
Adds light/lightSource and Camera with RTT as non shadow caster. (like transparent geometries waswas)


cedricpinson:
could you comment wich code fix want ? or split by commit

Kuranes:
separated on https://github.com/cedricpinson/osgjs/pull/309

PR #308, opened 2015-05-13
Texture compression

stephomi:
There's probably something wrong in the dxt5ToRgba8888 algorithm though.

dxt5ToRgba8888  and dxt1ToRgb565 are duplicated in the worker, it's in the example, so it's not really an issue.

I refactor a little bit the Image osgjs class to support mipmap.
Maybe it's not perfect but it should work.


jtorresfabra:
Hi @stephomi  regarding the promise clean up, one doubt: is this the correct way to do promises? I mean Q(promise) instead of Q.when(promise)? I ask because I', also using Q in other library.

Also another question (cc @cedricpinson ) what about upgrading Q to 1.3.0 (http://github.com/kriskowal/q/tree/v1.3.0)?

stephomi:
`Q.when(promise, cbSuccess, cbFail)` is equivalent to `promise.then(cbSuccess).fail(cbFail)`.
But `Q.when(promise).then(cbSuccess).fail(cbFail)` is different though !

Here's a fiddle :
http://jsfiddle.net/r64ov7m3/2/

sources/osg/Texture.js
```diff
@@ -382,13 +410,20 @@ define( [
             this._unrefImageDataAfterApply = bool;
         },
 
+        checkIsCompressed: function ( format ) {
+            var fo = format || this._internalFormat;
+            return fo !== Texture.RGBA && fo !== Texture.RGB && fo !== Texture.LUMINANCE && fo !== Texture.LUMINANCE_ALPHA;
```
cedricpinson: it would be better to test the compression enum

sources/osg/Texture.js
```diff
@@ -614,11 +670,9 @@ define( [
...
         Notify.log( 'Texture.createFromURL is deprecated, use instead osgDB.readImageURL' );
         var texture = new Texture();
-        Q.when( ReaderParser.readImage( imageSource ) ).then(
-            function ( img ) {
-                texture.setImage( img, format );
-            }
-        );
+        Q( ReaderParser.readImage( imageSource ) ).then( function ( img ) {
+            texture.setImage( img, format );
+        } );
```
cedricpinson: we should add a .fail in this case

stephomi:
I removed the promise commit from this PR.
A change in promise management will be done in https://github.com/cedricpinson/osgjs/pull/317.

PR #309, opened 2015-05-13
Bugfix/shadow pcf cullnodes fakepcf

Kuranes:
Fixes PCF bug
Fixes Shader error logging when parsing misses first error
Adds light/lightSource and Camera with RTT as non shadow caster. (like transparent geometries waswas)

### DIFF
--- a/examples/shadowmap/main.js
+++ b/examples/shadowmap/main.js
@@ -30,7 +30,7 @@
             'model': 'material-test',
             'shadowProjection': 'fov',
             'fov': 50,
-            'kernelSizePCF': '4Band(4texFetch)',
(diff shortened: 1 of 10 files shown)


Kuranes:
You can try here https://dl.dropboxusercontent.com/u/2024901/shadowpcf/examples/shadowmap/index.html

( changed default to 1 TAP (4 texfech) NOT fakePCF )

For Mipmap + linear + anisotropy you have to:
- enable fakePCF, 
- enable TAP pcf mode (4 or 9TAP) 
- enable FLOAT_LINEAR or HALF_FLOAT_LINEAR (otherwise you're mipmapping float as encoded rgba...))

( change lightheight and change to light direction for different model lighting settings )

PR #316, opened 2015-05-21
Shadow Receivers position on Shadows Light Space is now computed in

Kuranes:
fragment shader

### DIFF
--- a/sources/osgShader/Compiler.js
+++ b/sources/osgShader/Compiler.js
@@ -298,13 +298,7 @@ define( [
             for ( t = 0; t < this._shadows.length; t++ ) {
                 this.declareAttributeUniforms( this._shadows[ t ] );
             }
-            /*
-            for ( t = 0; t < this._shadowsTextures.length; t++ ) {
-                if ( this._shadowsTextures[ t ] !== undefined ) {
-                    this.declareAttributeUniforms( this._shadowsTextures[ t ] );
-                }
-            }
-*/
+
 
         },
 
@@ -579,24 +573,24 @@ define( [
             // TODO: multi shadow textures for 1 light
             for ( k = 0; k < this._shadows.length; k++ ) {
                 shadow = this._shadows[ k ];
-                if ( shadow.getLight() === light ) {
-                    lightIndex = k;
-                    for ( var p = 0; p < this._shadowsTextures.length; p++ ) {
-                        shadowTexture = this._shadowsTextures[ p ];
-                        if ( shadowTexture && shadowTexture.getLightUnit() === light.getLightNumber() ) {
-                            shadowTextures[ p ] = shadowTexture;
-                            hasShadows = true;
-                        }
+                if ( shadow.getLight() !== light ) continue;
+
+                lightIndex = k;
+                for ( var p = 0; p < this._shadowsTextures.length; p++ ) {
+                    shadowTexture = this._shadowsTextures[ p ];
+                    if ( shadowTexture && shadowTexture.getLightUnit() === light.getLightNumber() ) {
+                        shadowTextures[ p ] = shadowTexture;
+                        hasShadows = true;
                     }
                 }
-
             }
             if ( !hasShadows ) return undefined;
 
+            // Varyings
+            var vertexWorld = this.getOrCreateVarying( 'vec4','WorldPosition' ) ;
+
             // asserted we have a shadow we do the shadow node allocation
             // and mult with lighted output
-
-
             var shadowedOutput = this.createVariable( 'float' );
 
             // shadow Attribute uniforms
@@ -604,7 +598,7 @@ define( [
             inputs = MACROUTILS.objectMix( inputs, shadowUniforms );
 
             // shadowTexture  Attribute uniforms AND varying
-            var tex, shadowVertexProjected;
+            var tex;
             // TODO: better handle multi texture shadow (CSM/PSM/etc.)
             for ( k = 0; k < shadowTextures.length; k++ ) {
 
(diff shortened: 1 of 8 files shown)


sources/osgShader/Compiler.js
```diff
@@ -414,6 +408,7 @@ define( [
             if ( diffuseColor === undefined )
                 return undefined;
 
+            // investigate lowp
```
cedricpinson: does the comment is up to date ?

sources/osgShader/Compiler.js
```diff
@@ -515,6 +510,7 @@ define( [
             var texCoordUnit = unit;
             var texCoord = this.getVariable( 'FragTexCoord' + texCoordUnit );
             if ( texCoord === undefined ) {
+                // investigate lowp
```
cedricpinson: comment not removed ?

sources/osgShader/Compiler.js
```diff
@@ -926,9 +922,12 @@ define( [
         declareVertexVariables: function () {
             var texCoordMap = {};
 
+            // TODO: for performance:
+            // - remove non necessary attributes
+            // - remove unecessary varying (huge impact)
             this._vertexShader.push( [ '',
                 'attribute vec3 Vertex;',
-                'attribute vec4 Color;',
+                'attribute vec4 Color;',// investigate lowp
```
cedricpinson: comment ?

sources/osgShader/Compiler.js
```diff
@@ -936,15 +935,16 @@ define( [
                 'uniform mat4 ProjectionMatrix;',
                 'uniform mat4 NormalMatrix;',
                 '',
-                'varying vec4 VertexColor;',
-                'varying vec3 FragNormal;',
-                'varying vec3 FragEyeVector;',
+                'varying vec4 VertexColor;',// investigate lowp
```
cedricpinson: ok I guess what you wanted to say is:
we should investigate different precision for varying data ....

I thought you did test and you forget to remove them. Could you adjust the comments to reflex what it means, like a todo to test stuff

Kuranes: ha, yes. in fact, following this morning conversation, it's a bigger job that just enabling precision here (need check input, need change arraybuffer, etc)
 maybe an osgjs Issue ? 
(better handling of precsion in vertex attribute)
and then removing code comments here

Kuranes: Repushed with comment sanitized.
Add an osgjs issue ?

sources/osgShader/Compiler.js
```diff
@@ -953,20 +952,9 @@ define( [
...
-                var depthRange = shadowTextureUniforms.DepthRange;
-                var mapSize = shadowTextureUniforms.MapSize;
-                // uniforms
-                this._vertexShader.push( 'uniform mat4 ' + projMat.getName() + ';' );
-                this._vertexShader.push( 'uniform mat4 ' + viewMat.getName() + ';' );
-                this._vertexShader.push( 'uniform vec4 ' + depthRange.getName() + ';' );
-                this._vertexShader.push( 'uniform vec4 ' + mapSize.getName() + ';' );
-                // varyings
-                this._vertexShader.push( 'varying vec4 ' + shadowTexture.getVaryingName( 'VertexProjected' ) + ';' );
                 hasShadows = true;
```
cedricpinson: is it not a duplicate because of this https://github.com/cedricpinson/osgjs/pull/316/files?diff=split#diff-b88b11738b3c5573469532874321f000R953 ?

sources/osgShader/Compiler.js
```diff
@@ -1012,22 +1000,9 @@ define( [
                     continue;
                 if ( !hasShadows ) {
                     hasShadows = true;
-                    this._vertexShader.push( 'vec4 worldPosition = ModelWorldMatrix * vec4(Vertex,1.0);' );
+                    this._vertexShader.push( 'WorldPosition = ModelWorldMatrix * vec4(Vertex,1.0);' );
```
cedricpinson: Would it be simpler to create a function hasShadowTexture that does the loop ... that would do something like that:
if (this.hasShadowTexture)
     this._vertexShader.push( 'WorldPosition = ModelWorldMatrix * vec4(Vertex,1.0);' );

it seems easier to read and maintain.

sources/osgShader/Compiler.js
Kuranes: true, refactored code accordingly and more readabilidability with quick exits and less nested if/then

sources/osgShader/Compiler.js
Kuranes: seems risky business to have a global you *MUST* remember to reset when reusing the compiler ?
Some loops are needed anyway (gathering light/shadowtexture/shadowattribute soft linkages)

sources/osgShader/Compiler.js
```diff
@@ -1008,26 +993,12 @@ define( [
...
 
-                // varyings
-                var shadowVertProj = shadowTexture.getVaryingName( 'VertexProjected' );
-
-                this._vertexShader.push( 'vec4 shadowPos' + i + ' = ' + shadowView + ' *  worldPosition;' );
-                this._vertexShader.push( ' ' + shadowVertProj + ' = ' + shadowProj + ' * shadowPos' + i + ';' );
-                // varying packing using 1 vec4 fo both Projection vector & viewworld z pos
-                // and pre-linearize Z
-                this._vertexShader.push( ' ' + shadowVertProj + '.z  = shadowPos' + i + '.z ;' );
+                hasShadows = true;
```
cedricpinson: hasShadows seems useless here

Kuranes: Only one Left Alive

Removed all useless.

PR #317, opened 2015-05-21
Clean promise

stephomi:
For code review


stephomi:
@jtorresfabra Lots of change so you might be interested by trying out this branch :D.

The parsing should be faster for big osgjs file.
On some heavy graph we go from 28s down to 3.8s and the memory usage 830Mb down to 370Mb.

Optimizations includes : 
- Tries to avoid useless function creation (for example the check function but also the ones created before a loop)
- removes useless defer creation (for example this [one]
- All the parsing function should return a promise (that way we don't have to guess and do Q(something) all the time)
(https://github.com/cedricpinson/osgjs/blob/develop/sources/osgWrappers/serializers/osg.js#L48))
- Replaces Q by bluebird : Bluebird is faster than Q and it looks like that bluebird is faster than native promise (I didn't test it on bluebird though) so for now we stick with it. But with this PR the change in api management makes it very easy to change the promise lib.

jtorresfabra:
Hi @stephomi, it looks really promising! I'm going to check it out right now :)

Kuranes:
No promises it won't break your code :)

jtorresfabra:
@stephomi WOW!! Very big performance gain! Great job! I just checked out with one of my apps and it really rocks. I'm eagerly awaiting this PR to be merged. Thanks for advicing me.

PD: @Kuranes ;)

stephomi:
@jtorresfabra Is it fine if we merge it ? Did you encounter any troubles?

jtorresfabra:
@stephomi  Umm I'm still testing, anyway I think we can merge it. But as it is a big change and a lot of files involved, could we make a dev-release first? cc: @cedricpinson

cedricpinson:
@jtorresfabra I did a pre release with tag 0.2.2

PR #320, opened 2015-05-29
Added option to set DatabasePath manually for PagedLOD structures

jtorresfabra:


jtorresfabra:
It's weird as the test failed because:

Warning: The following files are not beautified:
sources/osgViewer/View.js
 
And that file is not modified in the PR

jtorresfabra:
I just updated the PR to develop, it should be enough.

jtorresfabra:
I can't understand what is failing, I'm closing this PR and resending it.

cedricpinson:
Ok it's normal because I merged something that check the coding style about indent ... if you need to fix you can use grunt beautify it will format the code as it should.
There are also tools for sublime/emacs/... that can do it inline while coding

PR #321, opened 2015-05-30
Fixed bug in Geometry.releaseGLObjects, stateset should exists to try…

jtorresfabra:
… to release it


jtorresfabra:
That code is older than the PR itself. Anyway it's fixed now.

jtorresfabra:
Is there any problem with this PR?

PR #322, opened 2015-05-30
Added flushAllDeletedGLObjects to be able to remove all the GLObjects…

jtorresfabra:
… without waiting next frames. It should be called before viewer destruction


hishamabdelrahman:
thanx hording. ..that was so much needed...

hishamabdelrahman:
I meant thanks Jordi

PR #324, opened 2015-06-01
Added option to set DatabasePath manually for PagedLOD structures

jtorresfabra:


jtorresfabra:
It seems something went wrong with travis in the previous PR, now it works ok,

PR #330, opened 2015-06-05
Bugfix/imagestream promise

Kuranes:
Fix bad promise callback binding 
Fix linter warning on Notify js
Fix two Examples ( hdr, cubemap ) missing gunzip vendor include


sources/osg/Notify.js
```diff
@@ -48,6 +48,7 @@ define( [], function () {
 
         return callstack;
     }
+
```
Kuranes: Wasn't passing the grunt check js-beautify without it...

PR #331, opened 2015-06-06
Adds FullscreenFakeQuad for faster postprocessing

Kuranes:
* Much less Fragment compute per pass ( no change needed in any shader)
* Much less state change (quad is a singleton, reused whatever framebuffer size it is)

### DIFF
--- a/sources/osg/Shape.js
+++ b/sources/osg/Shape.js
@@ -318,6 +318,35 @@ define( [
         return g;
     };
 
+    // better perf
+    // no more pixel shader hurt for nothing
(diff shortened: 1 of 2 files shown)


cedricpinson:
It seems very intersting, I would be really curious about the performance improvement. Could you report stats before/after to have an idea ?

Kuranes:
Only benchamarked on mobile:
On osgjs, a 3fps  (14 to 17fps) difference on basic  tonemapping postprocess example in favor of new technique 
Have yet to test on more complex scenario, as it should be better with multiple chained postprocess  + a background "quad".

cedricpinson:
Still it s interesting

Issue #226, opened 2015-01-09
Vertex Shader Graph


Kuranes:
Current vertex shader Compiler is concat based whereas node system would make it more robust and allow easier flexibility and better decoupling

PR #333, opened 2015-06-11
Makes vertex shader composition use the shader graph compiler

Kuranes:
Adds Vertex Shader Graph nodes and compilation
Adds Multiple Root Graph to Shader Graph Compiler (both VS and FS)
Optimizes Vertex Shaders Generation upon data (reducing uneeded vertex attributes and varying usage and computations)
Adds New Operation Nodes : MatrixMultPosition, MatrixMultDirection
Adds News "Root" Nodes: glPosition, glPointSize
Adds New Data Nodes: "Constant"
Adds Alias Nodes: SetFromNode (alias to Add Node but with One input, one output)

### DIFF
--- a/sources/osg/Material.js
+++ b/sources/osg/Material.js
@@ -15,7 +15,6 @@ define( [
         this._specular = [ 0.0, 0.0, 0.0, 1.0 ];
         this._emission = [ 0.0, 0.0, 0.0, 1.0 ];
         this._shininess = 12.5;
-        this._shadeless = false;
     };
 
     Material.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( StateAttribute.prototype, {
@@ -95,6 +94,7 @@ define( [
         getShininess: function () {
             return this._shininess;
         },
+
 
 
         setTransparency: function ( a ) {
--- a/sources/osgShader/Compiler.js
+++ b/sources/osgShader/Compiler.js
@@ -12,6 +12,7 @@ define( [
         this._textureAttributes = textureAttributes;
 
         this._variables = {};
+        this._varyings = {};
         this._vertexShader = [];
         this._fragmentShader = [];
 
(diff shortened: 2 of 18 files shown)


sources/osg/Material.js
```diff
@@ -96,6 +96,15 @@ define( [
             return this._shininess;
         },
 
+        setShadeless: function ( a ) {
```
cedricpinson: Not sure I want this, material is supposed to be like material of opengl. could you explain the link / logic behind this. I guess it's because of shader vertex to remove some attributes, but shadeless could be detected with no light in state during the shagerGenerator invocation ?

Kuranes: Just added get/set over existing _shadeless
Was indeed surprised to find the _shadeless member in Material here....

cedricpinson: Could you dig if we can remove it and how it appeared in material ?

Kuranes: Can do, nothing block that change.

sources/osgShader/Compiler.js
```diff
@@ -1165,6 +1477,7 @@ define( [
 
             var fragColor = factory.getNode( 'FragColor' );
```
cedricpinson: if we name gl variable with the prefix, maybe FragColor should follow the same pattern. Or we can remove gl prefix for other node, but it's better to make it consistent.

Kuranes: ok. I'm for the "gl" prefix (as "Position" node would be misleading)

cedricpinson: agree. Could you fix frag color and other if there are more ?

Kuranes: ok so we go for "glFragColor" as a node name (breaking change) ?

Kuranes:
closes https://github.com/cedricpinson/osgjs/issues/226

PR #336, opened 2015-06-11
Adds images mipmap support and compressed textures support

stephomi:


sources/osgDB/ReaderParser.js
```diff
@@ -30,6 +30,10 @@ define( [
     };
     ReaderParser.readImageURL = ReaderParser.readImage; // alias
 
+    ReaderParser.readBinaryArrayURL = function ( url, options ) {
```
cedricpinson: why adding this method ?

PR #337, opened 2015-06-12
Fixed a long standing bug in the camera projection matrix, it should …

jtorresfabra:
…be updated each time the  near/far changes in the renderer. Now intersectors works well in far distances.

### DIFF
--- a/sources/osgViewer/Renderer.js
+++ b/sources/osgViewer/Renderer.js
@@ -175,7 +175,8 @@ define( [
 
                 // store complete frustum
                 camera.setNearFar( this._cullVisitor._computedNear, this._cullVisitor._computedFar );
-
+                // Update near/far in the camera projection matrix
+                Matrix.clampProjectionMatrix( camera.getProjectionMatrix(), this._cullVisitor._computedNear, this._cullVisitor._computedFar, this._cullVisitor._nearFarRatio );
                 // restore previous state of the camera
                 this._cullVisitor.setCullSettings( previousCullsettings );
 


jtorresfabra:
@hishamabdelrahman it should fix the error you experienced when trying to pick from to far away. 
I was experiencing too this error.

jtorresfabra:
Probably we should apply this fix in CullVisitor too. Once the setNearFar is done: https://github.com/cedricpinson/osgjs/blob/develop/sources/osg/CullVisitor.js#L348

@Kuranes thoughts?

Kuranes:
the popCameraModelViewProjectionMatrix just above is supposed to do that already by calling the [PopProjectionMatrix](https://github.com/cedricpinson/osgjs/blob/08e095f7032d6c30865e6dfb6594091bed26d251/sources/osg/CullVisitor.js#L195).
It doesn't in your case  ?

jtorresfabra:
For my case (IntersectorVisitor) it inherits from NodeVisitor, so no clamp is done, it needs to get the matrix from camera, so the clamp should be done in the renderer. (If I recall well OSG does it also).

jtorresfabra:
@Kuranes Ok I see now that it indeed applies the clamp, but the _computedFar value is wrong.

jtorresfabra:
@Kuranes First camera never gets traversed by the CullVisitor so it never reaches that:
https://github.com/cedricpinson/osgjs/blob/08e095f7032d6c30865e6dfb6594091bed26d251/sources/osg/CullVisitor.js#L336

And the clamp is not done. 

So it makes sense to do it in Renderer

Kuranes:
Ha, yes, that first camera oddity not going through the same code as the rest. Bug prone.
Could be added on that code part, in View code

Seems osg does that too
https://github.com/openscenegraph/osg/blob/77d4847753cc28eb9c0006c4788ac6f0f7c0f273/include/osgUtil/CullVisitor#L254

Cannot find OSG renderer clamping projection matrix ?

jtorresfabra:
Yep, not in Renderer. It is done in SceneView

hishamabdelrahman:
@Jordi. ...thanks Jordi I was thinking that this bug can't be fixed so I was using polytope inter sector at far distance .but your fix is much neater way
thanks

cedricpinson:
@jtorresfabra could you add a link to the code in osg ? I dont remember this part.
thx a lot for the PR

cedricpinson:
The clamp is done in cullVisitor in popProjection and later in sceneview for OSG. SceneView is deprecated so I did not reproduce this class, instead I put the 'similar' code in Renderer.js . It seems the right way to do it @jtorresfabra 
Thank you very much it's a kind of tricky :+1:

Jordi:
I don't know how I got here, but I want to be out of this emails.

Thanks!

jtorresfabra:
Hi @Jordi, 
another user ( @hishamabdelrahman ) used your username instead of mine (@jtorresfabra). You should not get more messages again as this thread is closed. 

Sorry for the inconvenience.

PR #338, opened 2015-06-12
Replaced intersection ray/sphere method by a faster one

jtorresfabra:


jtorresfabra:
Hi all, 
Can you please test this at your end after #337 is merged? 
It should work as before (with the older method) without the merge anyway. 

cc: @stephomi @cedricpinson

stephomi:
Works fine for me.
Nice !
If I'm not wrong, without further instructions you can add a positive test before the ``s`` computation.

```
Vec3.sub( bsphere.center(), this._iStart, l );
var l2 = Vec3.length2( l );
if( l2 < bsphere.radius2()) return true;
var s = Vec3.dot( l, this._iEnd );
if ( s < 0 ) return false;
var m2 = l2 - ( s * s );
if ( m2 > bsphere.radius2() ) return false;
```

sources/osgUtil/LineSegmentIntersector.js
```diff
@@ -38,41 +38,18 @@ define( [
...
-                    return false;
-                }
-
-                if ( r1 >= 1.0 && r2 >= 1.0 ) {
-                    return false;
-                }
+                Vec3.sub( bsphere.center(), this._iStart, l );
+                var s = Vec3.dot( l, this._iEnd );
+                var l2 = Vec3.length2( l );
+                if ( s < 0 && l2 > bsphere.radius2() ) return false;
```
cedricpinson: use 0.0 instead of 0
it's possible the jit needs to convert type

jtorresfabra:
@cedricpinson  Fixed

jtorresfabra:
@stephomi  not sure about that, what do you mean exactly? BTW I picked the optimized algorithm from the Real-Time rendering book, third edition.

jtorresfabra:
Argh, travis is failing with no reason again. I'm closing this PR and doing a new one.

stephomi:
@jtorresfabra well if ``l2 >= bsphere.radius2()``, it means the there's an intersection. So we can return directly true (and this test is done afterwards anyway.

jtorresfabra:
@stephomi ok now I get you, s is in the case your are inside the bounding sphere so that is not necessary. Isn't it?

jtorresfabra:
@stephomi ok, no if (s<0) and  the ray origin is outside the sphere it means that the sphere is behind
the ray origin so we can return false. But agree with your modifications, them makes sense. I'm commiting them

stephomi:
Yup this particular test (``_iStart`` inside sphere) was already done in the last slow function actually.
Don't know if it should be ``<`` or ``<=`` though... :)

jtorresfabra:
Ok so  if l2 <= bsphere.radius2() it is inside or already touching the sphere, totally agree. Modificaction commited.

jtorresfabra:
It's so awkward that travis fails without reason.

sources/osgUtil/LineSegmentIntersector.js
```diff
@@ -38,41 +38,19 @@ define( [
...
-                }
-
-                if ( r1 >= 1.0 && r2 >= 1.0 ) {
-                    return false;
-                }
+                Vec3.sub( bsphere.center(), this._iStart, l );
+                var l2 = Vec3.length2( l );
+                if ( l2 <= bsphere.radius2() ) return true;
+                var s = Vec3.dot( l, this._iEnd );
+                if ( s < 0 ) return false;
```
cedricpinson: it's better to use float when you test float, like 0.0 instead of just 0. It helps the jit, like not converting type on the fly

jtorresfabra: Ops, I thought I had fixed it. Sorry about that.

jtorresfabra: I see: I fixed it and then introduced the bug again, shame on me... fixed again.

PR #344, opened 2015-06-18
Improves bounding box transform function

stephomi:
Didn't see any differences with the performance examples.


jtorresfabra:
I just tested and DrawTime falls with this PR. See https://github.com/cedricpinson/osgjs/pull/233. 
Just enable stats in develop branch vs your branch and see drawtime.

stephomi:
I remove the two "remove evil" commits. Now the PR is just about something else :).

PR #348, opened 2015-06-22
Added LODScacale, revamped the pagedLOD example. CullStack now inheri…

jtorresfabra:
…ts from CullSettings


cedricpinson:
That would be cool to have basic unittest at least to see if there are some regression linked to pagelod. Do you have some on you apps that could be transfered to osgjs ?

jtorresfabra:
It is kind of hard to test DatabasePager and/or PagedLOD in unitary tests, but I just coded a visual test in the PagedLOD example in this PR. Did you tried it?

PR #360, opened 2015-06-30
Makes shadow casting use shader graph compiler

Kuranes:


sources/osgShader/Compiler.js
```diff
@@ -106,15 +106,22 @@ define( [
...
+                } else if ( type === 'ShadowCastAttribute' ) {
+                    this._isShadowCast = attributes[ i ].isEnabled();
+                    this._shadowCastAttribute = attributes[ i ];
                 }
             }
         },
 
         initTextureAttributes: function () {
+
+            if ( this._isShadowCast )
```
cedricpinson: just to be sure it means we dont handle blending transparency in shadow right ?

sources/osgShader/Compiler.js
```diff
@@ -412,7 +423,7 @@ define( [
 
 
         getOrCreateInputPosition: function () {
-            return this.getOrCreateVarying( 'vec3', 'FragEyeVector' );
+            return this.getOrCreateVarying( 'vec4', 'FragEyeVector' );
```
cedricpinson: why we switched to a vec4 ?

sources/osgShader/Compiler.js
```diff
@@ -697,7 +710,7 @@ define( [
 
             }
             // TODO: shadow Attributes in node, is this the legit way
```
cedricpinson: That's the point it's a big subject to discuss, not easy to keep the this part flexible, because it almost needs to know all externals nodeType or at least big sub system to handle the logic here.
Problably we would need to debrief about this point, that's a very important one.

sources/osgShader/Compiler.js
```diff
@@ -1020,19 +1033,12 @@ define( [
             } );
 
         },
-        declareVertexTransformLighted: function ( glPosition ) {
 
+        declareTransformWithEyeSpace: function ( glPosition ) {
```
cedricpinson: could comments a lot what it is the space / math for the docs

sources/osgShader/Compiler.js
```diff
@@ -1040,20 +1046,26 @@ define( [
...
-            //glpos
+            // glposition
             factory.getNode( 'MatrixMultPosition' ).inputs( {
                 matrix: this.getOrCreateUniform( 'mat4', 'ProjectionMatrix' ),
                 vec: tempViewSpace
             } ).outputs( {
                 vec: glPosition
             } );
+        },
+        declareVertexTransformLighted: function ( glPosition ) {
```
cedricpinson: same describe with details what is it

sources/osgShader/Compiler.js
```diff
@@ -1288,15 +1305,32 @@ define( [
             return fragCol;
         },
 
+        createShadowCastFragmentShaderGraph: function () {
```
cedricpinson: describe in details what it does

sources/osgShader/Compiler.js
```diff
@@ -1288,15 +1305,32 @@ define( [
...
         // this is the main function that will generate the
         // fragment shader. If you need to improve / add your own
         // you could inherit and override this function
         createFragmentShaderGraph: function () {
 
-
             var roots = [];
 
+            // depth cast
+            if ( this._isShadowCast ) {
```
cedricpinson: a lot of logic is inside the function instead would it be possible to use instead the StateSet ShaderGeneratorName to hook the compiler in the graph ?
This way we could inherit for example the shadow compiler and it would directly replace createFragmentShaderGraph ?
It's just an idea if we could avoid to overload the internal logic

sources/osgShadow/ShadowMap.js
```diff
@@ -63,11 +65,15 @@ define( [
         apply: function ( node ) {
             var st = node.getStateSet();
             // check that and other things ?
+            // TODO: should check whole hierarchy in fact...
             if ( st ) {
                 var blend = st.getAttribute( 'BlendFunc' );
                 if ( blend !== undefined && blend.getSource() !== BlendFunc.DISABLE ) {
```
cedricpinson: explains in details the logic plz

cedricpinson:
Very interesting and that raises some questions that we need to discuss

sources/osgShader/Compiler.js
Kuranes: we don't handle alpha blending

sources/osgShader/Compiler.js
Kuranes: Check with @stephomi on why

sources/osgShadow/ShadowMap.js
Kuranes: It's either try to detect it, however imperfectly like here or it's user work to specify shadowcast node mask by himself on all it's alpha blended geoms

sources/osgShader/Compiler.js
Kuranes: need .W to linearize depth when rendering depth into a buffer

sources/osgShader/Compiler.js
cedricpinson: add it as comments to help autogenerated docs

sources/osgShader/Compiler.js
cedricpinson: adds it in comments for inline auto docs

sources/osgShader/Compiler.js
cedricpinson: adds comments

sources/osgShader/Compiler.js
cedricpinson: @stephomi could you explains ^

sources/osgShadow/ShadowMap.js
cedricpinson: probably it's better to only rely on users and document how to control it

cedricpinson:
@stephomi @Kuranes we should talk to be sync on this

sources/osgShadow/ShadowMap.js
```diff
@@ -60,20 +62,29 @@ define( [
...
             /*jshint bitwise: true */
         },
+        // Visiting whole casting scene
         apply: function ( node ) {
             var st = node.getStateSet();
             // check that and other things ?
+            // TODO: should check whole hierarchy in fact...
             if ( st ) {
+                // check for transparency not casting shadows
+                // as no alpha blending transparency shadow (no transmittance support)
```
cedricpinson: that would be could to keep this logic into function, if the user wants to override this logic he could

cedricpinson:
Good for me, adds a bit more of documentation, and move the function that check the cast with dept/blend

PR #362, opened 2015-07-01
Misc fixes: removed unnecessary code in Geometry and normalize norm…

jtorresfabra:
…als for Shape.createTexturedQuad. Two tiny fixes....


jtorresfabra:
why what? :)

cedricpinson:
@jtorresfabra probably it will break some example, what do we have as options ? a second function to build with normals ?

stephomi:
@jtorresfabra ok then if we don't provide the normals,  then we can comment all the normals block (l.401-418)

jtorresfabra:
Oh my god the intended change eas to normalize te normals, not to remove
it! Sorry about the mistake. I'll fix it later!

El miércoles, 1 de julio de 2015, Stéphane <notifications@github.com>
escribió:

> @jtorresfabra <https://github.com/jtorresfabra> ok then if we don't
> provide the normals, then we can comment all the normals block (l.401-418)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cedricpinson/osgjs/pull/362#issuecomment-117725763>.
>


-- 
Jordi Torres

jtorresfabra:
FIxed... that's the intended submission :sweat_smile:

PR #367, opened 2015-07-02
Adds Animation to shader vertex graph

Kuranes:
- Optimizes Heavily shader code, reusable BoneMatrix result transform
- Adds Size uniform parameter for Array of uniform from Compiler/Node/shaderGraph
- Sort Shader Attributes inside the DeclareGlobal Compiler code (and beautifies)

### DIFF
--- a/sources/osgAnimation/RigTransformHardware.js
+++ b/sources/osgAnimation/RigTransformHardware.js
@@ -1,122 +1,24 @@
 define( [
+
     'osg/Utils',
     'osg/NodeVisitor',
     'osg/Notify',
     'osg/Matrix',
     'osg/Program',
     'osg/Shader',
+    'osg/StateAttribute',
     'osg/Uniform',
+    'osgAnimation/AnimationAttribute',
     'osgAnimation/CollectBoneVisitor'
-], function ( MACROUTILS, NodeVisitor, Notify, Matrix, Program, Shader, Uniform, CollectBoneVisitor ) {
+
+], function ( MACROUTILS, NodeVisitor, Notify, Matrix, Program, Shader, StateAttribute, Uniform, AnimationAttribute, CollectBoneVisitor ) {
 
     'use strict';
-
-
-    function getShader( nbVec4 ) {
-        var vertexshader = [
-            '',
-            '#ifdef GL_ES',
-            'precision highp float;',
-            '#endif',
-            'attribute vec3 Vertex;',
-            'attribute vec3 Normal;',
-            'attribute vec4 Bones;',
-            'attribute vec4 Weights;',
-
-            'uniform mat4 ModelViewMatrix;',
-            'uniform mat4 ProjectionMatrix;',
-            'uniform mat4 NormalMatrix;',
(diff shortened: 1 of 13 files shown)


sources/osg/Uniform.js
```diff
@@ -43,14 +43,16 @@ define( [
         },
         apply: function ( gl, location ) {
             if ( this._dirty ) {
-                this.update.call( this.glData, this.data );
+                if ( this.update )
```
cedricpinson: why do you need to test update ?

Kuranes: because when it's a type array there is no update

sources/osg/Uniform.js
```diff
@@ -406,8 +430,10 @@ define( [
         };
         uniform.apply = uniform.applyMatrix;
         uniform.transpose = false;
-        uniform.glData = new MACROUTILS.Float32Array( uniform.data );
-        uniform.update = Uniform.prototype._updateFloat16;
+
+        var isTypedArray = Object.prototype.toString.call( value ) === '[object Float32Array]';
```
stephomi: value instanceof Float32Array ?

sources/osgShader/Compiler.js
```diff
@@ -945,6 +955,34 @@ define( [
...
+            // Attribute 0 Must Be vertex
+            // perf warning in console otherwiser in opengl Desktop
+            if ( func._text.length ) {
+                // sort in alphabetical order
+                // attr, unif, sample, varying
+                func._text.sort();
+                // now sort Attributes
+                // making sure Vertex is always coming first
+                var toShift = [];
+                for ( j = 0; j < func._text.length && func._text[ 0 ].indexOf( 'Vertex' ) === -1; j++ ) {
```
cedricpinson: it's hard to read 
j < func._text.length && func._text[ 0 ].indexOf( 'Vertex' )
could simplify the logic or split into multiline

sources/osgAnimation/RigTransformHardware.js
```diff
@@ -15,84 +19,74 @@ define( [
     function getShader( nbVec4 ) {
```
cedricpinson: Do we still to keep this shader ?

Kuranes: is for debug purpose ? 
To remove when  validated the remove of the "timeout" thingy
(as I couldn't reproduce here the bug that made that timeout necessary)

sources/osgShader/Compiler.js
Kuranes: Can add a "\n" after the "&&"  ?
Or you want to move a test inside the loop with a break ?

cedricpinson: it's better out of the loop

sources/osgAnimation/RigTransformHardware.js
cedricpinson: well it was as first implementation to test without shader. If we dont need it anymore, we can remove it.

PR #375, opened 2015-07-14
Adds functions to set imageCrossOrigin and imageLoadingUsePromise options

hishamabdelrahman:


jtorresfabra:
you only need to do 'grunt beautify' to beautify the sources. My recommendation is to close this PR and make another one with only one clean commit ( there are 4 messy commits now ). Also you don't need to change header in Input, why you do this? Review the lines you are changing before doing the PR.

hishamabdelrahman:
will redo it again

PR #376, opened 2015-07-14
adds Text2D feature, using a canvas2D

jtorresfabra:
@cedricpinson  you can flag it as Prototype. 	Even though it is working so far.


sources/osgText/Text.js
```diff
@@ -0,0 +1,269 @@
...
+                this._textX = this._canvas.width;
+                this._context.textBaseline = 'bottom';
+                this._textY = this._canvas.height;
+                break;
+            }
+        },
+
+        _nextPowerOfTwo: function ( value, pow ) {
+            pow = pow || 1;
+            while ( pow < value ) {
```
cedricpinson: probably there is a formula for this :)

cedricpinson:
Very cool to have osgText :+1: :)

jtorresfabra:
@cedricpinson  I think it's ready to be reviewed.

jtorresfabra:
I just added the 'Attribute' Suffix because in OSG the class Billboard exist, but it is derived from Geode. I mean, it is not an attribute in OSG. I thought about implement the Billboard Class using Billboard attribute later. 

Also I'm unsure about adding it to osgText, because Billboard is in osg library. 

What do you think?

cedricpinson:
you are right, so did you try with something else like models instead of just textured quad

jtorresfabra:
It does work with models too, but implementing the complete Billboard class would need some work, as implementing pivot points, axis, etc.

PR #379, opened 2015-07-16
Fixes and optimizes interpolator and animations serializer

gfaucill:


cedricpinson:
could you check why test fails

gfaucill:
It's because Stéphane merge/delete some function/field like (findAnimationUpdateCallback, _animationsUpdateCallback). So the tests can't call this functions.

PR #382, opened 2015-07-16
Adds webpack loader to strip debug code from build

cedricpinson:
Now code between
```
 /* develblock:start */
 your code
 /* develblock:end */
```

will be strip from builds except in 'grunt build-debug'

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -99,6 +99,7 @@ var gruntTasks = {};
 ( function () {
 
     var webpack = require( 'webpack' );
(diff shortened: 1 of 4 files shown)


Gruntfile.js
```diff
@@ -99,6 +99,7 @@ var gruntTasks = {};
 ( function () {
 
     var webpack = require( 'webpack' );
+    // var ExtractTextPlugin = require( 'extract-text-webpack-plugin' );
```
pierreant-p: :)

cedricpinson: oops

PR #383, opened 2015-07-17
Makes mandatory to give a lightsource to shadowmap ctor 

Kuranes:
So that shadowreceiveattribute ctor has a light number at creation.
Otherwise shadowreceiveattribute  may  be cloned without any "light" and then if using multiple shadowreceiveattribute, you'll have mutiple cloned default shadowreceiveattribute with typeMember !== key in the State.js AttributeMap

/!\ Breaking Change for all shadowmap user (lightSource of settings given to ShadowMap need a lightSource )

### DIFF
--- a/sources/osgShadow/ShadowReceiveAttribute.js
+++ b/sources/osgShadow/ShadowReceiveAttribute.js
@@ -17,25 +17,13 @@ define( [
      * @class ShadowReceiveAttribute
      * @inherits StateAttribute
      */
-    var ShadowReceiveAttribute = function ( lightNumber, disable ) {
+    var ShadowReceiveAttribute = function ( lightNum, disable ) {
         StateAttribute.call( this );
 
+        this._lightNumber = lightNum;
 
-        //this._lightNumber // (might be broken by light reordering change) // yeah agree
-
-        // just in case
-        // the object in case of change in ordering ?
-        // need to be able to link in compiler the light uniforms as we use light uniforms in shadow shader code
-        // well it's not supposed to be a problem, but yeah the ordering is an issue, so maybe it's better to link to
-        // the light attribute, in this case we dont need anymore the _lightNumber
-        // but re ordering light seems to not be a good practice in osg, it's still a risk
-        this._light = undefined;
-        // that could be more generic to have the unit instead of the object
 
         // see shadowSettings.js header for shadow algo param explanations
-
-        this._lightNumber = lightNumber !== undefined ? lightNumber : 0;
-
         // hash change var
         this._algoType = 'NONE';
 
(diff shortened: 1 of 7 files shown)


cedricpinson:
What's happen if the light attribute change the lightnum ?

Kuranes:
Same as before we remove the receiving stateset and add it once the light is changed in the attribute, check here: 
https://github.com/cedricpinson/osgjs/blob/bugfix/shadow-receive-attribute-initial-state/sources/osgShadow/ShadowMap.js#L383

cedricpinson:
Probably it could be smart to add safe guard with develblock and assert just to avoid to forget if it happens

Kuranes:
New code doesn't work at all without giving light to the ctor now.
Could add an assert, but the error/segfault it raises are pretty explanatory

cedricpinson:
no I mean if during the execution it changes that would be good add an assert to be sure the code that handle the change of lightnum does not fail

Kuranes:
added assert

cedricpinson:
not sure you need to use lightsource or light instead of lightnum. It works the same, and passing an object instead of lightnum does not guarranty the lightnum will not change so I dont see advantage to use an object instead of lightnum. I prefere to keep lightnum to avoid indirection information in object.

Kuranes:
LightSource is not be needed anymore ( now that we use cullvisitor "nodetrace" to get light matrix, but light )
it can be removed too from shadowreceiveatttribute to just lightnumber

Whereas in ShadowMap, Light does help getting discrepancy between light->lightnumber and the lightnumber we use/have in shadow map and its attribute, otherwise cannot detect a change [here](https://github.com/cedricpinson/osgjs/blob/bugfix/shadow-receive-attribute-initial-state/sources/osgShadow/ShadowMap.js#L383)

cedricpinson:
Yeah make sense for https://github.com/cedricpinson/osgjs/blob/bugfix/shadow-receive-attribute-initial-state/sources/osgShadow/ShadowMap.js#L383 yeah it would be good to check it with develblock at each frame

sources/osgShadow/ShadowReceiveAttribute.js
```diff
@@ -65,7 +69,7 @@ define( [
         attributeType: 'ShadowReceive',
 
         cloneType: function () {
-            return new ShadowReceiveAttribute( this._lightNumber, true );
+            return new ShadowReceiveAttribute( this._light, true );
```
cedricpinson: btw if lightnum change it will be detected by the check in State.js so maybe the light here does not add value for error checking

cedricpinson:
does this PR needs to update code on sketchfab ?

Kuranes:
Yes, need a one line change with it. from new shadow () to new shadow(shadowsettings)

sources/osgShadow/ShadowReceiveAttribute.js
Kuranes: you sure it's debug code ?
I tought of that as real code, when user change lightnum of its light ?

PR #384, opened 2015-07-20
Fix Animationttribute clone Type being different from original

Kuranes:


sources/osgAnimation/AnimationAttribute.js
```diff
@@ -11,28 +11,34 @@ define( [
...
      * @inherits StateAttribute
      */
-    var AnimationAttribute = function ( disable ) {
+    var AnimationAttribute = function ( boneSize, animationID, disable ) {
         StateAttribute.call( this );
+        this._boneSize = boneSize;
         this._enable = !disable;
+
+        this._animationID = animationID;
+        if ( !animationID ) AnimationAttribute.getAnimationID();
```
cedricpinson: You can use instanceID instead of creating a new id system for a static id

sources/osgAnimation/RigTransformHardware.js
```diff
@@ -86,10 +86,10 @@ define( [
 
             var st = geom.getOrCreateStateSet();
 
-            var animAttrib = new AnimationAttribute();
-            st.setAttributeAndModes( animAttrib, StateAttribute.ON );
+
+            var animAttrib = new AnimationAttribute( nbVec4Uniforms );
+            st.setAttributeAndModes( animAttrib, StateAttribute.ON | StateAttribute.OVERRIDE );
```
cedricpinson: why override appeared ?

sources/osgAnimation/AnimationAttribute.js
Kuranes: mmm was afraid of unexpected side effect if two object share the same instanceID 
(clone & original would have the same )

sources/osgAnimation/RigTransformHardware.js
Kuranes: my bad, slipped on test session on stateattribute flags. will remove

sources/osgAnimation/AnimationAttribute.js
cedricpinson: check the code of instanceID
instanceID is unique, you can do 
if ( animationID === undefined )
    animationID = this.getInstanceID()

Kuranes: nice reuse trick
Indeed will reuse instanceID like that

cedricpinson:
Needs a rebase

Kuranes:
rebased

PR #385, opened 2015-07-20
Adds More checks on node request/usage

Kuranes:
Adds More checks on node request/usage
Adds varyings check between shader stage
Adds Comments
Adds new Kind of variables: Outputs (glFragColor, glPointsize, glPosition)

### DIFF
--- a/sources/osgShader/Compiler.js
+++ b/sources/osgShader/Compiler.js
@@ -11,6 +11,9 @@ define( [
         this._attributes = attributes;
         this._textureAttributes = textureAttributes;
 
+        this._activeNodeList = {};
+        this._compiledNodeList = {};
+
         this._variables = {};
         this._varyings = {};
         this._vertexShader = [];
@@ -183,11 +186,45 @@ define( [
                 }
             }
         },
+
         // global accessor because it modifies
         // globally the compiler behavbiour
         isShadowCast: function () {
             return this._isShadowCast;
         },
+
+        // cache all requested node, so that we can list
+        // and log unused Node that where called
+        // or/cache unique or predefined node
+        // thus avoid mutiple getNode of a
+        // Node that HAS to be unique
+        getNode: function ( /*name, arg1, etc*/) {
+
+            // check unique Node
+            // for predefined GL variables
+            // gl_FragCoord, gl_Position, etc
+            // Extend to Varying
+            var n = factory.getNode.apply( factory, arguments );
+            var cacheID = n.getID();
+            this._activeNodeList[ cacheID ] = n;
+            return n;
+        },
+        // during compilation we pop
+        // all node we do encounter
+        // so that we can warn about
+        // "leftover" once compilation
+        // is finished
+        // Note: same node may be marked multiple time
+        // do not use it as a "once and for all mark thing"
+        markNodeAsVisited: function ( n ) {
+            var cacheID = n.getID();
+            if ( this._activeNodeList[ cacheID ] === n ) {
+                this._compiledNodeList[ cacheID ] = n;
+            } else {
+                Notify.warn( 'Node not requested by using Compiler getNode and/or not registered in nodeFactory' );
+            }
+        },
+
         getVariable: function ( nameID ) {
             return this._variables[ nameID ];
         },
@@ -201,7 +238,6 @@ define( [
             return undefined;
 
         },
-
 
 
         // TODO: add Precision qualifier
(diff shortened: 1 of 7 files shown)


sources/osgShader/Compiler.js
```diff
@@ -183,11 +186,43 @@ define( [
...
+            var cacheID = n.getID();
+            this._activeNodeList[ cacheID ] = n;
+            return n;
+        },
+        // during compilation we pop
+        // all node we do encounter
+        // so that we can warn about
+        // "leftover" once compilation
+        // is finished
+        tapNodeAsCompiled: function ( n ) {
```
cedricpinson: is it to say markNodeAsUsed/Compiled ?

sources/osgShader/Compiler.js
```diff
@@ -853,6 +890,9 @@ define( [
                 }
             }
             functor.call( functor, node );
+
+            // keep trace we visited.
+            this.tapNodeAsCompiled( node );
```
cedricpinson: ok maybe markNodeAsVisited ?

tests/osgShader/Compiler.js
```diff
@@ -78,7 +78,7 @@ define( [
                     value.apply( instance );
                     var t = instance.getType();
 
-                    if ( instance._name === 'Variable' ) {
+                    if ( instance._name === 'Variable' || instance._unique ) {
```
cedricpinson: we should use accessors getName and isUnique

sources/osgShader/Compiler.js
Kuranes: yes, but as we may encounter multiple times the node (shadergraph is visited for extension, defines, globalfunc, ) I thought "mark" was a bit "definitive" whereas it a Node can be tapped several times

sources/osgShader/Compiler.js
Kuranes: markNodeAsVisitedAtLeastOnce ?

tests/osgShader/Compiler.js
Kuranes: k.

sources/osgShader/Compiler.js
```diff
@@ -183,11 +186,45 @@ define( [
...
+        // "leftover" once compilation
+        // is finished
+        // Note: same node may be marked multiple time
+        // do not use it as a "once and for all mark thing"
+        markNodeAsVisited: function ( n ) {
+            var cacheID = n.getID();
+            if ( this._activeNodeList[ cacheID ] === n ) {
+                this._compiledNodeList[ cacheID ] = n;
+            } else {
+                Notify.error( 'Node not requested by using Compiler getNode and/or not registered in nodeFactory' );
```
cedricpinson: maybe it's not an error but a warning ?

sources/osgShader/Compiler.js
```diff
@@ -1239,30 +1290,55 @@ define( [
...
+            // Process defines, add precision, resolve include pragma
             var shader = this._shaderProcessor.processShader( shaderStr, defines, extensions, type );
 
+            // Check
+            var compiledNodes = Object.keys( this._compiledNodeList );
+            var activeNodes = Object.keys( this._activeNodeList );
+            activeNodes.filter( function ( i ) {
+                var found = compiledNodes.indexOf( i ) !== -1;
+                if ( !found ) {
+                    Notify.log( 'Nodes Not Used: ' + i + ' ' + this._activeNodeList[ i ].getName() );
```
cedricpinson: log or warning ?

cedricpinson:
That will help for sure, warning / error / log, could you check what's more appropriate ?

sources/osgShader/Compiler.js
Kuranes: true, could be a warning.

sources/osgShader/Compiler.js
Kuranes: warning indeed.

Kuranes:
Agree, will make a check

Kuranes:
reworked log to warning + better warning message.
Maybe We should use develblock so that those only appear/compute in debug build ?

sources/osgShader/node/data.js
```diff
@@ -126,12 +130,19 @@ define( [
...
+    // Derive from that for new outputs
+    // gl_FragDepth, etc.
+    var Output = function ( type, wholeName ) {
+        Variable.call( this, type, wholeName );
     };
-    glFragColor.prototype = MACROUTILS.objectInherit( Variable.prototype, {
 
+    Output.prototype = MACROUTILS.objectInherit( Variable.prototype, {
+        _unique: true,
+        getUnique: function () {
```
cedricpinson: isUnique

PR #386, opened 2015-07-20
Adds a specific (per RigGeometry) animation stateset

stephomi:


cedricpinson:
needs to comments all the new method and the why

stephomi:
Done.

PR #395, opened 2015-07-31
Adds osgAnimation to osgjs (skinning and solid)

stephomi:


stephomi:
Many files are not used, maybe we should clean it.

cedricpinson:
You are right I will check to remove them

PR #407, opened 2015-08-08
Fix Typo on quat:nlerp shortest path branch variable

Kuranes:
Typo Bug spotted by @stephomi

### DIFF
--- a/sources/osg/Quat.js
+++ b/sources/osg/Quat.js
@@ -1,7 +1,8 @@
 define( [
     'osg/Vec3',
+    'osg/Vec4',
     'osg/Notify'
-], function ( Vec3, Notify ) {
+], function ( Vec3, Vec4, Notify ) {
 
     'use strict';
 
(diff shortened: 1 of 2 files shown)


cedricpinson:
At the same time could you add a test?

sources/osg/Quat.js
```diff
@@ -101,17 +101,25 @@ define( [
 
 
         // http://physicsforgames.blogspot.fr/2010/02/quaternions.html
+        // called quatBlend
+        //  (  quaternion nlerp is supposed to be commutative, constant velocity, torque minimal )
```
cedricpinson: http://number-none.com/product/Understanding%20Slerp,%20Then%20Not%20Using%20It/ tells is not commutive

Kuranes:
added test on nlerp and slerp.

See comment about reusing Vec4 code directly ?
(not inheriting, but reusing code for smaller js files)

cedricpinson:
Agree you can add reference to Vec4 method no need for duplicate code

Kuranes:
Done.

PR #408, opened 2015-08-09
Adds Target to handle defaultValue / value

cedricpinson:


sources/osgAnimation/BasicAnimationManager.js
```diff
@@ -160,30 +148,25 @@ define( [
...
 
+
         update: function ( node, nv ) {
 
             if ( this._dirty ) {
                 this._findAnimationUpdateCallback( node );
-                this._assignTargetToAnimationCallback();
-                this._dirty = false;
+                this._registerTargetFoundInAnimationCallback();
+                this._registerAnimations();
```
Kuranes: strange that the dirty  check is here (with 3 method call) but it's _registerAnimtions is the one that make dirty false ?

examples/animation/main.js
```diff
@@ -172,6 +172,7 @@
                 config.currentAnim = config.anim || firstAnimation;
                 if ( config.currentAnim ) {
                     controller.play = function () {
+                        animationManager.stopAllAnimation();
```
Kuranes: no "already playing" check to avoid restarting from 0 and/ord compute everything ?
(Or do it only if dirty or smth ?)

sources/osg/Vec3.js
```diff
@@ -132,6 +132,7 @@ define( [], function () {
     };
 
     Vec3.zero = [ 0.0, 0.0, 0.0 ];
+    Vec3.one = [ 1.0, 1.0, 1.0 ];
```
Kuranes: vec3.unit ?

sources/osgAnimation/BasicAnimationManager.js
```diff
@@ -135,19 +130,12 @@ define( [
...
-            var i;
-            for ( i = 0; i < this._targetIDByTypes.length; i++ )
-                this._targetIDByTypes[ i ].length = 0;
+            this._resetTargets();
 
             this._activeAnimations = {};
             this._activeAnimationList.length = 0;
 
-            for ( i = 0; i < this._activeChannelsByTypes; i++ )
+            for ( var i = 0; i < this._activeChannelsByTypes; i++ )
```
Kuranes: we're not consistent on array reset... perhaps we could make an array method so that everyone use the same 
(splice, length=0, = [])

sources/osgAnimation/BasicAnimationManager.js
```diff
@@ -160,30 +148,25 @@ define( [
...
-            // register targetID per type
-            for ( var i = 0; i < newTargetAdded.length; i++ ) {
-                var type = newTargetAdded[ i ].type;
-                this._targetID.push( createTargetID( i, ResultType[ type ].create() ) );
-            }
-
+            // qeue them to assign target
+            Array.prototype.push.apply( this._animationsToRegister, instanceAnimationList );
             this._dirty = true;
+            this._registerAnimations();
```
Kuranes: dirty = true and next line the register will update a false ?
the ::update code block under dirty== true will work nonetheless ?

sources/osgAnimation/BasicAnimationManager.js
```diff
@@ -345,13 +328,54 @@ define( [
...
+                var targetName = instanceChannel.channel.target;
+                var name = instanceChannel.channel.name;
+                var uniqueTargetName = targetName + '.' + name;
+
+                // disply a warning if animation has a channel but not target found in the
+                // scene graph. We could probably optimize and removes those channels, but
+                // it must be a user decision in case the user plugin different scene
+                // graph together and target would appear later in the scenegraph
+                if ( !this._targetMap[ uniqueTargetName ] ) {
+                    Notify.warn( 'registerInstanceAnimation did not find targetName ('+ uniqueTargetName + ') in the scene graph' );
```
Kuranes: "user decision":  do not get what's the path to solve the problem user side from comment ?
What does he need to do when he update scenegraph to make sure targetmap is filled and he can re-run registerInstanceAnimation ? (mesh LOD change for instance)

"probably optimize" : do not get what is the effect of not having a valid targetID here?
 (don't we want to put some special value in there btw)

sources/osgAnimation/BasicAnimationManager.js
```diff
@@ -429,9 +457,11 @@ define( [
                 instanceChannel.t = t; // reset time
                 instanceChannel.instanceAnimation = instanceAnimation; // link with parent animation
                 var targetID = instanceChannel.targetID;
+
+                if ( targetID === -1 ) continue;
```
Kuranes: use a "js define" to encapsulate the magic number ? 
Animation.TARGET_UNDEFINED or something.
(allow for better search, does document what is the value, etc)

sources/osgAnimation/BasicAnimationManager.js
```diff
@@ -13,14 +13,6 @@ define( [
...
 
-    var createTargetID = function ( id, value ) {
-        return {
-            id: id,
-            channels: [],
-            value: value
-        };
-    };
-
     var Float = {
```
Kuranes: no clear  ? is that a StackedFloat  ?
if yes, no defaultValue like others ?

sources/osgAnimation/BasicAnimationManager.js
cedricpinson: In order to register animations we need first to find all callback from the scene graph, So when registering animations we can't register the animation entierly, we reference it but after we need to assign what found on the scene to the animations's channels.
the _registerAnimtions make the dirty false because it registers the last part of the animations.

examples/animation/main.js
cedricpinson: stopAllAnimation, just reset animation playing. no dirty or recomputing is involve here

sources/osg/Vec3.js
cedricpinson: why unit ? just to know why it would be better

sources/osgAnimation/BasicAnimationManager.js
cedricpinson: not sure understand what do you mean ? reset array should be lenght = 0, but we keep some predefined array

sources/osgAnimation/BasicAnimationManager.js
cedricpinson: Yeah adding an animation tag the manager dirty, but registerAnimation can fix dirty to
false if we already has the upcateCallback collected. It's due to the fact that to
register animation we need to first collect data in the graph. So if not ready the update
will check if dirty and will fill collected data from the scene graph and be able to
finish to register animation

sources/osgAnimation/BasicAnimationManager.js
cedricpinson: Target are collected from the scenegraph, but for PageLOD node ( that could be loaded and
attached to a scenegraph at any moment ) or other system that could attach more scenegraph
element, it's possible the first time we parse the scene graph we dont get all targets needed by animations.
It means register animation will be partial, and some elements animated by the animation will not be present in the scene.
- So one solution is to say, the animation has channels without targets ( not found in the scene ). So we could clean and removes it instead of keeping because it's useless.
- or we could let them and let the user call dirty when he adds new nodes to force the manager to reparse the graph and assign targets to previous unseen targets
That's why it's a warning because it depends on your context.

sources/osgAnimation/BasicAnimationManager.js
cedricpinson: Make sense

sources/osgAnimation/BasicAnimationManager.js
cedricpinson: Float is just used as operator of have operation like copy/create/init to fit into the same code like Quat/Vec3/...

float are used to animate rotate angle on an axis or any other float value

sources/osgAnimation/BasicAnimationManager.js
stephomi: ```for ( var i = 0; i < this._activeChannelsByTypes; i++ )```
should be
```for ( var i = 0; i < this._activeChannelsByTypes.length; i++ )```
or better
```for ( var i = 0, nb = this._activeChannelsByTypes.length; i < nb; i++ )```

File: sources/osgAnimation/BasicAnimationManager.js (hunk 8 above)
Comment (Kuranes) [inline sources/osgAnimation/BasicAnimationManager.js]: Ha, ok. it's just js Float but implementing our osgjs type "object Interface"

File: sources/osgAnimation/BasicAnimationManager.js [hunk 9]
```diff
@@ -135,19 +130,12 @@ define( [
...
-            for ( i = 0; i < this._targetIDByTypes.length; i++ )
-                this._targetIDByTypes[ i ].length = 0;
+            this._resetTargets();
 
             this._activeAnimations = {};
             this._activeAnimationList.length = 0;
 
-            for ( i = 0; i < this._activeChannelsByTypes; i++ )
+            for ( var i = 0; i < this._activeChannelsByTypes; i++ )
                 this._activeChannelsByTypes[ i ].length = 0;
```
Kuranes: My point on array reset is that in osgjs code, to reset an Array, as in "emptying" it,  we do use inconsistently many things: splice, length=0, = [] 
We should use a "single way" of doing it. (that could be enforced with a code standard document) 

Either we use array.splice(0, array.length) or length = 0, or the array = [], but we should do it consistently ?

sources/osgAnimation/BasicAnimationManager.js
Kuranes: Ok.

Still Seems risky to me,  dirty saying multiple things (are callback found, are target found in call back, are animations registered), and need correct order. As that code block is only to executed when dirty is true and no "_registerAnimations" has been called prior to the "update" call. 
Not that it does happens now, but code evolution might make the '_registerAnimation' be called prior an "update" call happens afterward, and those findAnimationCallback and _registerTargetFoundInAnimationCallback be not ever called once.

sources/osgAnimation/BasicAnimationManager.js
Kuranes: k. 
( maybe add that those solution to the comment if user get the warning and want to fix it in his code)

sources/osgAnimation/BasicAnimationManager.js
cedricpinson: Dirty means something must be done, whatever is it.
Yeah I would prefer to avoid async problem, but I dont see how to do. Animations needs targets, targets needs the scene graph.

if _registerAnimations is called before update it does nothing because if there will be no targets, so it will return without doing anything and will keep the dirty flag to true.

maybe more comments would help ?

sources/osgAnimation/BasicAnimationManager.js
cedricpinson: agree, I think the best was .lenght = 0

jtorresfabra: Umm not sure, I recall to do a jsperf and I think the fastest was array = [].

cedricpinson: the advantage of length = 0 is to keep the same array, in some place it's not important but in some it is especially if the array is referenced. it depends on the case.
also found this http://jsperf.com/array-destroy/151

jtorresfabra: In this jsperf there is not a array = []. Anyway if you do think that the best is to always do array.length = 0 ok for me. The point is to do all the same. :+1:

PR #410, opened 2015-08-12
[#97603894] Adds Supersample Example

Kuranes:


Kuranes:
No Merge before Refactor/Fix of the "round robin" composer current code

sources/osg/Shape.js
```diff
@@ -330,9 +330,9 @@ define( [
         var vertexes = new MACROUTILS.Float32Array( [ -1.0, -1.0, -1.0, 4.0, 4.0, -1.0 ] );
 
         var indexes = new MACROUTILS.Uint16Array( 3 );
-        indexes[ 0 ] = 0;
+        indexes[ 0 ] = 2;
```
cedricpinson: why this change ?

Kuranes: when I added cull backface on composer... I saw the fullscreen triangle was inverted... so inverted it and added the cull backface on the composer stateset

cedricpinson: Do we want to handle backface culling in the composer ? is there a reason for that ?

Kuranes: We want to reduce the number of fragment in composer, making sure tri si drawn fullscreen once, and do it as early  as possible, for all 2D fullscreen passes

cedricpinson: I dont understand, yes it's to write fullscreen triangle. Why inversing the triangle and adding culling changes something ?

Kuranes: before the change:
- no face culling on composer (waste of gpu resource)
- FS triangle NEEDED culling disabled otherwise it showed nothing at all.
( backface culling woudln't work (not intuitive at all)) 

after the change:
- FS triangle order is now following default order ( one allowing backface culling)
- add backface culliing (no more waste of gpu resource, faster path )

cedricpinson: I dont understand why culling is needed for composer, it only uses 'fullscreen triangles' to do post process. So it does not need culling at all, I feel I miss something. am I wrong ?

sources/osgUtil/Composer.js
```diff
@@ -164,51 +176,83 @@ define( [
...
                     camera.setRenderOrder( Camera.PRE_RENDER, 0 );
-                    texture = element.texture;
+                    textureResult = element.texture;
                     var textureTarget = element.textureTarget;
-                    if ( texture === undefined ) {
-                        texture = new Texture();
-                        texture.setTextureSize( w, h );
+                    if ( textureResult === undefined ) {
+                        textureResult = new Texture();
+                        textureResult.setName( 'composer Rtt ' + element.filter._fragmentName );
```
cedricpinson: use accessors getFragmentName

sources/osg/Shape.js
Kuranes: Just to make it clear it's about face culling as in https://www.opengl.org/wiki/Face_Culling
As soon as you draw a triangle, it's needed to specify the culling the HW should apply on the triangle ?

sources/osgUtil/Composer.js
Kuranes: k

sources/osgUtil/Composer.js
```diff
@@ -164,51 +176,83 @@ define( [
...
+                var quad = Shape.createTexturedFullScreenFakeQuadGeometry();
 
                 if ( element.filter.buildGeometry !== undefined )
                     quad = element.filter.buildGeometry( quad );
 
                 quad.setName( 'composer layer' );
 
-                lastTextureResult = texture;
+                // if rendering into texture framebuffer
+                if ( textureResult ) {
```
cedricpinson: I prefer you add function call for this instead of inline code, it's better to avoid spamming debug/devel code. If it was two line that would be ok, but if more it's better to keep this outside of core function. Maybe we should add this into the coding style

sources/osgUtil/Composer.js
```diff
@@ -355,48 +426,174 @@ define( [
...
         }
     } );
 
 
+    // filter that switch its render target and its input at each frame
+    // allowing to get input for last frame render.
+    Composer.Filter.PingPong = function ( cameraRtt0, rtt0, cameraRtt1, rtt1, fragmentShader, uniforms ) {
+        Composer.Filter.Custom.apply( this, [ fragmentShader, uniforms ] );
+
+        this.cameraRtt0 = cameraRtt0;
```
cedricpinson: coding style member should be prefixed by _

examples/temporal/main.js
```diff
@@ -0,0 +1,592 @@
+( function () {
```
cedricpinson: Example should start from template example like example/quad-texture no global variable except the instance of the example and the deps. If you need access for debug you can still access member of the example instance.

examples/temporal/passthrough.glsl
```diff
@@ -0,0 +1,29 @@
...
+uniform vec4 halton;
+
+void main()
+{
+    vec4 superSampleResult = (mod(halton[3], 2.0) == 0.0) ? texture2D(Texture0, FragTexCoord0) : texture2D(Texture1, FragTexCoord0);
+
+    gl_FragColor = superSampleResult;
+
+    /* develblock:start */
+/*
```
cedricpinson: if it's useless remove this part

examples/temporal/passthrough.glsl
```diff
@@ -0,0 +1,29 @@
...
+
+uniform sampler2D Texture0;
+uniform sampler2D Texture1;
+uniform sampler2D Texture2;
+
+uniform vec4 halton;
+
+void main()
+{
+    vec4 superSampleResult = (mod(halton[3], 2.0) == 0.0) ? texture2D(Texture0, FragTexCoord0) : texture2D(Texture1, FragTexCoord0);
```
cedricpinson: what's inside halton[3] ?

examples/temporal/supersample.glsl
```diff
@@ -0,0 +1,69 @@
+varying vec2 FragTexCoord0;
+uniform vec2 RenderSize;
+uniform sampler2D Texture0;
+uniform sampler2D Texture1;
+uniform vec4 halton;
```
cedricpinson: probably it would be useful to comments what's inside halton vec4 it's not obvious when reading the glsl code

examples/temporal/main.js
Kuranes: Ok, 

Note Explaining a problem here: 
I started Code with copy/paste from the sample the most like this one. (postprocess). 
Productivity side, adding new sample from scratch is very time consuming.

Using Example would give gains, but would works only if we're sharing code between example, as in inherit from a example js and then convert all Example once and for all.

Otherwise adding a new sample is taking code parts from 2 or 3 different sample, each with its particular code, and in many case it's copy/paste and hard to test in different conditions.

Here, for example, if you want to test supersample on different condition: with a cubemap, with animation, with another RTT. 
If all those code parts where accessible when you inherit example, you'd gain massive power for debug and sample app coding speed.
Otherwise  it's very time consuming.

examples/temporal/passthrough.glsl
Kuranes: useful for debuging, but now we're done. removed.

examples/temporal/supersample.glsl
Kuranes: k, added

sources/osgUtil/Composer.js
```diff
@@ -355,48 +426,172 @@ define( [
...
+            st1.setAttributeAndModes( this._program );
+
+            if ( this._uniforms ) {
+                for ( var k = 0; k < this._uniforms.length; k++ ) {
+                    st0.addUniform( this._uniforms[ k ] );
+                    st1.addUniform( this._uniforms[ k ] );
+                }
+            }
+
+            st0.addUniform( Uniform.createInt1( 0, 'Texture0' ) );
```
cedricpinson: you can use the same uniform for st0 and st1. no need to create Texture0 and Texture1 twice

sources/osgUtil/Composer.js
```diff
@@ -355,48 +426,172 @@ define( [
...
+            st0.addUniform( Uniform.createInt1( 1, 'Texture1' ) );
+
+            st1.addUniform( Uniform.createInt1( 0, 'Texture0' ) );
+            st1.addUniform( Uniform.createInt1( 1, 'Texture1' ) );
+
+            // copy input on both camera
+            // Composer::Build set the last render into the current filter stateset texture unit 0
+            // we copy that into each camera as Texture unit 0
+            st0.setTextureAttributeAndModes( 0, filterStateSet.getTextureAttribute( 0, 'Texture' ) );
```
cedricpinson: same here filterStateSet.getTextureAttribute( 0, 'Texture' ) can be save to avoid a second call, even if it's not critique here

sources/osgUtil/Composer.js
```diff
@@ -355,48 +426,172 @@ define( [
...
+            // rtt0 to texture unit 1
+            if ( i !== array.length - 1 ) {
+
+                // just translate stateset to the next filter
+                var nextSt = array[ i + 1 ].filter.getStateSet();
+
+                nextSt.setTextureAttributeAndModes( 0, this._rtt0 );
+                nextSt.setTextureAttributeAndModes( 1, this._rtt1 );
+
+                nextSt.addUniform( Uniform.createInt1( 0, 'Texture0' ) );
```
cedricpinson: no need to re create Texture0 Texture1 here too, use the same uniforms created before. Probably we could define globally those uniform in osgjs not sure about the side effect but it would be simpler

sources/osgUtil/Composer.js
```diff
@@ -355,48 +426,172 @@ define( [
...
+            // hide one of the two pass, as we will render only one each frame
+            this._cameraRtt1.setNodeMask( ~0x0 );
+
+            // last texture result, only one possible so the first will do
+            return this._rtt0;
+        },
+        switch: function () {
+
+            // PingPong
+            if ( this._cameraRtt0.getNodeMask() === ~0x0 ) {
```
cedricpinson: could it be simplified with:
```
var nodeMask0 = this._cameraRtt0.getNodeMask();
var nodeMask1 = this._cameraRtt1.getNodeMask();
this._cameraRtt0.setNodeMask(nodeMask1);
this._cameraRtt1.setNodeMask(nodeMask0);
```

examples/temporal/temporal.js
```diff
@@ -0,0 +1,54 @@
+( function () {
```
cedricpinson: this file should be upper case it describe a class, if moving it to osgjs source tree it would be renamed. It's good to keep the same rule in examples

examples/temporal/main.js
```diff
@@ -0,0 +1,592 @@
...
+            /*
+                group.addChild( osg.createGridGeometry( size, size, size,
+                    size, size, size,
+                    size, size, size,
+                    10, 10 ) );
+             */
+
+
+            var gridDefinition = 15;
+            var TheGrid = osg.createGridGeometry( -1.0, -1.0, 0.0, 2.0, 0.0, 0.0, 0.0, 2.0, 0.0, gridDefinition, gridDefinition );
```
cedricpinson: theGrid, not TheGrid

examples/temporal/main.js
```diff
@@ -0,0 +1,592 @@
...
+            var xOffset = optionsDebug.x;
+            var yOffset = optionsDebug.y;
+            _ComposerdebugNode.removeChildren();
+
+            var stateset;
+
+            stateset = _ComposerdebugNode.getOrCreateStateSet();
+            if ( !optionsDebug.fullscreen )
+                stateset.setAttributeAndModes( new osg.Depth( 'DISABLE' ) );
+            for ( var i = 0, l = rttsArray.length; i < l; i++ ) {
```
cedricpinson: should have empty line between conditional block code

examples/temporal/main.js
```diff
@@ -0,0 +1,592 @@
...
+
+            camera.setRenderOrder( osg.Camera.PRE_RENDER, 0 );
+            camera.attachTexture( osg.FrameBufferObject.COLOR_ATTACHMENT0, texture, 0 );
+
+            //
+            camera.setReferenceFrame( osg.Transform.ABSOLUTE_RF );
+
+            if ( is3D ) {
+                camera.attachRenderBuffer( osg.FrameBufferObject.DEPTH_ATTACHMENT, osg.FrameBufferObject.DEPTH_COMPONENT16 );
+                camera.setClearColor( osg.Vec4.create( [ 0.0, 0.0, 0.1, 1.0 ] ) );
```
cedricpinson: keep ff00ff as debug color if it's intended

sources/osgUtil/Composer.js
Kuranes: Indeed ! will do.

sources/osgUtil/Composer.js
Kuranes: k

sources/osgUtil/Composer.js
Kuranes: k.

Ideally adding them internally inside State handling code. 
State.js knows what texture is what unit https://github.com/cedricpinson/osgjs/blob/develop/sources/osg/State.js#L391

( *cough cough* Could allow for the infamous optimisation "don't bind the same texture twice if it's exaclty the same" )

sources/osgUtil/Composer.js
Kuranes: Totally better code-wise, but really feel more obsuce for code reader. 
Sure you want to optimize ?

examples/temporal/main.js
Kuranes: http://www.youtube.com/watch?v=S-d7iXuJlvE&t=0m9s

examples/temporal/main.js
Kuranes: k

examples/temporal/main.js
Kuranes: no debug, it's the background color for the 3D scene here

sources/osgUtil/Composer.js
cedricpinson: nop it's just the if is useless and does not help to understand. There is no need conditional to swap variable content. And the function already says what it does

examples/templates/Example.js
```diff
@@ -0,0 +1,206 @@
...
+            for ( var i = 0; i < keys.length; i++ ) {
+                var property = keys[ i ];
+                this._config[ property ] = queryDict[ property ];
+            }
+        },
+
+        readTextures: function ( textures ) {
+
+            var textureNames = textures || this._textureNames;
+            var path = '../media/textures/';
```
cedricpinson: the path is already defined line 31 it's better if we dont duplicate it

examples/templates/Example.js
```diff
@@ -0,0 +1,206 @@
...
+    var $ = window.$;
+
+
+    var Example = function () {};
+
+    Example.prototype = {
+        // user Config
+        // overriden by user param on url
+        // used for the GUI tweak
+        _config: {
```
cedricpinson: Is there a reason with vars are not initalized in the constructor instead of the prototype ?

Kuranes: no reason whatsoever. I'll change to the constructor, better for overriding ctor

examples/templates/Example.js
Kuranes: k missed that one.

examples/templates/Example.js
```diff
@@ -0,0 +1,212 @@
...
+            this._viewer.run();
+        },
+
+        initDatGUI: function () {
+
+            this._gui = new window.dat.GUI();
+
+        },
+
+        urlOptions: function () {
```
cedricpinson: should be renamed setConfigFromOptionsURL

examples/templates/Example.js
```diff
@@ -0,0 +1,212 @@
...
+        this._gui = undefined;
+        this._shaderProcessor = undefined;
+        this._root = undefined;
+        this._canvas = undefined;
+
+
+    };
+
+    Example.prototype = {
+        // TODO: split even more for better reuse ?
```
cedricpinson: we can remove this comment

examples/templates/index-dark.html
```diff
@@ -39,7 +49,7 @@
   <div class="osgjs-info osgjs-info-dark">
 
     <a href="http://osgjs.org/" class="osgjs-powered" title="Powered by OSG.JS">
-      <img src="http://osgjs.org/assets/logo.png" alt="" width="32" height="32">
+      <img src="../../website/assets/logo.png" alt="" width="32" height="32">
```
cedricpinson: are you sure it will not fail when deploying examples on the website ?

examples/templates/index-light.html
```diff
@@ -39,7 +49,7 @@
   <div class="osgjs-info osgjs-info-dark">
 
     <a href="http://osgjs.org/" class="osgjs-powered" title="Powered by OSG.JS">
-      <img src="http://osgjs.org/assets/logo.png" alt="" width="32" height="32">
+      <img  src="../../website/assets/logo.png" alt="" width="32" height="32">
```
cedricpinson: same rq here

examples/templates/main.js
```diff
@@ -1,71 +1,53 @@
...
-        var osgViewer = OSG.osgViewer;
-        var osg = OSG.osg;
+    // inherits for the ExampleOSGJS prototype
+    var Example = function () {};
 
+    Example.prototype = osg.objectInherit( ExampleOSGJS.prototype, {
 
-        var canvas = document.getElementById( 'View' );
+        // can be overriden with url parm ?&scale=1
+        _config: {
```
cedricpinson: why not moving this in the constructor ?

examples/temporal/main.js
```diff
@@ -0,0 +1,571 @@
...
+            if ( !this._rttDebugNode ) {
+                this._rttDebugNode = new osg.Node();
+                this._rttDebugNode.setName( '_rttDebugNode' );
+                this._root.addchild( this._rttDebugNode );
+            } else if ( this._rttDebugNode.getChildren().length !== 0 ) {
+                this._rttDebugNode.removeChildren();
+                return;
+            }
+
+            var _ComposerdebugNode = new osg.Node();
```
cedricpinson: _ is used for member variable
should be named composerDebugNode

examples/temporal/main.js
```diff
@@ -0,0 +1,571 @@
...
+            _ComposerdebugNode.setName( 'debugComposerNode' );
+            _ComposerdebugNode.setCullingActive( false );
+            var _ComposerdebugCamera = new osg.Camera();
+            _ComposerdebugCamera.setName( '_ComposerdebugCamera' );
+            this.rttDebugNode.addChild( _ComposerdebugCamera );
```
cedricpinson: should be named this._rttDebugNode

examples/temporal/main.js
```diff
@@ -0,0 +1,571 @@
...
+                screenW: 1024,
+                screenH: 768,
+                fullscreen: false
+            };
+            if ( optionalArgs )
+                osg.extend( optionsDebug, optionalArgs );
+
+            var matrixDest = _ComposerdebugCamera.getProjectionMatrix();
+            osg.Matrix.makeOrtho( 0, optionsDebug.screenW, 0, optionsDebug.screenH, -5, 5, matrixDest );
+            _ComposerdebugCamera.setProjectionMatrix( matrixDest ); //not really needed until we do matrix caches
```
cedricpinson: same here

examples/templates/index-dark.html
Kuranes: arf, indeed. forgot the website. 
Thing is coding without internet on examples gives you a 5 secondes task because of this thing here. perhaps we can add logo to media example folder so that the path is constant ?

examples/temporal/main.js
```diff
@@ -0,0 +1,571 @@
...
+            this._uniforms.halton.dirty();
+
+        },
+
+
+
+
+        rotateNode: function ( nv ) {
+            var t = nv.getFrameStamp().getSimulationTime();
+            var dt = t - this._rotate._lastUpdate;
```
cedricpinson: this._rotate._lastUpdate should not be allowed. if you should alawys use an accessors if access from external the class. It's acceptable because of it's an example but should always be avoided

examples/temporal/main.js
Kuranes: oups whole RTT debug thing is a code copy/paste from 2 yrs ago.
( Refrained to copy it inside the Example, but would have liked to )

examples/temporal/main.js
```diff
@@ -0,0 +1,571 @@
...
+            var composer = new osgUtil.Composer();
+            composer.setName( 'supersample composer' );
+
+            ///
+            // Create Texture And Camera
+            ///
+
+
+            // current Destination
+            var rttAA0 = this.createTextureRTT( 'antiAlias1',
```
cedricpinson: it should be kept on one line

examples/temporal/main.js
```diff
@@ -0,0 +1,571 @@
...
+        initDatGUI: function () {
+            var controller;
+
+            this._gui = new window.dat.GUI();
+
+            // add callbacks
+            var self = this;
+            // start/stop AA accumulation
+            this._config[ 'enableSuperSample' ] = function () {
+                if ( this.supersample ) {
```
cedricpinson: https://github.com/cedricpinson/osgjs/wiki/Coding-Style#blocks
https://github.com/cedricpinson/osgjs/wiki/Coding-Style#if-statement
not sure if we can adds those rules in the eslint

examples/temporal/supersample.glsl
```diff
@@ -0,0 +1,71 @@
...
+
+        return res;
+
+    }
+    /* develblock:start */
+    // debug
+    // check dual color alternance
+    //res +=  mod(n, 2.0) == 1.0 ? vec4(1.0, 0.0, 0.0, 1.0) : vec4(0.0, 1.0, 0.0, 1.0);
+    /* develblock:end */
+  return res;
```
cedricpinson: not indented correctly

sources/osgUtil/Composer.js
```diff
@@ -247,20 +320,18 @@ define( [
...
     Composer.Filter.defaultVertexShader = [
         'attribute vec3 Vertex;',
-        'attribute vec2 TexCoord0;',
         'varying vec2 FragTexCoord0;',
-        'uniform mat4 ModelViewMatrix;',
-        'uniform mat4 ProjectionMatrix;',
         'void main(void) {',
-        '  gl_Position = ProjectionMatrix * ModelViewMatrix * vec4(Vertex,1.0);',
-        '  FragTexCoord0 = TexCoord0;',
+        '  gl_Position = vec4(Vertex*2.0 - 1.0,1.0);',
```
cedricpinson: not indented 4 spaces but there are probably a lot of other cases

examples/temporal/main.js
Kuranes: best check what you need here: 

http://eslint.org/docs/rules/brace-style
http://eslint.org/docs/rules/no-lone-blocks
http://eslint.org/docs/rules/block-spacing
http://eslint.org/docs/rules/array-bracket-spacing
http://eslint.org/docs/rules/padded-blocks
http://eslint.org/docs/rules/no-lonely-if

sources/osgUtil/Composer.js
Kuranes: need a PR to export all glsl code in glsl files + a PR for splitting composer into composer.js filter.js + ilters/custom.js filters/blur.js etc

examples/templates/index-light.html
Kuranes: same rq as above.
(shameless up)

sources/osgUtil/Composer.js
```diff
@@ -355,48 +426,168 @@ define( [
...
+
+            var st0 = this._cameraRtt0.getOrCreateStateSet();
+            var st1 = this._cameraRtt1.getOrCreateStateSet();
+
+            // copy filter program and uniforms on the 2 cameras
+            st0.setAttributeAndModes( this._program );
+            st1.setAttributeAndModes( this._program );
+
+            if ( this._uniforms ) {
+                for ( var k = 0; k < this._uniforms.length; k++ ) {
```
stephomi: cache length?

sources/osgUtil/Composer.js
```diff
@@ -355,48 +426,168 @@ define( [
...
+
+                nextSt.setTextureAttributeAndModes( 0, this._rtt0 );
+                nextSt.setTextureAttributeAndModes( 1, this._rtt1 );
+
+                nextSt.addUniform( uniformTU0 );
+                nextSt.addUniform( uniformTU1 );
+            }
+
+            var quad = Shape.createTexturedFullScreenFakeQuadGeometry();
+            if ( this.buildGeometry !== undefined )
```
stephomi: !== undefined is not necessary here (it's only necessary with number of string)

sources/osgUtil/Composer.js
```diff
@@ -355,48 +426,168 @@ define( [
...
+            this._vertexShader = v;
+        },
+        autoBindFragmentUniformStateSet: function ( stateSet, fragmentShader, uniforms ) {
+
+            var unitIndex = 0;
+            // TODO: check if not a better place Utils
+            // and reuse (we already do this in shader)
+            // At least DEFINE the regexp somewhere somehow
+            var r = fragmentShader.match( /uniform\s+\w+\s+\w+/g );
+            if ( r !== null ) {
```
stephomi: early rejection?

sources/osgUtil/Composer.js
```diff
@@ -355,48 +426,168 @@ define( [
...
+                for ( var i = 0, l = r.length; i < l; i++ ) {
+                    var match = r[ i ].match( /uniform\s+(\w+)\s+(\w+)/ );
+                    var uniformType = match[ 1 ];
+                    var uniformName = match[ 2 ];
+                    var uniform;
+
+                    if ( uniforms[ uniformName ] !== undefined ) {
```
stephomi: early rejection too ? (continue;)

sources/osgUtil/Composer.js
```diff
@@ -113,22 +123,51 @@ define( [
...
+
+            for ( var k = 0, lt = stateSet.getNumTextureAttributeLists(); k < lt; k++ ) {
+                var textureBinded = stateSet.getTextureAttribute( k, 'Texture' );
+                Notify.assert( textureResult !== textureBinded, 'Composer: write/read at the same time on a texture is undefined behavior' );
+
+            }
+        },
+        /* develblock:end */
+
         build: function () {
```
stephomi: Is it possible to add an option to eslint to disallow functions longer than 100 lines? :)

Kuranes: possible to code one.
default one are on parameter number or block nested depth limit

PR #411, opened 2015-08-12
[#97604240] Fix shadow texture uniform warning

Kuranes:
Fix useless shadow texture binding during cast
Removes useless uniforms for none/pcf
Add shadow offset rotation

### DIFF
--- a/sources/osgShader/Compiler.js
+++ b/sources/osgShader/Compiler.js
@@ -169,7 +169,7 @@ define( [
         registerTextureShadow: function ( tuTarget, texUnit ) {
             var tName = tuTarget.getName();
             if ( !tName ) {
-                tName = 'ShadowTexture' + texUnit;
+                tName = 'Texture' + texUnit;
                 tuTarget.setName( tName );
             }
             this._shadowsTextures[ texUnit ] = tuTarget;
@@ -314,14 +314,23 @@ define( [
                 Notify.error( 'Cannot create unamed Uniform' );
             }
 
+
             var exist = this._variables[ nameID ];
             if ( exist ) {
                 // see comment in Variable function
-                if ( exist.getType() !== type ) {
+                if ( exist.getType() === type ) {
+                    return exist;
+                }
+                /*develblock:start*/
+                // texture has a particular "dual" type of uniform
+                // a sampler2D
+                // a int pointing to the texture unit the sampler2D represents
+                if ( exist.getType() === 'sampler2D' && type !== 'sampler2D' ) {
                     Notify.error( 'Same uniform, but different type' );
                 }
-                return exist;
+                /*develblock:end*/
             }
+
 
             var v = this.getNode( 'Uniform', type, nameID, size );
             this._variables[ nameID ] = v;
@@ -337,9 +346,13 @@ define( [
 
             var exist = this._variables[ nameID ];
             if ( exist ) {
+
+                /*develblock:start*/
                 if ( exist.getType() !== type ) {
                     Notify.error( 'Same attribute, but different type' );
                 }
+                /*develblock:end*/
+
                 return exist;
             }
 
@@ -366,9 +379,13 @@ define( [
 
                 var exist = this._variables[ nameID ];
                 if ( exist ) {
+
+                    /*develblock:start*/
                     if ( exist.getType() !== type ) {
                         Notify.error( 'Same constant name, but different type' );
                     }
+                    /*develblock:end*/
+
                     // see comment in Variable function
                     return exist;
                 }
@@ -389,13 +406,17 @@ define( [
 
             var exist = this._variables[ nameID ];
             if ( exist ) {
+
+                /*develblock:start*/
                 // something went wrong: you created a variable and try to access it like a varying
-                if ( !this._varyings[ nameID ] ) {
+                if ( !this._varyings[ nameID ] )
                     Notify.error( 'Error: requesting a varying not declared with getOrCreateVarying previously' );
-                }
+
                 if ( exist.getType() !== type ) {
                     Notify.error( 'Error: Same varying, but different type' );
                 }
+                /*develblock:end*/
+
                 // see comment in Variable function
                 return exist;
             } else {
@@ -416,12 +437,13 @@ define( [
                 }
             }
 
-
+            /*develblock:start*/
             // if it's not in Varying Cache, but requested from fragment shader
             // it means => error
             if ( this._fragmentShaderMode && !this._customVertexShader ) {
                 Notify.error( 'Error: requesting a varying not declared in Vertex Shader Graph.( if a Custom Vertex Shader in a custom processor, add this._customVertexShader to your custom processor): ' + nameID + ' ' + type );
             }
+            /*develblock:end*/
 
             var v = this.getNode( 'Varying', type, nameID );
             this._variables[ nameID ] = v;
@@ -627,21 +649,21 @@ define( [
 
             var samplerName = 'Texture' + unit.toString();
             var textureSampler = this.getVariable( samplerName );
-
-            if ( textureSampler === undefined ) {
-
+            if ( !textureSampler ) {
                 if ( texture.className() === 'Texture' ) {
                     textureSampler = this.getOrCreateSampler( 'sampler2D', samplerName );
                 } else if ( texture.className() === 'TextureCubeMap' ) {
                     textureSampler = this.getOrCreateSampler( 'samplerCube', samplerName );
                 } else if ( texture.className() === 'ShadowTexture' ) {
-                    textureSampler = this.getOrCreateSampler( 'sampler2D', samplerName );
-                    // return now to prevent creation of useless FragTexCoord
-                    //( shadow creates its own texcoord)
+                    // don't pre-declare texture for shadow
+                    // as it's uniform node caching/get/compilation must not
+                    // get the dual int/sampler2D thing.
+
+                    //textureSampler = this.getOrCreateSampler( 'sampler2D', samplerName );
+                    // return;
+
                     return;
                 }
-
-
             }
 
             // texture coordinates are automatically mapped to unit texture number
@@ -739,28 +761,14 @@ define( [
 
             // shadow Attribute uniforms
             var shadowUniforms = this.getOrCreateStateAttributeUniforms( this._shadows[ lightIndex ], 'shadow' );
-            inputs = MACROUTILS.objectMix( inputs, shadowUniforms );
+            var shadowInputs = MACROUTILS.objectMix( inputs, shadowUniforms );
 
             // shadowTexture  Attribute uniforms AND varying
-            var tex;
             // TODO: better handle multi texture shadow (CSM/PSM/etc.)
             for ( k = 0; k < shadowTextures.length; k++ ) {
-
                 shadowTexture = shadowTextures[ k ];
                 if ( shadowTexture ) {
-                    tex = this.getOrCreateSampler( 'sampler2D', shadowTexture.getName() );
-                    inputs.shadowTexture = tex;
-
-                    // per texture uniforms
-                    var shadowTextureUniforms = this.getOrCreateTextureStateAttributeUniforms( shadowTexture, 'shadowTexture', k );
-                    inputs = MACROUTILS.objectMix( inputs, shadowTextureUniforms );
-
-                    var shadowVarying = {
-                        vertexWorld: vertexWorld,
-                        lightEyeDir: inputs.lightEyeDir,
-                        lightNDL: inputs.lightNDL
-                    };
-                    inputs = MACROUTILS.objectMix( inputs, shadowVarying );
+                    shadowInputs = this.createShadowTextureInputVarying( shadowTexture, shadowInputs, vertexWorld, k );
                 }
 
             }
@@ -775,6 +783,36 @@ define( [
 
         },
 
+        createShadowTextureInputVarying: function ( shadowTexture, inputs, vertexWorld, tUnit ) {
+            var shadowTexSamplerName = 'Texture' + tUnit;
+
+            // per texture uniforms
+            var shadowTextureUniforms = this.getOrCreateTextureStateAttributeUniforms( shadowTexture, 'shadowTexture', tUnit );
+
+            // UGLY REMOVAL: special case of the sampler texture as method above create
+            // unforms for all shadow texture uniform,
+            // including texture unit Int 'Texture0' which becomes
+            // 'shadowTextureTexture0' which we won't use
+            // so we remove it from adding it here
+            var id = shadowTextureUniforms[ 'shadowTexture' + shadowTexSamplerName ].getID();
+            shadowTextureUniforms[ 'shadowTexture' + shadowTexSamplerName ] = undefined;
+            this._variables[ shadowTexSamplerName ] = undefined;
+            delete this._activeNodeList[ id ];
+            // end UGLY REMOVAL
+
+            var inputsShadow = MACROUTILS.objectMix( inputs, shadowTextureUniforms );
+
+            var tex = this.getOrCreateSampler( 'sampler2D', shadowTexSamplerName );
+            inputsShadow.shadowTexture = tex;
+
+            var shadowVarying = {
+                vertexWorld: vertexWorld,
+                lightEyeDir: inputsShadow.lightEyeDir,
+                lightNDL: inputsShadow.lightNDL
+            };
+            inputsShadow = MACROUTILS.objectMix( inputsShadow, shadowVarying );
+            return inputsShadow;
+        },
         // Shared var between lights and shadows
         createCommonLightingVars: function ( materials, enumLights, numLights ) {
 
@@ -1401,6 +1439,7 @@ define( [
             // Process defines, add precision, resolve include pragma
             var shader = this._shaderProcessor.processShader( shaderStr, defines, extensions, type );
 
+            /*develblock:start*/
             // Check
             var compiledNodes = Object.keys( this._compiledNodeList );
             var activeNodes = Object.keys( this._activeNodeList );
@@ -1414,7 +1453,7 @@ define( [
                 }
                 return found;
             }, this );
-
+            /*develblock:end*/
 
             // return the complete shader string.
             // now is compilable by gl driver
--- a/sources/osgShader/ShaderProcessor.js
+++ b/sources/osgShader/ShaderProcessor.js
@@ -100,7 +100,7 @@ define( [
         // recursively  handle #include external glsl
         // files (for now in the same folder.)
         preprocess: function ( content, sourceID, includeList, inputsDefines ) {
-            var _self = this;
+            var self = this;
             return content.replace( this._includeCondR, function ( _, name ) {
                 var includeOpt = name.split( ' ' );
                 var includeName = includeOpt[ 0 ].replace( /"/g, '' );
@@ -116,7 +116,8 @@ define( [
                     // if not defined we do not include
                     var found = false;
                     var defines = inputsDefines.map( function ( defineString ) {
-                        return _self._defineR.test( defineString ) && defineString.split( ' ' )[ 1 ];
+                        // find '#define', remove duplicate whitespace, split on space and return the define Text
+                        return self._defineR.test( defineString ) && defineString.replace( /\s+/g, ' ' ).split( ' ' )[ 1 ];
                     } );
 
                     for ( var i = 1; i < includeOpt.length && !found; i++ ) {
--- a/sources/osgShader/node/noise.glsl
+++ b/sources/osgShader/node/noise.glsl
@@ -1,0 +1,21 @@
+#pragma include "hash.glsl"
+
+//// noise glsl
+
+// Based on Morgan McGuire @morgan3d
+// https://www.shadertoy.com/view/4dS3Wd
+float noise2D(const in vec2 x) {
+    vec2 i = floor(x);
+    vec2 f = fract(x);
+
+    // Four corners in 2D of a tile
+    float a = hash21(i);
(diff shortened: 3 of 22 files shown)


Kuranes:
(Not Ready For Merge)

sources/osg/State.js
```diff
@@ -515,8 +515,10 @@ define( [
             // if the the stack has changed but the last applied attribute is the same
             // then we dont need to apply it again
             if ( attributeStack.lastApplied !== attribute ) {
-                gl.activeTexture( gl.TEXTURE0 + textureUnit );
-                attribute.apply( this, textureUnit );
+                if ( !attribute._textureNull ) {
```
cedricpinson: Adds a methods to Texture attribute like isTextureNull then use this method for the test

cedricpinson: Also you can split this commit to another PR because it's easy to merge it but maybe we will spend more time on shadow review

sources/osgShader/Compiler.js
```diff
@@ -317,10 +317,12 @@ define( [
...
             if ( exist ) {
                 // see comment in Variable function
-                if ( exist.getType() !== type ) {
+                if ( exist.getType() === type ) {
+                    return exist;
+                }
+                if ( exist.getType() === 'sampler2D' && type !== 'sampler2D' ) {
                     Notify.error( 'Same uniform, but different type' );
                 }
-                return exist;
```
cedricpinson: it means that if it's not the same type we dont return exist anymore, what is the consequence of changing that ?

sources/osgShader/Compiler.js
```diff
@@ -627,21 +629,14 @@ define( [
...
-                    // return now to prevent creation of useless FragTexCoord
-                    //( shadow creates its own texcoord)
-                    return;
-                }
-
-
+                textureSampler = this.getOrCreateSampler( 'sampler2D', samplerName );
+            } else if ( texture.className() === 'TextureCubeMap' ) {
+                textureSampler = this.getOrCreateSampler( 'samplerCube', samplerName );
+            } else if ( texture.className() === 'ShadowTexture' ) {
```
cedricpinson: I guess it's done somewhere else but document it

sources/osgShader/Compiler.js
```diff
@@ -748,13 +743,22 @@ define( [
 
                 shadowTexture = shadowTextures[ k ];
                 if ( shadowTexture ) {
```
cedricpinson: Can we create a function ?

sources/osgShader/Compiler.js
```diff
@@ -748,13 +743,22 @@ define( [
...
-                    inputs.shadowTexture = tex;
+                    var shadowTexSamplerName = 'ShadowTexture' + k;
+
 
                     // per texture uniforms
                     var shadowTextureUniforms = this.getOrCreateTextureStateAttributeUniforms( shadowTexture, 'shadowTexture', k );
+
+                    // TODO: UGLY REMOVAL
+                    // create a specific texture getuniform for compiler
+                    shadowTextureUniforms[ 'shadowTexture' + shadowTexSamplerName ] = undefined;
```
cedricpinson: how could we improve this ?

sources/osgShadow/ShadowMap.js
```diff
@@ -420,45 +420,45 @@ define( [
 
             this._shadowReceiveAttribute.setLightNumber( lightNumber );
 
-            this._receivingStateset.setAttributeAndModes( this._shadowReceiveAttribute, StateAttribute.ON | StateAttribute.OVERRIDE );
 
 
+            var fullOverride = StateAttribute.OVERRIDE | StateAttribute.ON;
```
cedricpinson: can you explain ?

sources/osgShadow/ShadowMap.js
```diff
@@ -420,45 +420,45 @@ define( [
...
+            var shouldGetMaxTextureUnits = 32;
             //
-            // TODO:: Should handle the alpha_mask that uses a texture
-            // case somehow...
+            // TODO:: Should handle the alpha_mask that uses a texture case somehow...
             // deduce from shader compil ?
-            //for ( var k = 0; k < shouldGetMaxTextureUnits; k++ ) {
-            //    this._casterStateSet.setTextureAttributeAndModes( k, blankTexture, StateAttribute.OFF | StateAttribute.OVERRIDE | StateAttribute.PROTECTED );
-            //}
+            for ( var k = 0; k < shouldGetMaxTextureUnits; k++ ) {
```
cedricpinson: maybe split this part into a function that says what it does, just to understand better. This part is to 'optimize' avoid texture binding right ? so maybe this could be done in a function with an appropriate name

sources/osgShadow/ShadowTexture.js
```diff
@@ -77,8 +77,9 @@ define( [
 
             }.bind( this ) );
 
-
+            // tells shader where to find it
```
cedricpinson: I dont understand the extra logic, could you explains why it's not like Texture attribute ?

sources/osgShadow/shaders/tapPCF.glsl
```diff
@@ -2,12 +2,44 @@
 
 #pragma include "shadowLinearSoft.glsl"
 
+float hash1D(float n) { return fract(sin(n) * 1e4); }
+float hash2D(vec2 p) { return fract(1e4 * sin(17.0 * p.x + p.y * 0.1) * (0.1 + abs(sin(p.y * 13.0 + p.x)))); }
+
+
+float noise2D(vec2 x) {
```
cedricpinson: is noise2D general ? if yes it should go in a different glsl file, else it should be prefixed by something like shadow something

sources/osg/State.js
Kuranes: ok

sources/osgShader/Compiler.js
Kuranes: ShadowTexture is the first Texture inherited class with uniform attached.
So when we get those for the compiler, we get uniform we don't want (the int 
texture location for shader texture pointer to bind)

That Int uniform conflicts with the sample2D uniform in our compiler.

Either we make compiler handle that case (not easy) or we make sure we don't do those conflicts (meaning a special getOrCreateUniform, with a parameter like "getShaderTextureLocation" flag at false when called from compiler, true when called from other code)

sources/osgShader/Compiler.js
Kuranes: k

sources/osgShadow/ShadowMap.js
Kuranes: Idea is to make sure the null texture is "binded" to all shadow casting scene, so we override all existing texture bind in the scene, preventing any texture bind.

When user implements alpha mask casting, they use PROTECTED to prevent OVERRIDE to remove their alpha mask texture needed.

sources/osgShadow/ShadowMap.js
Kuranes: k

sources/osgShadow/ShadowMap.js
Kuranes: I'll add a comment

sources/osgShadow/ShadowTexture.js
Kuranes: it's the same apart from the texture name change to reflect the class name ?

sources/osgShadow/shaders/tapPCF.glsl
Kuranes: ok, will move it

sources/osg/State.js
Kuranes: https://github.com/cedricpinson/osgjs/pull/412

sources/osgShader/Compiler.js
Kuranes: error is breaking and the user will get errors on compilation if the node is really used.
(in our case the node isn't used in compilation...)

sources/osgShadow/ShadowMap.js
Kuranes: splitter in the pr https://github.com/cedricpinson/osgjs/pull/412

sources/osgShadow/ShadowTexture.js
Kuranes: Note: that's actually the piece of code that cause problem. 
(as the compiler will make a uniform node out if it which will conflict with the sample2D uniform)

sources/osgShader/Compiler.js
Kuranes: done

sources/osgShadow/shaders/tapPCF.glsl
Kuranes: done

sources/osgShader/Compiler.js
cedricpinson: could you add a few comment about it

sources/osgShader/ShaderProcessor.js
```diff
@@ -116,7 +116,7 @@ define( [
                     // if not defined we do not include
                     var found = false;
                     var defines = inputsDefines.map( function ( defineString ) {
-                        return _self._defineR.test( defineString ) && defineString.split( ' ' )[ 1 ];
+                        return _self._defineR.test( defineString ) && defineString.replace( /\s+/g, ' ' ).split( ' ' )[ 1 ];
```
cedricpinson: why it changed ? adds a comment plz

sources/osgShader/Compiler.js
```diff
@@ -742,25 +737,11 @@ define( [
...
-                    var shadowTextureUniforms = this.getOrCreateTextureStateAttributeUniforms( shadowTexture, 'shadowTexture', k );
-                    inputs = MACROUTILS.objectMix( inputs, shadowTextureUniforms );
-
-                    var shadowVarying = {
-                        vertexWorld: vertexWorld,
-                        lightEyeDir: inputs.lightEyeDir,
-                        lightNDL: inputs.lightNDL
-                    };
-                    inputs = MACROUTILS.objectMix( inputs, shadowVarying );
+                    this.shadowTextureCreateUniformsInputAndVarying( shadowTexture, inputs, vertexWorld, k );
```
cedricpinson: use the same scheme of naming method something like
createShadowTextureAndVarying
createShadowTextureInput
maybe better

sources/osgShader/node/noise.glsl
```diff
@@ -0,0 +1,25 @@
+
+float hash1D(float n) { return fract(sin(n) * 1e4); }
```
cedricpinson: is there some links to understand this ?

sources/osgShader/Compiler.js
Kuranes: it's part of the thing that need refactor/thinking:
Sampler2D and Int shader Texture location uniform are actually both sharing the same name thus breaking compiler uniform duplicate detection of same uniform with different types

sources/osgShadow/ShadowTexture.js
cedricpinson: Ok we should probably talk about this. I will spend more time to understand this part.

sources/osgShadow/shaders/tapPCF.glsl
```diff
@@ -1,5 +1,7 @@
 
 
+#pragma include "noise.glsl" "_ROTATEOFFSET"
```
cedricpinson: Could you improve the coding style with the glsl part and the meaning of _ in defines plz

Kuranes: done

sources/osgShader/ShaderProcessor.js
Kuranes: done

sources/osgShader/node/noise.glsl
Kuranes: done updated with many links and source

sources/osgShader/Compiler.js
Kuranes: done

cedricpinson:
needs a rebase

sources/osgShader/Compiler.js
stephomi: Instead of ``=== undefined``, you can just write ``!textureSampler`` (it's not like we are testing a string or a number)
Also, looks like it's broken. We can't declare a samplerCube anymore.... since it will fallback in the first branch.
And, why do we always test for ``getVariable`` when we only do a ``getOrCreate`` just after?

In sketchfab we don't use ``declareTexture`` and ``declareTextures`` because they declare too much nodes even when it's not needed.
Ping me back so that we can discuss about re-use of sketchab compiler code about texture management.

Kuranes: - *!textureSampler* :+1:  
- *broken* :seems a if disapeared here, will fix 
- *why do we always test for getVariable when we only do a getOrCreate just after?* Cannot Agree More ! Especially if we already does that in other production code... Those "Declare" are not flexible at all.

sources/osgShader/Compiler.js
```diff
@@ -775,6 +783,34 @@ define( [
...
+            // UGLY REMOVAL
+            // TODO: create a specific texture getuniform for compiler
+            var id = shadowTextureUniforms[ 'shadowTexture' + shadowTexSamplerName ].getID();
+            shadowTextureUniforms[ 'shadowTexture' + shadowTexSamplerName ] = undefined;
+            this._variables[ shadowTexSamplerName ] = undefined;
+            this._activeNodeList[ id ] = undefined;
+            delete this._activeNodeList[ id ];
+            // end UGLY REMOVAL
+
+            inputs = MACROUTILS.objectMix( inputs, shadowTextureUniforms );
```
cedricpinson: dont override arguments variable, use a different name. not sure it works when strict is used. To avoid any doubt use a different variable name

sources/osgShader/ShaderProcessor.js
```diff
@@ -116,7 +116,8 @@ define( [
                     // if not defined we do not include
                     var found = false;
                     var defines = inputsDefines.map( function ( defineString ) {
-                        return _self._defineR.test( defineString ) && defineString.split( ' ' )[ 1 ];
+                        // find '#define', remove duplicate whitespace, split on space and return the define Text
+                        return _self._defineR.test( defineString ) && defineString.replace( /\s+/g, ' ' ).split( ' ' )[ 1 ];
```
cedricpinson: _self is a bad name variable but I know it was here before

sources/osgShader/node/shadows.js
```diff
@@ -13,7 +13,7 @@ define( [
 
     ShadowReceive.prototype = MACROUTILS.objectInherit( Node.prototype, {
         type: 'ShadowReceiveNode',
-        validInputs: [ 'lighted', 'shadowTexture', 'shadowTextureMapSize', 'shadowTextureProjectionMatrix', 'shadowTextureViewMatrix', 'shadowTextureDepthRange', 'lightNDL', 'vertexWorld', 'shadowbias' /* 'shadowexponent0', 'shadowexponent1', 'shadowepsilonVSM' */ ],
+        validInputs: [ 'lighted', 'shadowTexture', 'shadowTextureMapSize', 'shadowTextureProjectionMatrix', 'shadowTextureViewMatrix', 'shadowTextureDepthRange', 'lightNDL', 'vertexWorld', 'shadowbias', /* 'shadowexponent0', 'shadowexponent1', 'shadowepsilonVSM' */ ],
```
cedricpinson: if not used shadowexponent0 the put the comment outside the array.

Kuranes: it's used but not mandatory (EVSM/VSM/ESM only), that's the idea of the comment.

sources/osgShader/Compiler.js
Kuranes: k

cedricpinson:
this is not fixed https://github.com/cedricpinson/osgjs/pull/411/files#r39389990

sources/osgShadow/shaders/hash.glsl
```diff
@@ -1,24 +1,52 @@
...
-float hashNormalized11( const in float n )
+// Dave Hoskins: hash without sin
+//https://www.shadertoy.com/view/4djSRW
+#define MOD2 vec2(443.8975,397.2973)
+//----------------------------------------------------------------------------------------
+//  1 out, 1 in...
+//note: normalized uniform random, [0;1[
+float hash11(const in float p)
+{
+  vec2 p2 = fract(vec2(p) * MOD2);
```
cedricpinson: this file use 2 spaces for indentation but this one https://github.com/cedricpinson/osgjs/pull/411/files#diff-f88a75996357a297e59617395fc78b26R8 use 4 spaces. Looks like we have the same problem on osgjs, could you use 4 spaces for all glsl files plz ?

Kuranes: arf, editorconfig was 2 spaces... updated with 4 spaces.

sources/osgShadow/ShadowMap.js
```diff
@@ -236,7 +236,7 @@ define( [
 
         // Overridable Visitor so that user can override the visitor to enable disable
         // in its own shadowmap implementation
-        this._removeNodesNeverCastingVisitor = new RemoveNodesNeverCastingVisitor( this._castsShadowTraversalMask );
+        this._removeNodesNeverCastingVisitor = settings.userCallbackNodeCaster || new RemoveNodesNeverCastingVisitor( this._castsShadowTraversalMask );
```
cedricpinson: It's better to use setter, it's more the way used in osg/js to change content. do you use settings.userCallbackNodeCaster somewhere in the code ?

Kuranes: no other place than that.

We talked about ShadowSettings and agreed it was a to ok be used without setters as it's a sort of dictionnary

sources/osgShadow/ShadowTexture.js
```diff
@@ -77,17 +77,25 @@ define( [
...
+            // tells Shader Program where to find it
             var name = 'ShadowTexture' + unit;
+            this._nameSampler = name;
             var uniform = Uniform.createInt1( unit, name );
             uniforms[ 'ShadowTexture' + unit ] = uniform;
 
+            // Per Class Uniform Cache
             obj.uniforms[ unit ] = new Map( uniforms );
 
             this.latestUnit = unit;
```
cedricpinson: this._latestUnit

sources/osgShadow/ShadowMap.js
cedricpinson: I am not sure about passing function or class instance in settings, it's not used for all other osgjs code.

Kuranes: I started with only the apply of the default visitor calling a callback and in the end wasn't flexible enough, so I put the entire thing as a visitor.

But it could be a callback and give the shadow casted scene root node as parameter, and then user decide what he does with it.

sources/osgShadow/ShadowTexture.js
Kuranes: fixed

sources/osgShadow/ShadowMap.js
Kuranes: removed for now, will fix in the appropried Branch/PR: https://github.com/cedricpinson/osgjs/tree/feature/shadowcastervisitor

sources/osgShader/Compiler.js
```diff
@@ -775,6 +783,34 @@ define( [
...
+
+            // per texture uniforms
+            var shadowTextureUniforms = this.getOrCreateTextureStateAttributeUniforms( shadowTexture, 'shadowTexture', tUnit );
+
+            // UGLY REMOVAL
+            // TODO: create a specific texture getuniform for compiler
+            var id = shadowTextureUniforms[ 'shadowTexture' + shadowTexSamplerName ].getID();
+            shadowTextureUniforms[ 'shadowTexture' + shadowTexSamplerName ] = undefined;
+            this._variables[ shadowTexSamplerName ] = undefined;
+            this._activeNodeList[ id ] = undefined;
```
cedricpinson: this line seems useless because just after we delete it entry

sources/osgShader/Compiler.js
```diff
@@ -775,6 +783,34 @@ define( [
...
 
+        createShadowTextureInputVarying: function ( shadowTexture, inputs, vertexWorld, tUnit ) {
+            var shadowTexSamplerName = 'Texture' + tUnit;
+
+            // per texture uniforms
+            var shadowTextureUniforms = this.getOrCreateTextureStateAttributeUniforms( shadowTexture, 'shadowTexture', tUnit );
+
+            // UGLY REMOVAL
+            // TODO: create a specific texture getuniform for compiler
+            var id = shadowTextureUniforms[ 'shadowTexture' + shadowTexSamplerName ].getID();
```
cedricpinson: is that part need an update because of sampler named changed ?

Kuranes: no need, didn't change other uniform name, juste texture name ( it's the shadowTexSamplerName 2 line above I had to change)

sources/osgShadow/ShadowTexture.js
```diff
@@ -77,17 +77,24 @@ define( [
...
+            uniforms[ name ] = uniform;
 
+            // Per Class Uniform Cache
             obj.uniforms[ unit ] = new Map( uniforms );
 
-            this.latestUnit = unit;
             return obj.uniforms[ unit ];
         },
 
+        getSamplerName: function () {
```
cedricpinson: I guess we can remove it ?

Kuranes: yes

sources/osgShader/Compiler.js
Kuranes: added a explanation comment

sources/osgShader/Compiler.js
Kuranes: removed

PR #416, opened 2015-08-24
Improves the way we manage animation attribute and bones uniforms

stephomi:


sources/osgAnimation/AnimationAttribute.js
```diff
@@ -11,33 +12,27 @@ define( [
...
         this._enable = !disable;
-
-        this._animationID = animationID;
-        if ( !animationID ) this._animationID = this.getInstanceID();
-
     };
 
-
     AnimationAttribute.uniforms = {};
+    AnimationAttribute.maxBoneSize = 0;
```
cedricpinson: probably the limit this should not be global in the AnimationAttribute, doing it this way lock the usage of only one bones settings for the entire scene. For example you can't have 2 differents AnimationAttribute settings in the scene. An application could specialize AnimationAttribute for differents settings.
Ideally the viewer could use a visitor when his scene is loaded to configure bones limits as he wants. 
Having a second parameter in the constructor could give this flexibility and if not provided, it would use the global value.

stephomi: Updated.
(second parameter).

PR #417, opened 2015-08-25
[#101020636] Fixes null weight in animated model

gfaucill:


sources/osgShader/node/skeletal.glsl
```diff
@@ -14,6 +14,8 @@ mat4 skeletalTransform( const in vec4 weightsVec, const in vec4 bonesIdx ) {
     mat4 tmpMat = mat4(1.0);
     mat4 outMat = mat4(0.0);
 
+    if((weightsVec.x + weightsVec.y + weightsVec.z + weightsVec.w) == 0.0) return tmpMat;
```
cedricpinson: @Kuranes do you have a trick to do this in an optimized way ?

Kuranes: glsl vector comparison equal( vec, veczero) ?

cedricpinson: @gfaucill tu peux faire l'update

gfaucill:
@cedricpinson Done

sources/osgShader/node/skeletal.glsl
```diff
@@ -14,6 +14,8 @@ mat4 skeletalTransform( const in vec4 weightsVec, const in vec4 bonesIdx ) {
     mat4 tmpMat = mat4(1.0);
     mat4 outMat = mat4(0.0);
 
+    if(all(equal(weightsVec, vec4(0.0)))) return tmpMat;
```
cedricpinson: why all ? if it's not equal it should enough no ?

gfaucill:
equal returns a vec4 of boolean != boolean

PR #423, opened 2015-09-09
[#101123146] Fixes skeleton update order

gfaucill:


sources/osgAnimation/UpdateSkeleton.js
```diff
@@ -54,6 +54,10 @@ define( [
         needToValidate: function () {
             return this._needValidate;
         },
+        compare: function ( a /*, b */ ) {
+            return a instanceof Bone ? -1 : 1;
```
cedricpinson: use 0 instead of -1 plz

sources/osgAnimation/UpdateSkeleton.js
```diff
@@ -54,6 +54,13 @@ define( [
         needToValidate: function () {
             return this._needValidate;
         },
+        compare: function ( x, y ) {
```
cedricpinson: could you move out this function from the class, it does not require to be a member.
just define it outside like
var compareBone = function( a, b ) {...};
then use it directly

Issue #399, opened 2015-08-04
Unnecessary GL states call at "end of Renderer:draw" frame(up and above 18 gl state calls)


Kuranes:
repro steps:
run a sample, capture a glTrace with gl inspector, see all call after the last gl.DrawElement to end of frame.

Comment  State.apply here:
https://github.com/cedricpinson/osgjs/blob/a4587df5d542736eb45da16244db8f3d0c543c20/sources/osgViewer/Renderer.js#L193

re-run same sample, capture a  gltrace with glinspector.

And see that all the unnecessary calls are removed.

All samples/tutorial runs well without but "multiple viewer" sample

But it shouldn't be enabled in all the other use case, as GL call are very costly  (and pollute gl traces)

In case of FOG sample, half the gl call are unnecessary (18/36).

if use case is to "revert webgl state to a neutral state", better just share the "lastapplied" states thing between frame/viewers as it would allows for even fewer states on usecase where states are always the same during and between frames/viewers.)

stephomi:
When we create a viewer, when the run() function is called, it creates a shader without material, triggered by

https://github.com/cedricpinson/osgjs/blob/develop/sources/osgViewer/Renderer.js#L193

The shader is not send to the cg and is not used, so it's not really a problem but maybe we can prevent the useless code execution

cedricpinson:
Since a lot of stuff has been fixed to avoid duplicate state, it should be fixed

PR #430, opened 2015-09-16
[#104164510] Adds screenshot tutorial

Kuranes:
uses glReadPixel, freeing the need for preserveDrawingBuffer performance impacts.

### DIFF
--- a/sources/osgViewer/Viewer.js
+++ b/sources/osgViewer/Viewer.js
@@ -449,7 +449,9 @@ define( [
         beginFrame: function () {
             this._startFrameTick = Timer.instance().tick();
         },
-
+        setEndFrameCallBack: function ( u ) {
+            this._userEndFrameCallback = u;
+        },
         endFrame: function () {
 
             var frameNumber = this.getFrameStamp().getFrameNumber();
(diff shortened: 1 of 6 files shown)


Kuranes:
Hitting a wall on #399
because the glTrace show that we apply a default "gl.viewport(800,600)" breaking glReadPixel read.

Kuranes:
it' just about adding a camera lastDraw callback here:
https://github.com/cedricpinson/osgjs/blob/develop/sources/osg/RenderStage.js#L129
as done in OSG, see here
https://github.com/openscenegraph/osg/blob/f1a0ff43c6bd47387a480de89f35fff2d0ed7a95/src/osgUtil/RenderStage.cpp#L1316

cedricpinson:
@Kuranes yes you can implement it using the same method as in osg

sources/osgViewer/Viewer.js
```diff
@@ -449,7 +449,9 @@ define( [
         beginFrame: function () {
             this._startFrameTick = Timer.instance().tick();
         },
-
+        setEndFrameCallBack: function ( u ) {
```
cedricpinson: why this accerssors ? is viewer.getCamera().setFinalDrawCallback( cb ) not enough ?

Kuranes: mm first iteration added that before the camera callback. 
Not needed here, still would find that useful in many case/experiments with code.
can remove

PR #432, opened 2015-09-17
Reduces the number of useless default gl calls

stephomi:


jtorresfabra:
Hi @stephomi,
It doesn't work for me for multipleviewers example. Just talked about that yesterday with @Kuranes.

Kuranes:
- Doesn't work on linux box: mutliple viewer Example, the model on the right has black texture, and sometimes doesn't draw the model

=> perhaps the applyTextureAttributeMaps() has to be done too in the state::applyDefault() to clean some osgjs state info

jtorresfabra:
Adding  this.applyTextureAttributeMapList( this.textureAttributeMapList ); just fixes the issue.

stephomi:
adding back the texture apply

PR #433, opened 2015-09-21
[#103208722] Reverts the not binding texture null

Kuranes:
Details here: https://code.google.com/p/chromium/issues/detail?id=533345&thanks=533345&ts=1442573130

### DIFF
--- a/sources/osg/State.js
+++ b/sources/osg/State.js
@@ -383,13 +383,16 @@ define( [
 
                 if ( attribute.apply ) {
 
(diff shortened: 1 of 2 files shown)


cedricpinson:
Thx could you point to the example that generated too many bind ? just to have an overview of the improvement if we decide to optimize it

Kuranes:
https://sketchfab.com/models/3c8813e9937a48caa1543eba94c307c5/edit?autostart=1
(or task for stats on 3 sample models with before/after gl trace count )

PR #436, opened 2015-09-25
Adds screenshot tutorial

Kuranes:
Using Camera FinalDrawCallBack mechanism

Now can take screenshot using glReadPixel, freeing the need for
preserveDrawingBuffer performance impacts
And do other operation (texture copy or oth) on each camera final draw


Kuranes:
Depends on https://github.com/sketchfab/workflow-helpers/pull/2

PR #441, opened 2015-09-29
Adds ShadowCasterVisitor as easy override by user

Kuranes:
Allows either override with user provided visitor, false (disable any visitor) or defaut visitor (which handles lights, subscene with rtt, transparent nodes)

### DIFF
--- a/sources/osgShadow/ShadowMap.js
+++ b/sources/osgShadow/ShadowMap.js
@@ -23,98 +23,14 @@ define( [
     'osg/Viewport',
     'osgShader/ShaderProcessor',
     'osgShadow/ShadowReceiveAttribute',
+    'osgShadow/ShadowCasterVisitor',
     'osgShadow/ShadowFrustumIntersection',
     'osgShadow/ShadowCastAttribute',
     'osgShadow/ShadowTechnique',
     'osgShadow/ShadowTexture'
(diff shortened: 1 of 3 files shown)


sources/osgShadow/ShadowMap.js
```diff
@@ -236,7 +152,14 @@ define( [
 
         // Overridable Visitor so that user can override the visitor to enable disable
         // in its own shadowmap implementation
-        this._removeNodesNeverCastingVisitor = new RemoveNodesNeverCastingVisitor( this._castsShadowTraversalMask );
+        // settings.userShadowCasterVisitor:
+        // - undefined means using default
+        // - false means no removal visitor needed
+        // - otherwiser must be an instance of a class inherited from shadowCaster
+        if ( settings.userShadowCasterVisitor !== false ) {
```
Kuranes: As discussed before, settings is considerer a dictionnary not an object.
But can add an accessor if feeling is strong on that one.

cedricpinson:
do we need this feature now ?

Kuranes:
It fixes a bug in transparent shadow casting ?

cedricpinson:
I was talking about the visitor stuff, but ok never mind

Kuranes:
On osgjs side we still need the visitor to remove transparent material otherwise default rendering on par with osgjs material won't give good results.

 ( it doesn't add new code, the visitor is the old one we had before. it's just allowing user control/overwrite of that part )

PR #443, opened 2015-09-30
[#101016480] Adds animation skinning + morph 

cedricpinson:


sources/osgShader/node/morph.js
```diff
@@ -0,0 +1,74 @@
...
+            return str;
+
+        },
+
+        computeShader: function () {
+
+            var inputs = [ this._inputs.weights, this._inputs.vertex ];
+
+            for ( var i = 0; i < 4; i++ ) {
+                var index = i;
```
stephomi: Why not use i directly?

sources/osgShader/Compiler.js
```diff
@@ -1179,19 +1214,19 @@ define( [
...
+            var posOut = this.createVariable( 'vec3', 'normalAttribute' );
 
-            this.getNode( 'MatrixMultDirection' ).setInverse( true ).inputs( {
-                matrix: this.getOrCreateBoneMatrix(),
-                vec: inputNormal
-            } ).outputs( {
-                vec: normalAnimated
-            } );
-            return normalAnimated;
+            if ( this._morphAttribute && !this._skinningAttribute )
```
stephomi: should the output normal be normalized after morph?

sources/osgShader/node/morph.js
cedricpinson: typo before it used i + 1, I will remove it

PR #448, opened 2015-10-03
Feature/zero allocation render loop 104794834

Kuranes:
Prevent any repeating matrix allocation each frame by keeping stack whenever it makes sense or just plain removes it when not needed

### DIFF
--- a/sources/osg/Matrix.js
+++ b/sources/osg/Matrix.js
@@ -1224,43 +1224,29 @@ define( [
             };
         } )(),
 
-        multScale: function ( x, y, z, m, res ) {
+        preMultScale: function ( m, scale ) {
 
-            if ( !res ) {
+            var x = scale[ 0 ];
+            var y = scale[ 1 ];
+            var z = scale[ 2 ];
 
-                Notify.warn( 'no matrix destination !' );
-                res = Matrix.create();
+            m[ 0 ] *= x;
+            m[ 4 ] *= y;
+            m[ 8 ] *= z;
 
-            }
+            m[ 1 ] *= x;
+            m[ 5 ] *= y;
+            m[ 9 ] *= z;
 
-            // Optimize real case
-            // which is in place scale
-            // less often scale to new matrix
-            // is less optimized
-            if ( m !== res ) {
+            m[ 2 ] *= x;
+            m[ 6 ] *= y;
+            m[ 10 ] *= z;
 
-                // fill CPU cache L1
-                Matrix.copy( m, res );
+            m[ 3 ] *= x;
+            m[ 7 ] *= y;
+            m[ 11 ] *= z;
 
-            }
-
-            res[ 0 ] *= x;
-            res[ 4 ] *= y;
-            res[ 8 ] *= z;
-
-            res[ 1 ] *= x;
-            res[ 5 ] *= y;
-            res[ 9 ] *= z;
-
-            res[ 2 ] *= x;
-            res[ 6 ] *= y;
-            res[ 10 ] *= z;
-
-            res[ 3 ] *= x;
-            res[ 7 ] *= y;
-            res[ 11 ] *= z;
-
-            return res;
+            return m;
 
         },
 
--- a/sources/osg/Matrix.js
+++ b/sources/osg/Matrix.js
@@ -1224,6 +1224,46 @@ define( [
             };
         } )(),
 
+        multScale: function ( x, y, z, m, res ) {
+
+            if ( !res ) {
+
+                Notify.warn( 'no matrix destination !' );
+                res = Matrix.create();
+
+            }
+
+            // Optimize real case
+            // which is in place scale
+            // less often scale to new matrix
+            // is less optimized
+            if ( m !== res ) {
+
+                // fill CPU cache L1
+                Matrix.copy( m, res );
+
+            }
+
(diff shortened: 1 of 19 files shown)


sources/osg/Matrix.js
```diff
@@ -1223,6 +1223,34 @@ define( [
             };
         } )(),
 
+        scale: function ( x, y, z, m, res ) {
+            if ( res === undefined ) {
```
stephomi: !res ?

sources/osg/Matrix.js
```diff
@@ -1223,6 +1223,34 @@ define( [
...
+                Notify.warn( 'no matrix destination !' );
+                res = Matrix.create();
+            }
+            // fill CPU cache L1
+            Matrix.copy( m, res );
```
stephomi: I think it's best if we stick with the glmatrix version...
Plus... we are writing twice in the res array (and we are reading twice the same values too, once in the copy call and once again with the ``res[ x ] *=`` ).

sources/osg/Node.js
```diff
@@ -341,10 +365,41 @@ define( [
...
+
+            var m = Node.__reserveMatrixStack[ Node._reserveMatrixStack.current++ ];
+
+            if ( Node._reserveMatrixStack.current === Node._reserveMatrixStack.length ) {
+
+                Node._reserveMatrixStack.push( Matrix.create() );
+
+            } else {
+
+                Matrix.setIdentity( m );
```
stephomi: makeIdentity ?...

Kuranes:
Question left is MatrixClass for this PR or another one ?

cedricpinson:
It would be a good candidate to do a MatrixPool now I think, if you want we can talk about it

Kuranes:
ok.

cedricpinson:
Maybe MatrixPool is a better name ?

Kuranes:
MatrixMemoryPool then would be clearer as Pool might not be that clear ?
( indeed matrixstack can confuse with other matrix stack in osg )

cedricpinson:
needs instructions to test and demonstrate the improvement

Kuranes:
Yes, putting back need fixing as there ismultiple improvements  and each need scene test to be build by code, and some fairly complex to test.

For now, juste have the testing on frustum Example with ?count=25&starts=1 shows a Culltime from 45 to 30ms. which shows clearly a matrixstack usage improvement of computeWorldMatrix martrixmemorypool.

For the others
- Shadow Map improvements more complex, have to make a test scene with many nodes + shadow to see the improvements on computebounds. 
- StackedScale change need animations with many scale track to  test

cedricpinson:
We dont need a full report on all example just some sample scene built with code that can be executed easily

Kuranes:
```
Perf:                Before    After
Basic                254        140
Cull                 179        104
Shadow               425        408
```

cedricpinson:
What's the number means ? how do you get them ?

tests/osg/Matrix.js
```diff
@@ -1,9 +1,19 @@
 define( [
     'qunit',
     'tests/mockup/mockup',
+    'osg/Light',
```
cedricpinson: it's not the place to add more high level tests. Test with the same class name should unittest the class. For higher level test with sample scene or other should be in another directory.

cedricpinson:
We can tomorrow to make the review together

Kuranes:
@cedricpinson Pushed latest form, with its own benchmarks folder,

sources/osg/Matrix.js
cedricpinson: !fixed

sources/osg/Matrix.js
cedricpinson: !fixed

sources/osg/Node.js
cedricpinson: !fixed

tests/osg/Matrix.js
cedricpinson: !fixed

cedricpinson: !fix

cedricpinson: !fixed

sources/osg/Matrix.js
```diff
@@ -1223,6 +1223,46 @@ define( [
             };
         } )(),
 
+        multScale: function ( x, y, z, m, res ) {
+
+            if ( !res ) {
```
cedricpinson: Remove extra test it should never be called without res !fix

Kuranes: k, so we rename it preMultScale for consistency and we'll make the multScale with a matrix result != m .

Kuranes: !fixed

sources/osg/Matrix.js
```diff
@@ -1224,6 +1224,28 @@ define( [
             };
         } )(),
 
+        preMultScale: function ( x, y, z, m ) {
```
cedricpinson: Check how it's done for preMultTranslate, the order argument is not the same, matrix argument is always first.
The other point should be maybe to change premulttranslate to use it without array but scalar instead.
Or maybe we can do it in another commit. But for consistent both method should work the same !fix

Kuranes: !fixed

PR #449, opened 2015-10-04
Converts StateAttribute hash from string to cached hash int keys

Kuranes:
Reducing further the GC work by not allocating nor concatenating strings each loop for shadergeneration when nothing does change

### DIFF
--- a/sources/osg/StateAttribute.js
+++ b/sources/osg/StateAttribute.js
@@ -1,14 +1,23 @@
 define( [
+    'osg/Hash',
     'osg/Utils',
     'osg/Object'
-], function ( MACROUTILS, Object ) {
+], function ( Hash, MACROUTILS, Object ) {
 
     'use strict';
 
     var StateAttribute = function () {
         Object.call( this );
         this._dirty = true;
+        this._dirtyHash = true;
+
+        // default hash in case of missing getHashString override
+        // moreoften than not must also be called from
+        // inherited classes ctor at end of the ctor (once getHashString makes sense)
+        this._hash = Hash.hashComputeCodeFromString( this.attributeType );
+
     };
+
 
     StateAttribute.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( Object.prototype, {
 
@@ -16,25 +25,32 @@ define( [
             return this._dirty;
         },
 
+        // when the StateAttribute State changes.
+        // ( NOT when uniform associated change )
         dirty: function () {
             this._dirty = true;
+            this._dirtyHash = true;
         },
 
         setDirty: function ( dirty ) {
+
+            // this._dirtyHash or only if it WAS dirty and won't be anymore now
+            this._dirtyHash = this._dirtyHash || ( this._dirty && !dirty );
             this._dirty = dirty;
+
         },
 
         getType: function () {
             return this.attributeType;
         },
 
-        // basically, if you want two StateAttribute with the same attributeType in a stateSet/State
-        // their typeMember should be different
         getTypeMember: function () {
             return this.attributeType;
         },
 
-        apply: function () {},
+        apply: function () {
+            this.setDirty( false );
+        },
 
         // getHash is used by the compiler to know if a change in a StateAttribute
         // must trigger a shader build
@@ -43,9 +59,21 @@ define( [
         // but if you change a type or representation of your StateAttribute, then it should
         // if it impact the rendering.
         // check other attributes for examples
-        getHash: function () {
+        getHashString: function () {
             return this.getTypeMember();
+        },
+        // return int hash
+        getHash: function () {
+
+            if ( this._dirtyHash ) {
+                this._hash = Hash.hashComputeCodeFromString( this.getHashString() );
+                this._dirtyHash = false;
+            }
+
+            return this._hash;
+
         }
+
 
     } ), 'osg', 'StateAttribute' );
 
--- a/sources/osg/Utils.js
+++ b/sources/osg/Utils.js
@@ -226,5 +226,6 @@ define( [
 
     } )();
 
+
     return Utils;
 } );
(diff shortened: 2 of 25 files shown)


sources/osgShader/ShaderGenerator.js
```diff
@@ -105,18 +107,17 @@ define( [
                     continue;
 
                 if ( attr.getHash ) {
-                    hash += attr.getHash();
+                    this.pushToHashStack( attr.getHash() );
                 } else {
-                    hash += attr.getType();
+                    this.pushToHashStack( attr.hashComputeCode( attr.getType() ) );
```
stephomi: hashComputeCodeFromString

sources/osgShader/ShaderGenerator.js
```diff
@@ -145,14 +146,13 @@ define( [
                         continue;
 
                     if ( attr.getHash ) {
-                        hash += attr.getHash();
+                        this.pushToHashStack( attr.getHash() );
                     } else {
-                        hash += attr.getType();
+                        this.pushToHashStack( attr.hashComputeCode( attr.getType() ) );
```
stephomi: hashComputeCodeFromString

sources/osg/Utils.js
```diff
@@ -226,5 +226,59 @@ define( [
...
+    Utils.hashComputeCodeFromIntList = function ( intList ) {
+
+        var hash = 0;
+        var intAt = 0;
+
+        if ( intList === undefined || intList.length === 0 ) {
+            return hash;
+        }
+
+        for ( var i = 0; i < intList.length; i++ ) {
```
stephomi: Always cache length...
Shouldn't it be intList.current ?
It's not a list but an array so maybe hashComputeCodeFromIntArray

sources/osg/Utils.js
```diff
@@ -226,5 +226,59 @@ define( [
...
+    Utils.hashComputeCodeFromString = function ( str ) {
+
+        var hash = 0;
+        var chara = 0;
+
+        if ( str === undefined || str.length === 0 ) {
+            return hash;
+        }
+
+        for ( var i = 0; i < str.length; i++ ) {
```
stephomi: cache length

sources/osg/StateAttribute.js
```diff
@@ -33,7 +37,9 @@ define( [
             return this.attributeType;
         },
 
-        apply: function () {},
+        apply: function () {
+            this.dirty( false );
```
stephomi: I don't think I like that... we compute a code from a string (by iterating on a string) each time we apply an attribute.

We apply way more attribute than we generate a shader I think...

+ we have to make sure all the other apply will call the original stateAttribute apply (error prone?)

sources/osgShader/ShaderGenerator.js
```diff
@@ -203,13 +215,17 @@ define( [
 
             return function ( state ) {
                 // extract valid attributes
-                var hash = '';
+
+                this._hashStack.current = 0;
```
stephomi: It's bad practice to add attribute to an array.
Add an additional variable if needed.
Also... can the array be a global (global to all the ShaderGenerator)?

Anyway in that case I don't think it's super useful to avoid allocation...
We only push to this hash on a rendering leaf... (so typically a small number) and we only push a small of a stuffs (number of attributes).

I did some test a while ago by using a similar array workaround to avoid push/pop... and for relatively small array, I didn't notice any gain (probably because of the similar pushToHashStack and because js VM handles well small array).
But still, I'm against useless push/pop :).

sources/osg/Utils.js
```diff
@@ -226,5 +226,59 @@ define( [
 
     } )();
 
+    Utils.hashComputeCodeFromString = function ( str ) {
+
+        var hash = 0;
+        var chara = 0;
+
+        if ( str === undefined || str.length === 0 ) {
```
stephomi: if ( !str ) do the same

sources/osg/Utils.js
```diff
@@ -226,5 +226,59 @@ define( [
...
+        return hash;
+
+    };
+
+    Utils.hashComputeCodeFromIntList = function ( intList ) {
+
+        var hash = 0;
+        var intAt = 0;
+
+        if ( intList === undefined || intList.length === 0 ) {
```
stephomi: Maybe ``if ( !intList || intList.length === 0 )`` because if it's null ``intList.length`` won't fail but....

@cedricpinson Can we agree on ``!obj`` vs ``obj === undefined`` when we test an **object**?
I often made this comment but I don't know if we agreed or not on this practice

stephomi:
Do you have a specific case/example where we win performances because I don't really think the hash building and test has ever been a bottleneck due to string concatenation (since it's only done per shader creation / draw leaf).

Also, we still do string concatenation because of the ``dirty( false )`` call in ``apply``

sources/osgShader/ShaderGenerator.js
Kuranes: indeed.
Yet I never ever encountered something going in this branch, and don't understand why it's there.

sources/osgShader/ShaderGenerator.js
Kuranes: Good point about that array member added. I think I checked how we did it in cullStack but maybe I misread.

Array could be global too, why not indeed.

Rationale and results are explained in the pivotal task, but anyway:

tl;dr: getHash does pop regularly on js profiler, this patch does reduce by half GC firefox in the boxe: https://sketchfab.com/models/663db593a73e4ceabf21a457f5ebcb26 and was my guess on the matcap/shadeless thing (as it's the only real difference between the two)

it's called each frame, very often (the cache flatten lots of not "flattened out" stateset like it theoritically could), and it'll it's all stateattributes which in turn does string concat of it parameter each frame + all texture attributes stack (notice the "stack")

sources/osg/StateAttribute.js
stephomi: setDirty( false ) ?

Kuranes:
useful for diffing the rebase of death

PR #451, opened 2015-10-06
Avoid rendering empty text. Eslint fixes in Text.js

jtorresfabra:


cedricpinson:
I updated the eslint rules to removes the annoying ones

PR #454, opened 2015-10-09
Makes WebGLCaps a singleton

Kuranes:
Makes it now viewer independant, can be called even before osg viewer
initialisation, uses its own canvas.
Viewer still needs to enable extension for their canvas
Adds platform information to start degrading gracefull only on defect gpu/os/browser instead of degrading everyone.

### DIFF
--- a/sources/osg/WebGLCaps.js
+++ b/sources/osg/WebGLCaps.js
@@ -1,23 +1,94 @@
 define( [
-    'osg/Texture'
-], function ( Texture ) {
+    'osg/Texture',
+    'osgViewer/webgl-utils'
+], function ( Texture, WebGLUtils ) {
 
     'use strict';
 
-    var WebGLCaps = function ( gl ) {
-        this._gl = gl;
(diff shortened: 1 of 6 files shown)


sources/osg/WebGLCaps.js
```diff
@@ -31,12 +73,16 @@ define( [
...
+
             var gl = this._gl;
-            if ( gl === undefined )
+            if ( !gl )
                 return false;
             var key = typeFloat + ',' + typeTexture;
-            if ( this._checkRTT[ key ] !== undefined )
+
+            // check once only
+            if ( this._checkRTT[ key ] )
```
stephomi: We need the !== undefined here because _checkRTT is a boolean

Kuranes: k, thx, btw about using the  Boolean() instead ? advantage is that eslint does check for it, and it's primed for visibiliy 

 !!foo => Boolean(foo); http://eslint.org/docs/rules/no-implicit-coercion.html

PR #461, opened 2015-10-16
Handles more than 4 morph targets

stephomi:


sources/osgAnimation/UpdateMorph.js
```diff
@@ -47,21 +56,174 @@ define( [
...
+                morphExtraTargets[ i ] = 0.0;
+            }
+
+            var targets = this._targets;
+            for ( var j = 0, nb = targets.length; j < nb; ++j ) {
+                var weight = targets[ j ].value;
+                if ( gpuMorphed[ j ] === true ) continue; // ignore gpu morphed targets
+                if ( Math.abs( weight ) < WEIGHT_CPU_EPS ) continue;
+
+                weight /= extraWeightSum;
```
Kuranes: No chance that extraWeightSum could be near 0 here ?

sources/osgAnimation/UpdateMorph.js
```diff
@@ -47,21 +56,174 @@ define( [
...
+                if ( Math.abs( weight ) < WEIGHT_CPU_EPS ) continue;
+                sum += weight;
+            }
+            return sum;
+        },
+
+        _morphBufferArrayCPU: function () {
+
+            // compute new weights of the 4th target (all the extra target will be merged inside this one)
+            var extraWeightSum = this._weights[ 3 ] = this._computeExtraWeightsSum();
```
Kuranes: if 0 here, Early Out

sources/osgAnimation/UpdateMorph.js
```diff
@@ -47,21 +56,174 @@ define( [
...
+
+                for ( var j = 0, nbMorphs = morphs.length; j < nbMorphs; ++j ) {
+
+                    var morph = morphs[ j ];
+                    var vAttrs = morph.getVertexAttributeList();
+                    var morphNames = morph.getMorphTargetNames();
+                    for ( var k = 0, nbNames = morphNames.length; k < nbNames; ++k ) {
+
+                        var attName = morphNames[ k ];
+                        vAttrs[ attName + '_' + i ].setActiveArray( vAttrs[ attName + '_' + index ] );
```
Kuranes: "IF" it's done each frame, caching "'_'+i" and "'_'+index" two loop above can save a bit of gc/string/itoa compute

sources/osgAnimation/UpdateMorph.js
stephomi: It can be "near 0" but always >= WEIGHT_CPU_EPS 
the same code is used to compute the extra weight https://github.com/cedricpinson/osgjs/blob/morph-many-targets/sources/osgAnimation/UpdateMorph.js#L145-L147

si it should be safe

sources/osgAnimation/UpdateMorph.js
stephomi: it can't be zero, at least we have https://github.com/cedricpinson/osgjs/blob/morph-many-targets/sources/osgAnimation/UpdateMorph.js#L214

sources/osgAnimation/UpdateMorph.js
stephomi: indeed +1

PR #470, opened 2015-10-26
Optimises updateTargetType when blend is useless

stephomi:


Kuranes:
Check on basicanimation benchmark in ms.

Machine              Before   After
windowsCPU      1727          7
androidnexus7    16989        73

(going to double check!)

PR #471, opened 2015-10-26
Uses a faster isNaN function

stephomi:


cedricpinson:
tests fails in vec2 not sure why check the log in travis details

stephomi:
Fixed it.

(the test was checking for [ 'a', 0 ] validity.
I think it's safe to assume that Vec3, Vec4, etc will only contains number.

cedricpinson:
awesome

PR #475, opened 2015-10-27
Fix mousewheel

stephomi:
Tested on chrome, IE and firefox and it works (mousewheel)

### DIFF
--- a/sources/osgUtil/DisplayNodeGraphVisitor.js
+++ b/sources/osgUtil/DisplayNodeGraphVisitor.js
@@ -1,8 +1,9 @@
 define( [
     'osg/Utils',
-    'osg/NodeVisitor'
+    'osg/NodeVisitor',
+    'osgAnimation/MorphGeometry'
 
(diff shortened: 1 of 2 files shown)


stephomi:
@jtorresfabra @Kuranes

sources/osgUtil/DisplayNodeGraphVisitor.js
```diff
@@ -130,7 +130,11 @@ define( [
                 }
             }
 
+            if ( node.getSourceGeometry ) node.addChild( node.getSourceGeometry() );
+
             this.traverse( node );
+
```
cedricpinson: wtf ?

stephomi: It's temporary.
It's just one way to see if we have a morph or not when using the debug node graph tool.

stephomi: ok, don't merge I'll do something a bit better

PR #481, opened 2015-11-03
Feature/hashstring to hashint gc dev 5588

Kuranes:
- Converts StateAttribute hash from string to cached hash int keys	
- Add webpack-strip debug functions (activated by default)
- Fixes incorrect develblock (very rigid format must be followed)	
- Adds "one time" log limitations to limit log perf impact (non inversible matrix)	
- Adds comment on "computeCanvasSize" reasons	
- Adds Whole Frame benchmark with complete Webgl mockup (return value when needed to ensure no init-loop, etc.)


review-ninja:
<a href="https://app.review.ninja/cedricpinson/osgjs/pull/481" target="_blank">ReviewNinja

Kuranes:
(need the bench thing)

Kuranes:
Note: Needs heavy PR review discussion.
Notably the Dirty() and setDirty dichotomy
* previous state
  * 2 method doing the same thing
* New state
 * 2 method doing different thing but with misleading names

stephomi:
I don't know when if/this PR will be merged but if the last 4 commits are important I guess they could be merged quickly if done in another separate PR.

As for the setDirty maybe it's best to renames the dirty methods... the behaviour of the setDirty is hard to understand and very implicit, we should speak about it I guess.

cedricpinson:
in osg dirty means that some code needs to (re)evaluate something. Like regenerating texture filtering option, recompiling buffer if you changed the data content. The dirty flag can automatically triggered by functions or manually by the user.
It can't related with the apply for StateAttribute like for texture but not always, for example for Material we always need to send uniforms so dirty does not make sense in this case.

It's a complex PR, we will need to talk about it tomorrow

PR #482, opened 2015-11-03
Adds debug graph to display internal system rendering

stephomi:
It breaks the compatibility, but still very easy to use (cf main.js)


review-ninja:
<a href="https://app.review.ninja/cedricpinson/osgjs/pull/482" target="_blank">ReviewNinja

PR #489, opened 2015-11-10
Feature/animation lerp loop

stephomi:


cedricpinson:
Could you adds extra tests, it's typically something that could be break by another developer without knowing it.

PR #490, opened 2015-11-10
[#DEV-6256] Adds visitors benchs & matrixpool

Kuranes:


Kuranes:
first round of memory pooling, on IntersectionVisitor
```
Before          After
1082ms        720ms
```

Kuranes:
Note that mobile annotations are very ugly it's 50ms per query...

On the TODO *real optimisations* of the picking, :
* optimise current kdtree code : 
** 40% cpu profile is matrix.inverse, but maybe it's not so much needed (like mult mat and inverse only on finale matrix results usage
** "same frame" matrix computation reuse as most matrices doesn't need recompute (nodepath multiparent thing) 
* other classic intersection library optimisation 
** cache results (two consecutive frame with same queries ins't that rare)
** use last result "neareast ray" as basis to select the node to start with (aka restart from nearest tree leaf) 
** etc

cedricpinson:
Your measure is only the picking ? like is this annotation visible ?

Kuranes:
No visibility check in osgjs afaik

cedricpinson:
ok thx

stephomi:
> Note that mobile annotations are very ugly it's 50ms per query, 3ms per query on top notch desktop is huge too...

How many nodes in the scenegraph? Number of triangles?
I assume the optimisation is not about the kdtree itself but all the stuffs that goes around that is done *per geometry*.

The kdtree itself was fast if I recall well (a few nodes but big geometries)
The PR is more about the Intersection visitor that the kdtree itself (there's no Matrix.inverse in the KdTree for example).
The intersection visitor (which does many stuffs and need to be optimised) can work without kdtree too.

Kuranes:
Yep, Agree. 
As said above ^^:
* "the real culprit is the osgjs node path( reducing number of TransformMatrix nodes reduce drastically the problem)" 
* "benchmark is done on a 25*25 nodes" with basic quad geoms, so not much benchmarking the kdtree itself

cedricpinson:
@jtorresfabra could you test if you see improvement for your usecase ?

PR #492, opened 2015-11-11
Adds Webgl2 Support When available

Kuranes:
Chrome and Firefox have 100% support but both behind browser flags, so we can just use webgl2 context when available. 
It's progressive enhancement, as it mocks webgl1 extension and capabilities check so that it's all
transparent osgjs side.
Adds a  webglCaps.isWebGL2() in case of webgl2 only need to check feature when we'll progessively add them (texturearray, texture3d, ubo, etc.)

### DIFF
--- a/sources/osg/Options.js
+++ b/sources/osg/Options.js
@@ -8,7 +8,8 @@ var OptionsDefault = {
     'enableFrustumCulling': false,
     'stats': false, // display canvas with stats for the viewer
     'statsNoGraph': false, // display only text
-    'scrollwheel': true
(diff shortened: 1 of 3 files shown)


cedricpinson:
Dont want to select by default webgl2 context, I prefer something that keep webgl1 by default and you add a flag to select webgl2. at least for now.

Kuranes:
True. Note that enabling webgl2 on browser is already under heavy flags.
But anyway, don't want to be surprised when it won't be anymore before we can react, true.

PR #495, opened 2015-11-17
Fixes resize support for oculus and cardboard

stephomi:
Obviously breaking change.
Maybe future tweaking might be need in the future depending of devices.


examples/dualcam/main.js
```diff
@@ -81,6 +88,8 @@
         }
         // Disable VR
         else {
+            viewer._eventProxy.WebVR._enable = false;
```
cedricpinson: Adds accessors instead

sources/osgUtil/WebVRCustom.js
```diff
@@ -174,6 +154,16 @@ var createOrthoRtt = function ( left, viewportSize, canvasSize, cardboard ) {
     Matrix.makeOrtho( -0.5, 0.5, -0.5, 0.5, -5, 5, orthoCamera.getProjectionMatrix() );
     orthoCamera.setRenderOrder( Camera.NESTED_RENDER, 0 );
     orthoCamera.setReferenceFrame( Transform.ABSOLUTE_RF );
+
+    var stateSet = orthoCamera.getOrCreateStateSet();
+    stateSet.setTextureAttributeAndModes( 0, texture );
+    stateSet.setAttributeAndModes( getWebVRShader() );
+    stateSet.addUniform( new Uniform.createFloat2( webVRUniforms.scale, 'uScale' ) );
```
cedricpinson: should Uniform.createFloat2 not *new* Uniform.createFloat2

cedricpinson:
We could probably set a custom render stage in the futur to cull only once the scene and adjust only the rendering RTT

PR #496, opened 2015-11-17
Feature/fullrenderer mock dev 6297

Kuranes:
- Fixes incorrect develblock
- Adds "one time" log limitations to avoid mes log repetitions cluttering and slowing
- Adds comments & link on computeCanvasSize reasons	
- Adds Whole Frame benchmark with full webgl mock


sources/osg/Notify.js
```diff
@@ -48,14 +48,22 @@ function getStackTrace( err ) {
     return callstack;
 }
 
+var logCache = new window.Map();
```
cedricpinson: I dont want to have this generalized, It's specific when we want to warn the user only once. So It should be implemented on specific case.
eg you load a file and print a warning, you load a second file you will not have because of this.

Kuranes:
Sorry, I had problem with rebase on the cjs previous PR and seems there is lots of code that shouldn't be there !
( There shouldn't be any change in shadergenerator nor framebufferobjects ! )

Kuranes:
@jtorresfabra fbo and rbo delection code already changed in a previous commit and is in develop now: https://github.com/cedricpinson/osgjs/commit/2b6c038e12b6bf83fc3237d45bb500b39b39adef

jtorresfabra:
@Kuranes I see, anyway if you don't call RenderBufferObject.flushDeletedGLRenderBuffers() in https://github.com/cedricpinson/osgjs/blob/develop/sources/osgViewer/View.js#L251, then it never gets deleted.

Kuranes:
yes, ok, I see. Thanks :+1:

Kuranes:
Added A PR for the missing flush, thanks again https://github.com/cedricpinson/osgjs/pull/497

sources/osg/CullVisitor.js
```diff
@@ -545,7 +545,12 @@ CullVisitor.prototype[ Geometry.typeID ] = ( function () {
             depth = this.distance( bb.center( tempVec ), modelview );
         }
         if ( osgMath.isNaN( depth ) ) {
-            Notify.warn( 'warning geometry has a NaN depth, ' + modelview + ' center ' + tempVec );
+
+            if ( !loggedOnce ) {
```
cedricpinson: What you want actually is a loggedOncePerFrame, it can be done using the frame stamp in the visitor

Kuranes: Imho, Best We have a Verbal Assisted Reunion
Otherwise I feare a neverending discussion thread

sources/osg/Matrix.js
```diff
@@ -957,9 +957,10 @@ var Matrix = {
             // Compute determinant of rot from 3 elements just computed
             var det = ( r00 * result[ 0 ] + r10 * result[ 1 ] + r20 * result[ 2 ] );
             if ( Math.abs( det ) < Number.EPSILON ) {
-                Notify.log( 'Warning can\'t inverse matrix ' + matrix );
+                Notify.log( 'Warning can\'t inverse matrix ' + matrix, false, true );
```
cedricpinson: can be changed to Notify.info

PR #497, opened 2015-11-17
Adds missing RenderBufferFlush Call

Kuranes:


sources/osgViewer/View.js
```diff
@@ -256,6 +256,7 @@ View.prototype = {
         availableTime = Program.flushDeletedGLPrograms( gl, availableTime );
         availableTime = Shader.flushDeletedGLShaders( gl, availableTime );
         availableTime = FrameBufferObject.flushDeletedGLFrameBuffers( gl, availableTime );
+        availableTime = FrameBufferObject.flushDeletedGLRenderBuffers( gl, availableTime );
```
jtorresfabra: You mean RenderBufferObject no? ;)

Kuranes: yes :) needs fixing

Kuranes:
Finally used FrameBufferObject as it's not really clear as the current code is which is responsible for what.
RenderBufferObject wasn't really a class on its own, needing its own js file, as it's handled really by the FrameBufferObjectClass.

(fbo is under rework in the PR about renderbin RTT, and will soon be under more work again, multidraw buffer ftw !)

jtorresfabra:
Great to hear that! :+1:

Kuranes:
# _**Webgl2 is coming**_

PR #498, opened 2015-11-18
Clean dirty usage

cedricpinson:


jtorresfabra:
@cedricpinson about the mosaic of all examples I think it is now a little overkill. 
As a first step is ok, but I believe it would be better to have multiple viewers instead of iframes. So you can viewer.setDone(true) when you are not hovering, and viceversa. So frame it's not called in a viewer if you are not hovering that viewer in the mosaic (like shadertoy).

cedricpinson:
@jtorresfabra agree, it's just a trap in order someone else improve the system :)

sources/osgUtil/NodeGizmo.js
```diff
@@ -599,7 +599,13 @@ NodeGizmo.prototype = MACROUTILS.objectInherit( MatrixTransform.prototype, {
     getCanvasPositionFromWorldPoint: function ( worldPoint, out ) {
         var cam = this._viewer.getCamera();
         var mat = Matrix.create();
-        Matrix.preMult( mat, cam.getViewport() ? cam.getViewport().computeWindowMatrix() : Matrix.create() );
+
+        if ( cam.getViewport() ) {
```
Kuranes: Would have been a good occasion for closuring all thos temporary variables...
Perhaps that function could be in View or Viewer
(duplicate here https://github.com/sketchfab/showwebgl/commit/cde9ca749051e3ea9bb7346ba5444554b5c335b5 )

sources/osg/CullFace.js
```diff
@@ -22,28 +19,32 @@ CullFace.FRONT_AND_BACK = 0x0408;
...
     },
+
     getMode: function () {
         return this._mode;
     },
+
     apply: function ( state ) {
         var gl = state.getGraphicContext();
-        if ( this._mode === CullFace.DISABLE ) {
+        if ( !this._mode ) {
```
Kuranes: Myself not very fan of undocumenting code for a very small gain.
( as in Enum Use should be value independant. )

sources/osg/Program.js
```diff
@@ -82,6 +94,14 @@ Program.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( GLO
         return new Program();
     },
 
+    dirty: function () {
```
Kuranes: Why only program and not StateAttribute get the dirty() and isDirty() accessor ?

sources/osg/Uniform.js
```diff
@@ -10,11 +10,13 @@ var Vec4 = require( 'osg/Vec4' );
...
-var Uniform = function () {
-    this.transpose = false;
-    this._dirty = true;
-    this.name = '';
-    this.type = undefined;
+var Uniform = function ( name ) {
+    this._transpose = false;
+    this._glCall = '';
+    this._name = name;
+    this.name = name;
```
Kuranes: why two names ?

cedricpinson: it was temporary while used in State.js no need anymore it will be removed

sources/osgUtil/NodeGizmo.js
cedricpinson: agree will do

sources/osg/CullFace.js
cedricpinson: yep

sources/osg/Program.js
cedricpinson: there is no rules, some resources needs to be re uploaded some not. A StateAttribute is not always a resources and it depends how it's used. So for Texture it makes sense but not for CullFace. I tried to generalize but it does not make sense. So let's say that it depends on context/usage

cedricpinson:
@Kuranes fixed

sources/osg/Uniform.js
```diff
@@ -26,395 +27,217 @@ Uniform.isUniform = function ( obj ) {
...
+            gl[ this._glCall ]( location, this._data );
     },
-    applyMatrix: function ( gl, location ) {
-        if ( this._dirty ) {
-            this.update.call( this.glData, this.data );
-            this._dirty = false;
-        }
-        this.glCall( gl, location, this.transpose, this.glData );
+
+    setArray: function ( array ) {
```
Kuranes: names and interface feels strange for Float1 and *unique* value. 
But indeed it's the only way to have that  update "by reference"  instead of copy

sources/osg/Uniform.js
```diff
@@ -26,395 +27,217 @@ Uniform.isUniform = function ( obj ) {
...
-        } else {
-            this.data = value;
-        }
-        this.dirty();
-    };
-
-    uniform.name = name;
-    uniform.type = 'float';
+    uniform.setArray( value );
+    uniform._glCall = glSignature;
```
Kuranes: not sure but couldn't we store the function pointer instead ?
```
uniform._glCall = gl[glSignature].bind(gl);
```
So that we do avoid a hash search each call of the gl.uniform
(as said, not sure if any interest)

sources/osgShadow/ShadowTexture.js
```diff
@@ -117,19 +117,16 @@ ShadowTexture.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInheri
...
-        uniformMap.ProjectionMatrix.set( this._projectionMatrix );
-        uniformMap.DepthRange.set( this._depthRange );
-        uniformMap.MapSize.set( this._mapSize );
+        uniformMap.ViewMatrix.setArray( this._viewMatrix );
+        uniformMap.ProjectionMatrix.setArray( this._projectionMatrix );
+        uniformMap.DepthRange.setArray( this._depthRange );
+        uniformMap.MapSize.setArray( this._mapSize );
 
-        this.setDirty( false );
+        this._dirty = false;
```
Kuranes: why a dirty false here ?

cedricpinson: It's useless because the texture will be applied before and the dirty will be already false

sources/osg/Uniform.js
cedricpinson: you need gl, so can't do without the context. not sure it worth doing it at runtime. But a test would be better to validate this assumption

sources/osg/Uniform.js
cedricpinson: I changed to fit the osg api, setArray / getArray set the container on which other functions will copy their data.
On c++ you can override type so in C++ set is used. But because we can't do that in js, I did a set + the type. https://github.com/cedricpinson/osg/blob/master/include/osg/Uniform
I am open to other suggestions that fit osg API

Kuranes: Idea to convey would be along the lines of setArrayDirect or setDirect, etc.
(as it does makes clearer the opposition between the others way that does need copy at each value change)

cedricpinson: setInternalArray could do it

Kuranes: ok, with me. othewise could go the pedantic strict type "setDataReference"

PR #502, opened 2015-11-19
Simplifies the way to block the orbit y rotation

stephomi:
It does the same thing.


cedricpinson:
Did you check on example ? it does not seem to work

stephomi:
Fixed, typo on ```tmp``` instead of ```this._up```

PR #505, opened 2015-11-21
[#DEV-6320] Fixes MemeLeaks

Kuranes:
Cullsettings being allocated each camera cullvisits
CullingSet stack reserve "cursor" wasn't rersetted to 0

Those were the last vec2/3/4 & Matrix Leaks during render loop.

/!\ changes in polytope/planes

### DIFF
--- a/sources/osg/CullSettings.js
+++ b/sources/osg/CullSettings.js
@@ -1,34 +1,54 @@
 'use strict';
-var Vec3 = require( 'osg/Vec3' );
 
 
 var CullSettings = function () {
-    this._computeNearFar = true;
-    this._nearFarRatio = 0.005;
 
-    var lookVector = Vec3.createAndSet( 0.0, 0.0, -1.0 );
-    /*jshint bitwise: false */
-    this.bbCornerFar = ( lookVector[ 0 ] >= 0 ? 1 : 0 ) | ( lookVector[ 1 ] >= 0 ? 2 : 0 ) | ( lookVector[ 2 ] >= 0 ? 4 : 0 );
-    this.bbCornerNear = ( ~this.bbCornerFar ) & 7;
-    /*jshint bitwise: true */
-    this._enableFrustumCulling = false;
+    // Not doing a this.reset()
+    // because of multiple inheritance
+    // it will call the wrong reset
+    // cullstack reset for isntance()
+    CullSettings.prototype.reset.apply( this );
 
-    // who sets the parameter
-    // if it's cullvisitor
-    // it's an OVERRIDER for enableFrustumCulling
-    // allowing for global EnableFrustimCulling
-    this._settingsSourceOverrider = this;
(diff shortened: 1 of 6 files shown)


sources/osg/CullVisitor.js
```diff
@@ -50,6 +50,9 @@ var CullVisitor = function () {
     this._reserveLeafStack = [ new RenderLeaf() ];
     this._reserveLeafStack.current = 0;
 
+    this._reserveCullSettingsStack = [ new CullSettings() ];
```
cedricpinson: Why it is different from stack matrix ? we said that we dont want to add members in array

sources/osg/CullSettings.js
```diff
@@ -1,34 +1,54 @@
...
-    // allowing for global EnableFrustimCulling
-    this._settingsSourceOverrider = this;
-    //LOD bias for the CullVisitor to use
-    this._LODScale = 1.0;
+
+    // Not doing a this.reset()
+    // because of multiple inheritance
+    // it will call the wrong reset
+    // cullstack reset for isntance()
+    CullSettings.prototype.reset.apply( this );
```
cedricpinson: the initialization of fields before was recommended for best practice. @stephomi do you remember where we read this ?

cedricpinson:
very cool to eliminate leak

sources/osg/CullVisitor.js
Kuranes: we can make a "generic" pool object, just have to "standardize" the init/clear/reset so that we can make "stack" for any object (leaf, cullsettings, etc.) ?

cedricpinson: yep in the meantime split the current in another variable, and add a task to make a 'generic' stack object

sources/osg/CullSettings.js
cedricpinson: @Kuranes could you check this point ^ I think it comes from a google talk or best practice for performance

sources/osg/CullVisitor.js
Kuranes: k

sources/osg/CullSettings.js
Kuranes: Will look

Kuranes: http://blog.getify.com/sanity-check-object-creation-performance/#hidden-classes

Perf Wise, the idea should be something like in Ctor, make sure the "hidden class" get the correct size to alloc

So note that: 
- cullsettings creation time is not perf critical
- init call in the ctor does it.
- in fact we altogether try not to have critical creation time :) (no new/delete during loop)

Kuranes:
@jtorresfabra there is some changes in CullSettings/Polytope/Planes you might want to check

jtorresfabra:
@Kuranes I will do it, thanks for the advice.

PR #507, opened 2015-11-23
Example/pbr

cedricpinson:
Finally, still missing the article with details


cedricpinson:
@Kuranes if you want to do a quick review

examples/pbr/index.html
```diff
@@ -0,0 +1,64 @@
...
+    <!-- Javascript -->
+    <script type="text/javascript">
+    window.useExternalModels = false;
+    </script>
+    <script type="text/javascript" src="../vendors/jquery-2.1.1.js"></script>
+    <script type="text/javascript" src="../vendors/bluebird.js"></script>
+    <script type="text/javascript" src="../vendors/hammer.js"></script>
+    <script type="text/javascript" src="../vendors/dat.gui.js"></script>
+    <script type="text/javascript" src="../vendors/gunzip.min.js"></script>
+    <script type="text/javascript" src="../../builds/active/OSG.js"></script>
```
Kuranes: I think we do not have an "active" folder in lateset OSGJS

examples/pbr/shaders/tangentVertex.glsl
```diff
@@ -0,0 +1,34 @@
...
+void main(void) {
+
+    osg_FragEye = vec3(ModelViewMatrix * vec4(Vertex, 1.0));
+    osg_FragNormal = vec3(NormalMatrix * vec4(Normal, 0.0));
+    osg_FragVertex = Vertex;
+
+    osg_FragTangent = Tangent;
+    osg_FragTexCoord0 = TexCoord0;
+
+    gl_Position = ProjectionMatrix * ModelViewMatrix * vec4(Vertex,1.0);
```
Kuranes: nitpicking perf aside
``` 
ProjectionMatrix * (ModelViewMatrix * vec4(Vertex,1.0));
``` 
Otherwise it's Matrix Mult Matrix and then only vec4 Mult instead of vec4 Mult  then vec4 Mult

examples/pbr/main.js
```diff
@@ -0,0 +1,1411 @@
...
+
+            //var environment = 'textures/bus_garage5/';
+            //var environment = 'textures/walk_of_fame/';
+            //var environment = 'textures/airport/';
+            //var environment = 'textures/tmp/';
+
+            //var model = new ModelLoader( 'models/cerberus/' );
+
+            var modelPromises = [];
+            modelsPBR.forEach( function ( modelString, index ) {
```
Kuranes: shouldn't that be modelList ?

examples/pbr/main.js
```diff
@@ -0,0 +1,1411 @@
...
+    window.METALNESS_TEXTURE_UNIT = 4;
+    window.NORMAL_TEXTURE_UNIT = 5;
+    window.SPECULAR_TEXTURE_UNIT = 4;
+
+
+    window.formatList = [ 'FLOAT', 'RGBE', 'RGBM', 'LUV' ];
+
+    var modelsPBR = [ 'cerberus', 'c3po', 'devastator' ];
+
+    var modelList = [ 'sphere', 'model' ];
```
Kuranes: no config.json for those two ?

examples/pbr/main.js
```diff
@@ -0,0 +1,1411 @@
...
+            result[ 2 ] = b / 255.0;
+            //console.log( result );
+            return result;
+        }
+
+
+    };
+
+    // convert rgbe image to float texture
+    var TextureRGBEToFloatTexture = function ( texture, dest, textureTarget ) {
```
Kuranes: main.js starting to get very big... perhaps export more classes in separate js files

examples/pbr/main.js
```diff
@@ -0,0 +1,1411 @@
...
+    if ( window.useExternalModels ) {
+        modelList = modelList.concat( modelsPBR );
+    }
+
+
+    var Example = function () {
+
+        this._shaderPath = 'shaders/';
+
+        if ( optionsURL.colorEncoding )
```
Kuranes: Could derive from Example, if only to get OptionsURL for free as it's now part of Example, and the parameters ends in Example::_config.

(but surely can get other things like shaderprocessor, shader loads, etc)

examples/pbr/main.js
```diff
@@ -0,0 +1,1411 @@
...
+
+            group.addChild( this.createSampleModels() );
+
+            // add node that contains model loaded
+            group.addChild( this._proxyRealModel );
+
+            return group;
+        },
+
+        createShaderPanorama: function ( defines ) {
```
Kuranes: deriving from example template would allow all the "createShader*" funcs to be replaced with example::createShader https://github.com/cedricpinson/osgjs/blob/develop/examples/templates/Example.js#L159

examples/pbr/index.html
cedricpinson: will fix

examples/pbr/shaders/tangentVertex.glsl
cedricpinson: will fix

examples/pbr/main.js
cedricpinson: nop, we need to load only external pbr model, others a loaded differently. The var name should be externalModelPBR

examples/pbr/main.js
cedricpinson: nop internal model does not have config because no texture provided with them

examples/pbr/main.js
cedricpinson: agree but later

examples/pbr/main.js
cedricpinson: agree but later

examples/pbr/main.js
cedricpinson: agree  but later

PR #509, opened 2015-11-25
Small clean

stephomi:


sources/osgViewer/eventProxy/DeviceOrientation.js
```diff
@@ -49,6 +45,9 @@ DeviceOrientation.prototype = {
         if ( !this._enable )
             return false;
 
+        if ( !this._deviceOrientation )
```
cedricpinson: could you explain why this has been added ?

stephomi: Don't use the default orientation if the device don't support orientation
api.

sources/osg/Polytope.js
```diff
@@ -334,9 +334,9 @@ Polytope.prototype = MACROUTILS.objectInherit( Object.prototype, {
      * calculated elsewhere then use transformProvidingInverse() instead.
      * See http://www.worldserver.com/turk/computergraphics/NormalTransformations.pdf*/
     transform: function ( matrix ) {
-        var inverse = new Matrix();
-        inverse.invert( matrix );
-        this.transformProvidingInverse( inverse );
+        var inv = Matrix.create();
```
stephomi: @jtorresfabra this function was never called I guess (Matrix is not a function)

Not sure if we can re-use matrix or not.

Kuranes: *cough*cough* seems that might be me the code author, and afaicr the code can be removed now

PR #514, opened 2015-12-10
[DEV-6488] Robustifies osg.image.isReady

Kuranes:
Fixes previous osg Texture code which was using a very relaxed test to check if it has an HTML image or a OSG image inside, which gives incorrect results to user with ublock extension.

Ublock Extension add a HTML image "isReady" method, with exaclty the same name as osg Image leading to current code believing incorrectly that the "wrapped by ublock Image" is an osg Image, or is not a HTML image thus never loading it in a gpu texture.

Any inconsiderate browser extension wrapping the Image html element,
and there may be even more in the future when browser will freeze the
'natives' objects like Image, thus needs a "detection by feature" as
last resort of the isReady method.

### DIFF
--- a/sources/osg/Image.js
+++ b/sources/osg/Image.js
@@ -1,9 +1,10 @@
 'use strict';
+var Notify = require( 'osg/Notify' );
 var MACROUTILS = require( 'osg/Utils' );
 var Object = require( 'osg/Object' );
 
 
-var Image = function ( image ) {
+var ImageObject = function ( image ) {
     Object.call( this );
 
     this._imageObject = undefined;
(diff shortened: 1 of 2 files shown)


sources/osg/Image.js
```diff
@@ -196,8 +199,22 @@ Image.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( Objec
...
+        // by "feature" detect it
+        var imageTry = this.getImage();
+        if ( imageTry.complete ) {
+            if ( imageTry.naturalWidth !== undefined && imageTry.naturalWidth === 0 ) {
+                return false;
+            }
+            return true;
+        }
+
+        // no. it's not something we recognise
```
cedricpinson: yep add a warning plz

cedricpinson:
plz explain what this PR fix, it's easier when browsing history to understand why and how

PR #519, opened 2015-12-17
Morph box faster box

stephomi:
On the picking example with the highly tesselated sphere, the computation of box : 
- ~470 ms
- now, ~95ms

To give an overview of why "writing in array is slower" An alternate version (simpler but slower, ~145ms))
````
for ( var idx = 0, l = vertexes.length; idx < l; idx += itemSize ) {
    var v0 = vertexes[ idx ];
    var v1 = vertexes[ idx + 1 ];
    var v2 = vertexes[ idx + 2 ];
    if ( v1 > min[ 0 ] ) min[ 0 ] = v1;
    if ( v1 < max[ 0 ] ) max[ 0 ] = v1;
    if ( v2 > min[ 1 ] ) min[ 1 ] = v2;
    if ( v2 < max[ 1 ] ) max[ 1 ] = v2;
    if ( v3 > min[ 2 ] ) min[ 2 ] = v3;
    if ( v3 < max[ 2 ] ) max[ 2 ] = v3;
}
```

The 3 times are very consistent

### DIFF
--- a/sources/osgUtil/SphereIntersector.js
+++ b/sources/osgUtil/SphereIntersector.js
@@ -8,20 +8,24 @@ var SphereIntersector = function () {
     this._center = Vec3.create();
     this._iCenter = Vec3.create();
     this._radius = 1.0;
+    this._iRadius = 1.0;
     this._intersections = [];
 };
 
 SphereIntersector.prototype = {
     set: function ( center, radius ) {
+        // we copy iCenter and iRadius in case setCurrentTransformation is never called
         Vec3.copy( center, this._center );
-        this._radius = radius;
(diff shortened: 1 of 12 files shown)


sources/osgUtil/SphereIntersector.js
```diff
@@ -7,21 +7,24 @@ var TriangleSphereIntersector = require( 'osgUtil/TriangleSphereIntersector' );
 var SphereIntersector = function () {
     this._center = Vec3.create();
     this._iCenter = Vec3.create();
-    this._radius = 1.0;
+    this._iRadius = 1.0;
```
cedricpinson: this._radius is not declared but used in set

PR #521, opened 2015-12-21
Improves scale world computation for normal debug

stephomi:
Now it works if the matrix has some rotation in it too.
I don't handle the non-uniform scale case for now but it should be enough for debug.

### DIFF
--- a/sources/osgUtil/DisplayNormalVisitor.js
+++ b/sources/osgUtil/DisplayNormalVisitor.js
@@ -41,7 +41,7 @@ CompilerOffsetNormal.prototype = MACROUTILS.objectInherit( Compiler.prototype, {
 
         return [ frag ];
     },
-    _getOffsetVec: function () {
+    _getDirectionVec: function () {
         return this.getOrCreateNormalAttribute();
     },
     getOrCreateVertexAttribute: function () {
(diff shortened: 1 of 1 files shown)


sources/osgUtil/DisplayNormalVisitor.js
```diff
@@ -51,7 +51,7 @@ CompilerOffsetNormal.prototype = MACROUTILS.objectInherit( Compiler.prototype, {
         vertexOffset = this.createVariable( 'vec3', 'vertexOffset' );
 
         // normalize len (divide scale of worldmat, don't work for non uniform scale, but work enough for debug)
-        var str = '%out = %offset == 1.0 ? %vertex + normalize(%vecOffset.xyz) * %scale / %world[0][0] : %vertex;';
+        var str = '%out = %offset == 1.0 ? %vertex + normalize(%vecOffset.xyz) * %scale / length(%world[0].xyz) : %vertex;';
```
Kuranes: Nice

"vecOffset" parameter name is a bit misleading though if it's a direction we do normalize to compute the offset ?
Why not "vecOffsetDirection" or even directly "direction" ?
(vecOffset sounds like the result we want to get)

For non unif scale why not do that in world space ? ()
```
var str = '%out = %offset == 1.0 ? %world*vec4(%vertex.xyz,1.0) + normalize(world*vec4(%vecOffset.xyz,0.0)) * %scale / %vertex;';
```
and then using the out to compute View*vertexWorld instead of ModelView*vertex

stephomi: Yes, it's a direction (not normalized though)

getOrCreateVertexAttribute return the vertex in local space (morphed and not rigged), we can't simply return a world pos (need to hook another function)
+
we want to "normalize" the length of the debugged normal to be consistent in screen space.
We don't want it to be worldunit = 1

PR #523, opened 2016-01-13
Compute the bounding box for non influenced vertices of a RigGeometry

stephomi:
Not a big fan of the copy-paste, but there's no workaround as the part that needs to be changed is inside the for loop


cedricpinson:
Could you add comments in the code to explain why this code is here and what it solves

stephomi:
updated

PR #527, opened 2016-01-28
Handles UpdateMatrixTransform in a skeleton

stephomi:


sources/osgAnimation/AnimationUpdateCallback.js
```diff
@@ -12,6 +13,25 @@ var AnimationUpdateCallback = function () {
     Object.call( this );
 };
 
+// check if the path is animated, it could be elsewhere though
+AnimationUpdateCallback.checkPathIsAnimated = function ( path ) {
```
cedricpinson: why not in the prototype ?

cedricpinson: It's because it's like a static function I guess. nevermind

PR #529, opened 2016-02-03
Makes cullvisitor geometry function easier to hook

stephomi:


stephomi:
@jtorresfabra I added a getBoundingBox function for nodes.
Currently ``getBound()`` will still return the usual boundingSphere, but I renamed ``computeBound()`` into ``computeBoundingSphere()`` as it should mainly be for internal use only.

jtorresfabra:
@stephomi that's great, thanks for the info!

PR #530, opened 2016-02-05
[#DEV-6828] Bugfix/memleaks

Kuranes:
Removes GC pressure by reusing previously allocated objects
Vec3 at Each pop/push modelview
RenderStage is created at each camera with framebuffer cullvisitor pass.

### DIFF
--- a/sources/osg/RenderStage.js
+++ b/sources/osg/RenderStage.js
@@ -18,24 +18,40 @@ var Vec4 = require( 'osg/Vec4' );
  * like multistage pixel shading or impostors.
  */
 var RenderStage = function () {
+
     RenderBin.call( this );
-    this.positionedAttribute = [];
-    this.clearDepth = 1.0;
-    this.clearColor = Vec4.createAndSet( 0.0, 0.0, 0.0, 1.0 );
-    /*jshint bitwise: false */
-    this.clearMask = Camera.COLOR_BUFFER_BIT | Camera.DEPTH_BUFFER_BIT;
-    /*jshint bitwise: true */
-    this.camera = undefined;
-    this.viewport = undefined;
+    this.clearColor = Vec4.create();
     this.preRenderList = [];
     this.postRenderList = [];
-    this._renderStage = this;
+    // calling prototype to make sure
+    // we call renderstage and not renderbin init
+    RenderStage.prototype.init.call( this );
+
 };
 
 RenderStage.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( RenderBin.prototype, {
 
     // temporary, Utils.createPrototypeClass will solve this
     constructor: RenderStage,
+
+
+    init: function () {
+
+        RenderBin.prototype.init.call( this );
+        this.positionedAttribute.length = 0;
+        this.clearDepth = 1.0;
+        Vec4.set( 0.0, 0.0, 0.0, 1.0, this.clearColor );
+        /*jshint bitwise: false */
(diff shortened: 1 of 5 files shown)


sources/osg/RenderStage.js
```diff
@@ -37,6 +30,41 @@ RenderStage.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit(
     // temporary, Utils.createPrototypeClass will solve this
     constructor: RenderStage,
 
+    cloneType: function () {
+        return this.constructor.prototype.getOrCreate().init();
```
Kuranes: we can/could make this go way up directly to Object ?


```
Object.prototype = MACROUTILS.objectLibraryClass( 
...
...
cloneType: function () {
if (this.constructor.prototype.getOrCreate && this.constructor.prototype.init)
       return this.constructor.prototype.getOrCreate().init();
else
     return new this.constructor();
....
```

cedricpinson: I dont think the cache needs to be low level, you can just let the cloneType create new class and have a cache by type in the cull visitor because it's here it's used.
like a CullVisitor.getOrCreateRenderStage( type, args )

Kuranes: I'm ok with that, but that means changing the current cullvisitor::_rootRenderStage mechanism used by application providiing their own RenderStage ?

cedricpinson: It's already the case I think

Kuranes: https://github.com/cedricpinson/osgjs/blob/develop/sources/osg/CullVisitor.js#L351
^^ changing that in a CullVisitor.getOrCreateRenderStage( type, args ) 
means replaceing that cullvisitor::_rootRenderStage mechanism by a cullvisitor override mechanism.

cedricpinson: Yep data should be handled in CullVisitor

sources/osg/RenderStage.js
```diff
@@ -19,17 +19,10 @@ var Vec4 = require( 'osg/Vec4' );
...
-    /*jshint bitwise: false */
-    this.clearMask = Camera.COLOR_BUFFER_BIT | Camera.DEPTH_BUFFER_BIT;
-    /*jshint bitwise: true */
-    this.camera = undefined;
-    this.viewport = undefined;
+    this.clearColor = Vec4.create();
     this.preRenderList = [];
     this.postRenderList = [];
-    this._renderStage = this;
+    RenderStage.prototype.init.call( this );
```
cedricpinson: this.init() is more appropriate here

Kuranes: it would calls renderbin init

cedricpinson:
Could you validate it works well with all examples and check the number renderbin/renderstage behaviour. I think about the vr example that could trigger bug because of two rendering...

sources/osg/RenderStage.js
cedricpinson: Yep I saw after that it's tricky when you inherit. ok for me

Kuranes:
Tested, works well, even in VR.

On a model with 2 transparent material and 3 shadows:
- In VR it's 5 renderstage per frame, 19 renderbin ,  10 Vec4, 
- In Non VR, no postprocessing: 5 renderstage & 14 renderbins, 10 Vec4, 
- In Non VR,  full postprocessing: 19 renderstage & 14 renderbins, 40 Vec4,

PR #536, opened 2016-02-15
Fix billboard on matrix scaled nodes

jtorresfabra:


cedricpinson:
could you fix the name of the PR ? :)

jtorresfabra:
hehe it's ok now?

cedricpinson:
yep :) is there an example using it ?

jtorresfabra:
Yep, Texts use billboarding, also I've been using this in production to draw billboards. The problem is that it was not respecting  MatrixTransforms scale.
I noticed it when developing my next PR about AutoTransform node. As it was not respecting the scale when  autoscaling to screen. 

Also in that PR text supports characterSizeModes (http://trac.openscenegraph.org/documentation/OpenSceneGraphReferenceDocs/a00873.html#a80fb5ff01377d8ac78935653752fee82) and the example is updated. 

If you prefer I can do it all in the same PR, and close this one.

cedricpinson:
Yeah if we can only have one PR for this stuff that would be great

jtorresfabra:
oks, closing this one then.

PR #537, opened 2016-02-15
[#DEV-6887] Adds env Light

Kuranes:
Adds an new type of light that "removes" light frome a scene where
object is not lighted, allowing for envirronment shadowing


cedricpinson:
It feels bad to have this type of light in osgjs, not sure what the best way to do it, maybe setting the Light class customizable to add new type. But it does not feel right to add a new type of light only useful for sketchfab in the general Light class. We should talk about it this week

cedricpinson:
I feel it's not good to put this light type in regular Light class or in osgjs. Maybe Light should be customizable to add new type outside or in application specific. This light will be only useful for sketchab, not sure it should go into Light default type. We could talk about it this week

Kuranes:
When ? Sooner than later ?

PR #539, opened 2016-02-19
Add AutoTransform node and support fort characterSizeModes in osgText.Text

jtorresfabra:


sources/osg/CullVisitor.js
```diff
@@ -508,16 +509,9 @@ CullVisitor.prototype[ MatrixTransform.typeID ] = function ( node ) {
...
-    if ( node.getReferenceFrame() === TransformEnums.RELATIVE_RF ) {
-
-        var lastMatrixStack = this.getCurrentModelViewMatrix();
-        Matrix.mult( lastMatrixStack, node.getMatrix(), matrix );
-
-    } else {
-        // absolute
-        Matrix.copy( node.getMatrix(), matrix );
-    }
+    var lastMatrixStack = this.getCurrentModelViewMatrix();
```
jtorresfabra: Just changed this to support all Transform nodes, not only MatrixTransforms

sources/osg/Matrix.js
```diff
@@ -611,6 +649,22 @@ var Matrix = {
         };
     } )(),
 
+    preMultRotate: function ( matrix, quat ) {
+        if ( Quat.zeroRotation( quat ) )
+            return;
+        var r = Matrix.create();
```
cedricpinson: it's slow to create object like that, why not using regular function directly if it's slow ?

sources/osg/Matrix.js
```diff
@@ -611,6 +649,22 @@ var Matrix = {
...
+        Matrix.makeRotateFromQuat( r, quat );
+        Matrix.preMult( matrix, r );
+    },
+
+    postMultRotate: function ( matrix, quat ) {
+        if ( Quat.zeroRotation( quat ) )
```
cedricpinson: same here it's probably slower to do that than using simpler function

sources/osg/AutoTransform.js
```diff
@@ -0,0 +1,319 @@
...
+                Matrix.preMultTranslate( matrix, Vec3.neg( this._position, neg ) );
+                Matrix.postMultScale( matrix, scaleInverse );
+                Matrix.postMultTranslate( matrix, this._pivotPoint );
+            }
+            return true;
+        };
+    } )(),
+
+    computeBound: ( function () {
+        var matrix = Matrix.create();
```
cedricpinson: Could you check it works correctly with the bounding box PR merged recently ?

sources/osg/Matrix.js
jtorresfabra: Fixed

sources/osg/Matrix.js
jtorresfabra: Fixed

sources/osg/AutoTransform.js
jtorresfabra: It seems to work, as intersection also works.

jtorresfabra:
Added Billboardign computation for AutotransformNodes in js. Also added intersection tests in Text2D example.

cedricpinson:
could you rebase your PR on develop, also there are some commits 'merge' that could be removed

sources/osg/Matrix.js
```diff
@@ -65,6 +65,15 @@ var Matrix = {
         return matrix;
     },
 
+    equal: function ( a, b ) {
+        if ( a === b ) return true;
+        if ( a === undefined || b === undefined ) return false;
```
cedricpinson: better to use if ( !a ) instead of undefined. It will be valid for undefined and null

sources/osg/Matrix.js
cedricpinson: nop it does not seem fixed see https://github.com/cedricpinson/osgjs/blob/jtorresfabra/AutoTransform/sources/osg/Matrix.js#L640 to create the temp qtemp

sources/osg/Vec3.js
```diff
@@ -125,8 +125,12 @@ var Vec3 = {
         r[ 1 ] = a[ 1 ] + ( b[ 1 ] - a[ 1 ] ) * t;
         r[ 2 ] = a[ 2 ] + ( b[ 2 ] - a[ 2 ] ) * t;
         return r;
+    },
+    equal: function ( a, b ) {
+        if ( a === b ) return true;
+        if ( a === undefined || b === undefined ) return false;
```
cedricpinson: better with !a to handle undefined and null

stephomi: imho better to *not* do any check, why would it be undefined or even null???
The check should be done before the call and handled elsewhere.

sources/osg/Matrix.js
```diff
@@ -611,6 +649,22 @@ var Matrix = {
...
+        Matrix.makeRotateFromQuat( r, quat );
+        Matrix.preMult( matrix, r );
+    },
+
+    postMultRotate: function ( matrix, quat ) {
+        if ( quat[ 0 ] === 0.0 && quat[ 1 ] === 0.0 && quat[ 2 ] === 0.0 && quat[ 3 ] === 1.0 )
+            return;
+        var r = Matrix.create();
+        Matrix.makeRotateFromQuat( r, quat );
+        Matrix.postMult( matrix, r );
```
stephomi: The function should return matrix, to be linkable in chain.

sources/osg/Matrix.js
```diff
@@ -611,6 +649,22 @@ var Matrix = {
...
+            return;
+        var r = Matrix.create();
+        Matrix.makeRotateFromQuat( r, quat );
+        Matrix.preMult( matrix, r );
+    },
+
+    postMultRotate: function ( matrix, quat ) {
+        if ( quat[ 0 ] === 0.0 && quat[ 1 ] === 0.0 && quat[ 2 ] === 0.0 && quat[ 3 ] === 1.0 )
+            return;
+        var r = Matrix.create();
```
stephomi: r in closure to avoid dyna alloc.

sources/osg/Matrix.js
```diff
@@ -611,6 +649,22 @@ var Matrix = {
         };
     } )(),
 
+    preMultRotate: function ( matrix, quat ) {
+        if ( quat[ 0 ] === 0.0 && quat[ 1 ] === 0.0 && quat[ 2 ] === 0.0 && quat[ 3 ] === 1.0 )
+            return;
+        var r = Matrix.create();
+        Matrix.makeRotateFromQuat( r, quat );
+        Matrix.preMult( matrix, r );
```
stephomi: same as postMultRotate

sources/osg/Matrix.js
stephomi: same comment as below, check shouldn't be there

sources/osg/AutoTransform.js
stephomi: It can't work, there is no computeBound function, it's computeBoundingSphere instead.

sources/osg/AutoTransform.js
```diff
@@ -0,0 +1,310 @@
...
+        };
+
+    } )(),
+
+    computeWorldToLocalMatrix: ( function () {
+        var neg = Vec3.create();
+        var rotInverse = Quat.create();
+        var scaleInverse = Vec3.create();
+        return function ( matrix /*, nodeVisitor */ ) {
+            if ( this.scale[ 0 ] === 0.0 && this.scale[ 1 ] === 0.0 && this.scale[ 2 ] === 0.0 ) {
```
stephomi: this._scale

sources/osg/AutoTransform.js
jtorresfabra: The PR is so outdated now...

sources/osg/Vec3.js
jtorresfabra: agree, will remove it.

sources/osg/Matrix.js
jtorresfabra: ok

sources/osg/Matrix.js
jtorresfabra: ok

sources/osg/AutoTransform.js
jtorresfabra: good catch

stephomi:
There shouldn't be any "Merge" commit in this branch rebase should be done instead.
In your case a reset --hard develop + cherry-pick to fix it

jtorresfabra:
@stephomi first of all thanks for the review. I appreciate it.

> There shouldn't be any "Merge" commit in this branch, rebase from develop should be done instead.
> In your case a reset --hard develop + cherry-pick to fix it + push force :)

Yep I know, I am having problems to squash the merge commits, so it's easier to redo the PR.
I will close this one as soon as I have the new one prepared. Thanks.

stephomi:
Ok, good!

sources/osg/Matrix.js
jtorresfabra: @stephomi About this, I don't see how to do Matrix.create() in a closure. None of the methods in matrix does that. All of them create the matrix as I did, inside the function body. Do you mean to have only the var name in the closure like this?  https://github.com/cedricpinson/osgjs/blob/develop/sources/osg/Matrix.js#L1095

jtorresfabra:
@cedricpinson @stephomi I'm closing this issue as I opened a new one. About the closures in Matrix I'll wait for @stephomi's answer.

PR #540, opened 2016-02-23
Bugfix/better log compilation dev 6946

Kuranes:
Fixes develblock typo making shader compil always in debug mode
Fixes shader compilation log incorrectly accesing lines out of line number
Fixes acces to null PrecisionFormat


sources/osg/Shader.js
```diff
@@ -108,25 +111,31 @@ Shader.prototype = MACROUTILS.objectInherit( GLObject.prototype, {
             // get error line
             var line = parseInt( m[ 1 ] );
 
+            if ( line > linesLength ) continue;
             // webgl error report.
             Notify.error( 'ERROR ' + m[ 2 ] + ' in line ' + line, false, true );
 
+            var minLine = Math.max( 0, line - 7 );
```
cedricpinson: why - 7

PR #542, opened 2016-03-02
Improves the screen size normalization of the gizmo

stephomi:
Now the widget size doesn't depend of the width-height ratio and the canvas size.

Stops reading directly width and height attribute on canvas too (for
performance).

### DIFF
--- a/sources/osgAnimation/MorphGeometry.js
+++ b/sources/osgAnimation/MorphGeometry.js
@@ -2,6 +2,7 @@
 var MACROUTILS = require( 'osg/Utils' );
 var BufferArrayProxy = require( 'osg/BufferArrayProxy' );
 var Notify = require( 'osg/Notify' );
+var Vec3 = require( 'osg/Vec3' );
 var Geometry = require( 'osg/Geometry' );
 var StateSet = require( 'osg/StateSet' );
 var MorphAttribute = require( 'osgAnimation/MorphAttribute' );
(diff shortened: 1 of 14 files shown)


sources/osgAnimation/MorphGeometry.js
```diff
@@ -17,18 +18,27 @@ var StateAttribute = require( 'osg/StateAttribute' );
...
 
+// sync with UpdateMorph
+var EFFECTIVE_EPS = MorphGeometry.EFFECTIVE_EPS = 0.05;
+
 MorphGeometry.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( Geometry.prototype, {
 
-    init: function () {
+    init: function ( updateMorph ) {
+        // only the updateMorph contains all the target weights so we might need it for computeTransformedVertex
+        this._updateMorph = updateMorph;
```
cedricpinson: I dont see where it's used ?

stephomi:
I don't think anyone used the index triangle since it just couldn't work properly
- first the kdtree is re-ordering the triangles and uses its own indexes https://github.com/cedricpinson/osgjs/blob/develop/sources/osg/KdTree.js#L92
- second not sure how to handle multiple primitiveSet

sources/osgAnimation/MorphGeometry.js
stephomi: not useful anymore indeed

PR #545, opened 2016-03-08
[DEV-7032] Allows custom shadow effect shader

Kuranes:
In order to be able to make custom shadow computation the final shadow
effect computation is now a separate method that can be overridden.

### DIFF
--- a/sources/osgShader/Compiler.js
+++ b/sources/osgShader/Compiler.js
@@ -771,8 +771,19 @@ Compiler.prototype = {
             float: shadowedOutput
         } ).setShadowAttribute( shadow );
 
+        // allow overwrite by inheriting compiler
+        // where shadow inputs ( NDotL notably)
+        // can be used for non standard shadows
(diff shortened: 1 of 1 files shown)


sources/osgShader/Compiler.js
```diff
@@ -771,8 +771,19 @@ Compiler.prototype = {
...
         } ).setShadowAttribute( shadow );
 
+        // allow overwrite by inheriting compiler
+        // where shadow inputs ( NDotL notably)
+        // can be used for non standard shadows
+        return this.doShadowLight( lightedOutput, shadowedOutput, shadowInputs );
+
+    },
+
+    doShadowLight: function ( lightedOutput, shadowedOutput ) {
```
cedricpinson: the name of the function is a bit generic, why not connectShadowLightNode or something that tell more what is it if you have better. I am not telling my suggestion is the best

Kuranes: Didn't find satisfying name, "connectShadowLight" seems better to me too.

PR #546, opened 2016-03-09
Fix for wrong lighting when using textures

jtorresfabra:
According to RedBook computed color in lighting should be 

finalColor = primaryColor ( all lighting stuff using materials) * textureColor + secondaryColor.

That long standing bug made the textured objects appear more lighted than they should.
Easy to see if using texture + material with ambient component + ambient light.

### DIFF
--- a/sources/osgShader/Compiler.js
+++ b/sources/osgShader/Compiler.js
@@ -1774,48 +1774,8 @@ Compiler.prototype = {
 
         var materialUniforms = this.getOrCreateStateAttributeUniforms( this._material );
 
-
-        // diffuse color
-        var diffuseColor = this.getDiffuseColorFromTextures();
-
-        if ( diffuseColor === undefined ) {
-
-            diffuseColor = materialUniforms.diffuse;
-
-        } else {
-
-            this.getNode( 'InlineCode' ).code( '%color.rgb *= %diffuse.rgb;' ).inputs( {
-                diffuse: materialUniforms.diffuse
-            } ).outputs( {
-                color: diffuseColor
-            } );
-        }
-
         // vertex color needs to be computed to diffuse
-        diffuseColor = this.getVertexColor( diffuseColor );
-
-
-        // compute alpha
-        var alpha = this.createVariable( 'float' );
-
-        var textureTexel = this.getFirstValidTexture();
-
-        var alphaCompute;
-        if ( textureTexel ) // use alpha of the first valid texture if has texture
-            alphaCompute = '%alpha = %color.a * %texelAlpha.a;';
-        else
-            alphaCompute = '%alpha = %color.a;';
-
-        // Discard fragments totally transparents when rendering billboards
-        if ( this._isBillboard )
-            alphaCompute += 'if ( %alpha == 0.0) discard;';
-
-        this.getNode( 'InlineCode' ).code( alphaCompute ).inputs( {
-            color: materialUniforms.diffuse,
-            texelAlpha: textureTexel
-        } ).outputs( {
-            alpha: alpha
-        } );
+        var diffuseColor = this.getVertexColor( materialUniforms.diffuse );
 
         // 2 codes path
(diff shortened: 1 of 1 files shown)


jtorresfabra:
I also reordered the code to make more sense when reading it. Maybe @Kuranes and @stephomi could take a look also.

Kuranes:
From what I understand, before was
```js
if (textureDiffuse) diffuse = materialDiffuse*textureDiffuse
else diffuse = materialDiffuse
if (vertexDiffuse) diffuse = diffuse *vertexDiffuse

finalColor = lighting(diffuse)

```
after PR it is
```js
if (vertexDiffuse) diffuse = materialDiffuse*vertexDiffuse
else diffuse = materialDiffuse

finalColor = lighting(diffuse)

if (textureDiffuse) finalColor = finalColor * textureDiffuse 
```

meaning texture diffuse color is not a diffuse texture (scaled by light equation NdotL) but more like an Ambient occlusion (light independant) ?

( for me a secondary color is a second vertex color or a second texture color ? )

Maybe there is a possiblity to do "secondary color", but only if we have a vertex color and a texture color, we could check /consider the texture color a secondary (or the vertex color as the secondary) but that's not what current code does ?
(as it always interpret texture color as a secondary, even if there is no primary ?)

Kuranes:
about reordering, perhaps we could move the whole 'alpha' part prior to lighting ?
would make sense to premult before the NDotL as it scale diffuse... may need to read on that.

(at least the alpha, but not premult, as in case of 'discard' and 'mulitple light', it would save a lot of perf)

jtorresfabra:
@Kuranes thanks for the review.

The point of the PR is not the secondary color, as you say it should come with an alternate vertex color. 
The point is the texture should not be affected by lighting. 

> meaning texture diffuse color is not a diffuse texture (scaled by light equation NdotL) but more like an Ambient occlusion (light independant) ?

That's the point, it is like that in OpenGL. Texture Color is combined with primary color ( being primary color materials + color + ligthing ). I can search some links, but it is in the RedBook.

jtorresfabra:
Maybe that shed some light: OpenGL  https://books.google.es/books?id=ckpNmNlLE_sC&pg=PA168&lpg=PA168&dq=fragment+lighting+texture+secondary&source=bl&ots=KGM9tDb0Lx&sig=qiDROsq8Hv8vhG_623bsqhmsIac&hl=es&sa=X&ved=0ahUKEwjXi4PF04jMAhWC2RoKHQcVD8UQ6AEIHzAA#v=onepage&q=fragment%20lighting%20texture%20secondary&f=false

OpenGL computes lighting before TextureMapping, in its default mode.

Kuranes:
Thanks for the reference, much clearer. 
Seems indeed to be Opengl 1.4 behaviour !

jtorresfabra:
Yep, also the seconday is used to apply specular colo, because it could cause some problems. ( See redbook page 450: Secondary Specular Color when lighting is enabled )

Kuranes:
Do you happen to have more on the lighting part,  to be sure to get opengl canonic lighting equation ?
( to double check the createLightiing shader code, particularly light ambient )

jtorresfabra:
That's all I have at the moment, with this PR I just see my scenes as in OSG. But I also noticed that all the ambient variables are not used in the shader functions, so maybe them should be removed from the function parameters. ( materialAmbient and lightAmbient not used at all in https://github.com/cedricpinson/osgjs/blob/develop/sources/osgShader/node/lights.gls)

jtorresfabra:
One more thing, I recall now., by not using materialAmbient and lightAmbient in the lighting functions then it seems attenuation does not affect ambient stuff which maybe wrong. But I don't have a PR for this.

Kuranes:
Thanks for the infos!

cedricpinson:
good for me, could you rebase on develop plz ?

jtorresfabra:
Done.

PR #547, opened 2016-03-10
Added reference frame to LightSource

jtorresfabra:


sources/osg/LightSource.js
```diff
@@ -9,6 +11,7 @@ var Node = require( 'osg/Node' );
 var LightSource = function () {
     Node.call( this );
     this._light = undefined;
+    this.referenceFrame = TransformEnums.RELATIVE_RF;
```
cedricpinson: _referenceFrame instead

jtorresfabra:
Done.

sources/osg/LightSource.js
```diff
@@ -18,7 +21,25 @@ LightSource.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit(
...
+    },
+    getReferenceFrame: function () {
+        return this._referenceFrame;
+    },
+    computeBound: function ( bsphere ) {
+        Node.prototype.computeBound.call( this, bsphere );
+        if ( this._light !== undefined && this._referenceFrame === TransformEnums.RELATIVE_RF ) {
+            var position = this._light.getPosition();
+            if ( position[ 3 ] === 0.0 ) {
+                var div = 1.0 / position[ 3 ];
```
Kuranes: isn't it a divide By Zero there?
I'm sorry, I  don't get what that code part should do ?
maybe Light does have impact on bounding box to prevent frustum culling, but why ligth position is "scaled"?

jtorresfabra: Ouch, it  should be !==0.

jtorresfabra: Fixed

sources/osg/LightSource.js
```diff
@@ -18,7 +21,25 @@ LightSource.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit(
...
+    getReferenceFrame: function () {
+        return this._referenceFrame;
+    },
+    computeBound: function ( bsphere ) {
+        Node.prototype.computeBound.call( this, bsphere );
+        if ( this._light !== undefined && this._referenceFrame === TransformEnums.RELATIVE_RF ) {
+            var position = this._light.getPosition();
+            if ( position[ 3 ] !== 0.0 ) {
+                var div = 1.0 / position[ 3 ];
+                bsphere.expandByVec3( Vec3.createAndSet( position[ 0 ] * div, position[ 1 ] * div, position[ 2 ] * div ) );
```
Kuranes: sorry, I still have hard time  understanding :( but surely related to my lack of knowlege of OSG... 

position[ 3 ]  for me is just 0 or 1, "1" being for non-directionnal light 
What can it be used for too ? 
(other value would breaks the light::isSpotlight light::isPointLight etc)

( trying hard to understand as that code does introduce allocation at each compute with the div thing)

jtorresfabra: Hey @Kuranes, 

Just picked the code from OSG https://github.com/openscenegraph/OpenSceneGraph/blob/master/src/osg/LightSource.cpp#L61. 

What I understand is: if the w homogeneus coordinate(p[3]) is zero the light is at infinity, if its 1 then the light is in the world space, so to get the real coordinates (x,y,z ) all of them must be divided by w. 

Then we calculate the bounding sphere using this values, just to know when the lightSource should be traversed.

Kuranes: ok, makes a bit more sense, So that means we should be able to tranform light postion.
perhaps we miss something that is in osg that does allow that ?
( as unless you transform light position, w will always be 1.0 )

jtorresfabra:
Indeed Light position could be transformed, also Light has a setPosition method using a Vec4. So you can set any w you want.

sources/osg/LightSource.js
jtorresfabra: Light position could be transformed, also Light class has a setPosition method with a vec4 as a parameter, so you could just have any w you want.

Kuranes:
I'm afraid osgjs doesn't support that, as current Light.js code uses "w" as light type... so that transforming position would break the code that relies on that ( positionned attribute / shader compiler)

jtorresfabra:
Hummm can you point me to the code? Also for me is not a big deal as my lights are not transformed, but I use reference frame. So not overriding the computeBound does not affect me. Anyway it should be stated as a comment at least if this PR moves forward. 

Thoughts?

Kuranes:
Sorry, seems I'm wrong, test is 'if position[3 ]!== 0' so that should be ok when w !== 1

Only caveat seem to be setting lightType after setting the position as it's overriding position values
( see https://github.com/jtorresfabra/osgjs/blob/Adds_RFLightSource/sources/osg/Light.js#L264 )

jtorresfabra:
I see, so you always use a MatrixTransform for positioning the lights when coding js.
For models coming from OSG, position also could be defined, in fact the wrapper takes that into account. So adding the reference frame stuff is needed ( at least for my use case... hehe )

Kuranes:
ok, makes sense the me , now that I underestand more the in and out
PR does it fixes osgjs to be more OSG, always good :)

cedricpinson:
@Kuranes needs to check how to handle scene bounding box in sketchfab

Kuranes:
Tested & Approved

jtorresfabra:
Related ( https://github.com/cedricpinson/osg/pull/64 ) to be able to export from OSG scenes...

PR #548, opened 2016-03-16
Bug/multiple output nodes dev 7081

Kuranes:
- Light Shader nodes outputs: color but also ndl, lighted.
When a shader node has mutliple output, and some are used by other node, they should be part of shader node graph traversal, otherwise dependency traversal is wrong.
fix osgjs traversal to handle that.
fix node with mutiple output so that they declare them explicitely.

- Adds Light to connecShadowLightMode

### DIFF
--- a/sources/osgShader/Compiler.js
+++ b/sources/osgShader/Compiler.js
@@ -774,11 +774,11 @@ Compiler.prototype = {
         // allow overwrite by inheriting compiler
         // where shadow inputs ( NDotL notably)
         // can be used for non standard shadows
-        return this.connectShadowLightNode( lightedOutput, shadowedOutput, shadowInputs );
+        return this.connectShadowLightNode( light, lightedOutput, shadowedOutput, shadowInputs );
 
     },
 
-    connectShadowLightNode: function ( lightedOutput, shadowedOutput ) {
+    connectShadowLightNode: function ( light, lightedOutput, shadowedOutput ) {
 
         var lightAndShadowTempOutput = this.createVariable( 'vec3', 'lightAndShadowTempOutput' );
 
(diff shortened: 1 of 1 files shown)


sources/osgShader/Compiler.js
```diff
@@ -877,7 +888,11 @@ Compiler.prototype = {
                 inputs.eyeVector = this.getOrCreateNormalizedPosition();
 
             this.getNode( nodeName ).inputs( inputs ).outputs( {
-                color: lightedOutput
+                color: lightedOutput,
+                lightEyePos: inputs.lightEyepos, // spot and point only
```
cedricpinson: it should be inputs.lightEyePos instead of inputs.lightEyepos no ? same for eyedir

sources/osgShader/Compiler.js
```diff
@@ -1058,6 +1078,8 @@ Compiler.prototype = {
 
         func._map = {};
```
cedricpinson: There is multiple time reset field like:
func._text = [];
func._mapTraverse = [];
this._traversedNodeList = {};

is it possible to write a function that reset it, I feel like it will grow and will be error prone

sources/osgShader/Compiler.js
```diff
@@ -947,10 +964,23 @@ Compiler.prototype = {
...
 
-        // keep trace we visited.
+        // keep trace we visited
         this.markNodeAsVisited( node );
+
     },
 
+    // clean necessary bits before traversing
+    // called in each evaluate func belows
+    preTraverse: function ( visitor ) {
```
cedricpinson: maybe initTraverse is a better name ?

Kuranes: I started with that, but was afraid "init" was implying a "once and for all"
Whereas "Pre" goes well with compiler language (multi-pass, pre-pass, post-pass) ?

cedricpinson: right it's ok, I will merge like that

PR #552, opened 2016-03-23
Fix bbox

stephomi:


stephomi:
@jtorresfabra you might want to take a look (change in getHomeBound function)

jtorresfabra:
@stephomi  will do, I'll checkout with the CADManipulator.

jtorresfabra:
@stephomi not sure about "flag" name parameter, is a little bit misleading. We could probably find out a better name for the enum and use it in the function call as the name of the parameter (like boundingType or whatever). What do you think?

stephomi:
ok for boundType

stephomi:
In any case getHomeBound returns a boundingSphere 
so, I went for boundStrategy  :).

jtorresfabra:
:+1: Much better.

jtorresfabra:
Also boundPolicy could work, but for me is enough :)

cedricpinson:
let's go for boundType

cedricpinson:
boundStrategy is ok for me

PR #554, opened 2016-03-28
Adds Vertex Array Object Support

Kuranes:
~~WebGL2 makes it mandatory to support vertex Array object.~~

It's not true


Kuranes:
Todo: 
- fix webglcaps strange test fail (webpack include order fail)
- double check "morph cpu" path

cedricpinson:
https://github.com/cedricpinson/osgjs/pull/554#issue-144004134
is it not true

sources/osgViewer/webgl-utils.js
```diff
@@ -182,8 +182,7 @@ var WebGLUtils = function () {
  */
 if ( !window.requestAnimationFrame ) {
     window.requestAnimationFrame = ( function () {
-        return window.requestAnimationFrame ||
-            window.webkitRequestAnimationFrame ||
+        return window.webkitRequestAnimationFrame ||
```
cedricpinson: not related add this into another commit

sources/osg/WebGLCaps.js
```diff
@@ -103,6 +103,8 @@ WebGLCaps.prototype = {
 
         this._isGL2 = typeof window.WebGL2RenderingContext !== 'undefined' && gl instanceof window.WebGL2RenderingContext;
 
+        Texture = require( 'osg/Texture' );
```
cedricpinson: not related add this into another commit

sources/osg/DrawElements.js
```diff
@@ -39,6 +39,12 @@ DrawElements.prototype = {
         var gl = state.getGraphicContext();
         gl.drawElements( this.mode, this.count, this.uType, this.offset );
     },
+    drawElements: function ( state, glParam ) {
```
cedricpinson: Let's keep index buffer for now, so this function can be removed

sources/osg/Geometry.js
```diff
@@ -59,24 +72,197 @@ Geometry.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( No
...
+                p.draw( state, gl );
+            }
+        }
+
+        //this._extVAO.bindVertexArrayOES( null );
+        state._currentVAO = vao;
+    },
+
+    drawImplementationVAO: function ( state ) {
+
```
cedricpinson:
```
drawImplementationVAO(state) {

    if ( needUpdate ) {
        state.clearVertexAttribCache()
        bindVertexArrayObject()
        this.drawImplementation() 
    } else {
        bindVertexArrayObject()
        var primitives = this.primitives
        for ( var j = 0, m = primitives.length; j < m; ++j ) {
             primitives[' + j + '].draw(state);
    } 
    unbindVertexArrayObject()
}
```

Kuranes: Current DrawVAO is ok, the version proposed doesn't add anything, and in case of "update" makes VAO recreation +  All VertexBuffer reupload at each frame (instead of just uploading modified buffers)

Kuranes:
```
drawImplementationVAO(state) {

    if (!vao) {
        state.clearVertexAttribCache()
        bindVertexArrayObject()
        this.drawImplementation() 
    } 

  drawVAO(vao)
}
```

sources/osg/DrawElements.js
Kuranes: it's git diff gone wild. 
That func drawElements not added  added by this push 
(perhaps added glParam to avoid state.getGraphicContext() each call but nothing else)

sources/osg/WebGLCaps.js
Kuranes: was related but maybe not needed anymore.
getting VAO extension existance at Geometry.js parse time was causing circular dependencies
will check if still needed and/or remove when cleaning PR

sources/osgViewer/webgl-utils.js
Kuranes: woops

PR #556, opened 2016-03-31
A few necessary changes for our use case

batiste:
- We needed to be able to rewrite the name of the file on the fly
- computeURL was adding the automatically created prefix to an already complete URL we passed using the options.readImageURL
- We img.crossOrigin = "Anonymous"; as we load cross origin images

### DIFF
--- a/sources/osgDB/Input.js
+++ b/sources/osgDB/Input.js
@@ -108,7 +108,7 @@ Input.prototype = {
         if ( typeof this._defaultOptions.prefixURL === 'string' &&
             this._defaultOptions.prefixURL.length > 0 ) {
 
-            if(url.indexOf(this._defaultOptions.prefixURL) === 0) {
+            if ( url.indexOf( this._defaultOptions.prefixURL ) === 0 ) {
                 return url;
             }
 
@@ -175,7 +175,6 @@ Input.prototype = {
         // crossOrigin does not work for inline data image
         var isInlineImage = ( url.substring( 0, checkInlineImage.length ) === checkInlineImage );
         var img = new window.Image();
-        img.crossOrigin = "Anonymous";
         img.onerror = function () {
             Notify.warn( 'warning use white texture as fallback instead of ' + url );
             image.setImage( Input.imageFallback );
@@ -415,8 +414,8 @@ Input.prototype = {
 
         var url = vb.File;
         var defer = P.defer();
-        if(options.rewriteFileUrl) {
-            url = options.rewriteFileUrl(url)
+        if ( options.rewriteBinaryArrayURL ) {
+            url = options.rewriteBinaryArrayURL( url );
         }
 
         this.readBinaryArrayURL( url ).then( function ( array ) {
--- a/sources/osgDB/Input.js
+++ b/sources/osgDB/Input.js
@@ -108,6 +108,10 @@ Input.prototype = {
         if ( typeof this._defaultOptions.prefixURL === 'string' &&
             this._defaultOptions.prefixURL.length > 0 ) {
 
+            if(url.indexOf(this._defaultOptions.prefixURL) === 0) {
+                return url;
+            }
+
             return this._defaultOptions.prefixURL + url;
         }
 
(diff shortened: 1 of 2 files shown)


jtorresfabra:
Hi @batiste, 

Travis tests are failing, ir seems you need to beautify your code first ( $grunt beautify ). It's better if you get a clean compilation ( as doing $grunt && grunt test ) before doing a PR. 

Also I don't fully understand what are you trying to achieve:

Why img.crossOrigin = "Anonymous" should be a default as it is done if you pass the option a few lines down ( https://github.com/cedricpinson/osgjs/blob/develop/sources/osgDB/Input.js#L183 )?

Also if you are adding a new rewriteURL -which I am not against- function you should declare it in osgDB.Options, so the rest of users know it could be hooked.

https://github.com/cedricpinson/osgjs/blob/develop/sources/osgDB/Options.js

batiste:
I have done the necessary changes!

jtorresfabra:
Hi @batiste 

I'm not the one who merge the commits here, anyway a couple more of comments:

> computeURL was adding the automatically created prefix to an already complete URL we passed using the options.readImageURL

I don't fully understand this, if you are hooking the readImageURL call, then the compute URL should not be being called at all: 
https://github.com/cedricpinson/osgjs/blob/develop/sources/osgDB/Input.js#L215

I mean computeURL is done after the call to the hooked function: https://github.com/cedricpinson/osgjs/blob/develop/sources/osgDB/Input.js#L220

Also what it seems to me is that you only need to hook 
options.initializeBufferArray
https://github.com/cedricpinson/osgjs/blob/develop/sources/osgDB/Input.js#L409

So I think you only need to pass options to initializeBufferArray as you did (https://github.com/batiste/osgjs/blob/develop/sources/osgDB/Input.js#L413) and then hook that function to do whatever you want. 

BTW initializeBufferArray should'nt be added to Options? ( @cedricpinson ) : 
https://github.com/cedricpinson/osgjs/blob/develop/sources/osgDB/Options.js

batiste:
@jtorresfabra the thing is that I am calling readImageURL back like it is described in the comment. you might be right on the first point and maybe I can make it work without recalling the old method (this delete option thing is crazy by the way), I will give it another try.

For the second point how can I hook initializeBufferArray and still reuse this part of the code? I don't see how I can do that: https://github.com/cedricpinson/osgjs/blob/develop/sources/osgDB/Input.js#L413

batiste:
Tried this approach for readImageURL, here is my working function:

    function readImageURL(url, options) {
      delete options['readImageURL'];
      return this.readImageURL(SOMETHING.osgjs.texture, options);
    }

SOMETHING.osgjs.texture is an absolute path, and without my modification the path is added again. Now if I try to just return the absolute URL

    function readImageURL(url, options) {
      return SOMETHING.osgjs.texture
    }

I get an error of type input.readImageURL(...).then is not a function. This is because readImageURL return a Promise, not an URL... This is bad is you ask me.

Last try I try to rewrite the promise part:

    function readImageURL(url, options) {
        var image = new Image();
        var defer = P.defer();
        this.fetchImage( image, SOMETHING.osgjs.texture, options, defer );
        return defer.promise;
    }

Well I am surprise this code doesn't just blow up on the P.defer() as I didn't know I had the Promise lib loaded in my scope. Anyway it now fails with "image.setURL is not a function"

Needless to say that monkey patching a library I know nothing about just to configure an URL is neither easy nor fun.

jtorresfabra:
Sorry @batiste I didn't had the chance to answer:

> function readImageURL(url, options) {
>     var image = new Image();
>     var defer = P.defer();
>     this.fetchImage( image, SOMETHING.osgjs.texture, options, defer );
>     return defer.promise;
> }

> Well I am surprise this code doesn't just blow up on the P.defer() as I didn't know I had the Promise lib loaded in my scope. Anyway it now fails with "image.setURL is not a function"

You were so close! try this one, just tested and it works for me. 

var image = new osg.Image();

You need to create an osg.Image instead of an image. In the Input class it works because there is a declaration at the begining of the file ( var Image = require( 'osg/Image' ); )

Also think that readImageURL return an "actual" osg.Image, not an URL. Maybe it should be readImageFromURL to make things clearer.

> Needless to say that monkey patching a library I know nothing about just to configure an URL is neither easy nor fun.

Yep, totally agree, maybe asking in the mailing list before patching the code would have been less disappointing. 
Hope this helps.

cedricpinson:
Hi,
Yeah the need to delete something in options sucks, we should probably split the function to avoid to do that. Also I am not sure what you want to achieve exactly, could you provide a minimal test case ?
The computeURL is executed after the call to the hook function, not sure it makes sense actually.

cedricpinson:
no news, I close it for now

PR #557, opened 2016-04-01
Added the possibility to have attributes declaring uniforms conditionaly

jtorresfabra:


cedricpinson:
careful because it means that if the uniforms can exist or not for an attributes, it also mean that you should have a getHash that contains the list of uniforms. It also means that you can create different shader depending on the list of your uniform.

jtorresfabra:
> careful because it means that if the uniforms can exist or not for an attributes

Yes, I'm aware of this.

> it also mean that you should have a getHash that contains the list of uniforms

Instead of that I declare my uniforms based in a class variable ( GeometryAttr._geometryType for example ), and I add that clas variable to the Hash.

> It also means that you can create different shader depending on the list of your uniform.

Yep, and that's the point, it adds a bit more of flexibility, and I think ( I could be wrong ) it doesn't hurt the current behaviour. 

Thoughts?

cedricpinson:
good for me

PR #558, opened 2016-04-06
Adds TimerGPU

Kuranes:
Adds a Timer GPU using webgl extenstion https://www.khronos.org/registry/webgl/extensions/EXT_disjoint_timer_query/

Have to call it before and after a set of gl calls to get a timer.
```javascript
TimerGPU.instance().start('mypostproc');
glcalls...
TimerGPU.instance().end('mypostproc', function(averageTime){ console.log(averageTime)});
```

### DIFF
--- a/sources/osg/TimerGPU.js
+++ b/sources/osg/TimerGPU.js
@@ -1,0 +1,245 @@
+'use strict';
+
+// use EXT_disjoint_timer_query
+// to time webgl calls GPU side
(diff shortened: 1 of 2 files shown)


Kuranes:
Problem is mainly how to use it.
For instance timing a camera draw would need a camera::BeforeDraw callback (like the FinalDraw callback), etc

PR #560, opened 2016-04-12
[#DEV-7321] Adds TimerGPU

Kuranes:
Adds a Timer GPU using webgl extenstion https://www.khronos.org/registry/webgl/extensions/EXT_disjoint_timer_query/

Have to call it before and after a set of gl calls to get a timer.
```js
TimerGPU.instance().start('mypostproc');
glcalls...
TimerGPU.instance().end('mypostproc');
```

### DIFF
--- a/sources/osg/TimerGPU.js
+++ b/sources/osg/TimerGPU.js
@@ -1,0 +1,299 @@
+'use strict';
+var Notify = require( 'osg/Notify' );
+
+// use EXT_disjoint_timer_query
+// to time webgl calls GPU side
+// average over multiple frames
+// for consistent results
+// use double buffer queries for that.
+// see http://www.reedbeta.com/blog/2011/10/12/gpu-profiling-101/
+var TimerGPU = function ( gl ) {
+
+    this._enabled = false;
+
+    if ( gl ) {
+
+        var ext = gl.getExtension( 'EXT_disjoint_timer_query' );
+        if ( !ext ) return this;
+        // https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/conformance/extensions/ext-disjoint-timer-query.html#L102
+        this._hasTimeElapsed = ext.getQueryEXT( ext.TIME_ELAPSED_EXT, ext.QUERY_COUNTER_BITS_EXT ) >= 30;
+        this._hasTimeStamp = ext.getQueryEXT( ext.TIMESTAMP_EXT, ext.QUERY_COUNTER_BITS_EXT ) >= 30;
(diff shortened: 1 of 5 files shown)


sources/osg/TimerGPU.js
```diff
@@ -0,0 +1,273 @@
...
+
+        if ( !this._hasTimeElapsed && !this.hasTimeStamp ) {
+            return;
+        }
+
+        // no timestamp means not start/end absolute time
+        // which means each start must be followed by a end
+        // BEFORE any other start (of other queryID)
+        if ( !this.hasTimeStamp ) {
+            console.log( 'Warning: do not use interleaved GPU query' );
```
cedricpinson: osg.warn

sources/osg/TimerGPU.js
```diff
@@ -0,0 +1,273 @@
...
+    // query waiting async results from GPU 
+    this._waitingQueries = {};
+};
+
+TimerGPU.instance = function ( gl ) {
+
+    if ( !TimerGPU._instance ) {
+        TimerGPU._instance = new TimerGPU( gl );
+    } else if ( gl && TimerGPU._instance.getContext() !== gl ) {
+        TimerGPU._instance.setContext( gl );
```
cedricpinson: why ?

cedricpinson:
seems good, Could you add this into a osgjs sample ? like shadow maybe ?

sources/osg/TimerGPU.js
Kuranes: k

sources/osg/TimerGPU.js
Kuranes: In case first call to timerGPU.instance() is without the gl parameter. 
Cannot remember how that did happen.
Will remove.

Kuranes:
Miss a way to use those in GL code.
Either do some renderstage override ?
OR my preferred solution add a osg:Camera:beforeFirstDrawCall() callback mechanism where we could put the timerGPU::start
(we already have the osg:Camera:FinalDrawCall() https://github.com/cedricpinson/osgjs/blob/develop/sources/osg/Camera.js#L74 )

cedricpinson:
Check in osg the different callback and how they are called https://github.com/cedricpinson/osg/blob/master/include/osg/Camera#L611

cedricpinson:
Seems like to leak, it would be more appropriate to have a number of frame and using a modulo to reuse same entry in map/array

Kuranes:
- fixed leaks
- rebased

PR #561, opened 2016-04-12
[#DEV-7341] Handles Better Webgl Context Lost

Kuranes:
Once context Lost, prevent any further render, inform osgjs client code with a callback
Otherwise client may continue to ask new renders and thus getting
endless errors.

### DIFF
--- a/sources/osgViewer/Viewer.js
+++ b/sources/osgViewer/Viewer.js
@@ -115,6 +115,8 @@ var Viewer = function ( canvas, userOptions, error ) {
 
     this._hmd = null;
     this._useVR = false; // if we use custom requestAnimationFrame
+
+    this._contextLost = false;
 };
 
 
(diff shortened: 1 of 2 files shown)


Kuranes:
add unit test using webgl debug js context lost simulator

sources/osgViewer/Viewer.js
```diff
@@ -193,8 +193,15 @@ Viewer.prototype = MACROUTILS.objectInherit( View.prototype, {
...
 
+    setContextLostCallback: function ( cb ) {
+        this._contextLostCallback = cb;
+    },
     contextLost: function () {
         Notify.log( 'webgl context lost' );
+        if ( this._contextLostCallback ) {
+            this._contextLostCallback();
+        }
+        this._contextLost = true;
```
cedricpinson: this._contextLost is initialized somewhere ?

Kuranes: no, undefined by defaut.

cedricpinson: It's better to initialize vars in the constructor, it's easier to know member variables used in the class. The few exceptions is when we need to call a init method that duplicate a lot of code initialized.

PR #566, opened 2016-04-20
Feature/vao support dev 7320

Kuranes:
Adds support for Vertex Array object (on gl bind draw call for all vertex buffer binding and type)


Kuranes:
- squashed commits for cleaner merge

sources/osg/DrawElements.js
```diff
@@ -39,6 +39,12 @@ DrawElements.prototype = {
         var gl = state.getGraphicContext();
         gl.drawElements( this.mode, this.count, this.uType, this.offset );
     },
+    drawElements: function ( state, glParam ) {
```
cedricpinson: This part is not used we can remove the code

Kuranes:
- removed no more necessary code

cedricpinson:
refactored in another PR https://github.com/cedricpinson/osgjs/pull/570

PR #567, opened 2016-04-21
Fix for uniforms declared as arrays

jtorresfabra:
Uniforms declared like this:
`uniform float myarray[10];` or 
`uniform float myarray [10]; //note blank space before [ ` 
were not working.


Kuranes:
:+1:

PR #569, opened 2016-04-22
Fixes uniform array regexp

Kuranes:
Handles now all cases

        uniform float        ubones [6];
        uniform float        ubones[6];
        uniform float        ubones;

### DIFF
--- a/sources/osg/Program.js
+++ b/sources/osg/Program.js
@@ -216,12 +216,13 @@ Program.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( GLO
     },
 
     cacheUniformList: function ( gl, str ) {
+
         var r = str.match( /uniform\s+\w+\s+\w+((\s)?\[(.*?)\])?/g );
         var map = this._uniformsCache;
         if ( r !== null ) {
             for ( var i = 0, l = r.length; i < l; i++ ) {
                 var uniform = r[ i ].match( /uniform\s+\w+\s+(\w+)/ )[ 1 ];
-                var uniformName = r[ i ].match( /uniform\s+\w+\s+(\w+((\s)?\[(.*?)\])?)/ )[ 1 ];
+                var uniformName = r[ i ].match( /uniform\s+\w+\s+(\w+)(\s?\[.*?\])?/ )[ 1 ];
                 var location = gl.getUniformLocation( this._program, uniform );
                 if ( location !== undefined && location !== null ) {
                     if ( map[ uniformName ] === undefined ) {
@@ -231,6 +232,7 @@ Program.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( GLO
                 }
             }
         }
+
     },
 
     cacheAttributeList: function ( gl, str ) {


Kuranes:
@jtorresfabra animation sample wasn't working anymore, fixed the regexp, and tested against 3 cases.

jtorresfabra:
Ummm weird, it workder for me in my linux box, maybe a linux/windows issue?

Kuranes:
validated regexp in an online tool 
(confirmed previous wasn't working and this one does)

jtorresfabra:
ok, let me check ( I believe you, but it is weird that it works for me if it is wrong)

jtorresfabra:
@Kuranes also that's pretty weird that if your version works for you then https://github.com/cedricpinson/osgjs/blob/bugfix/uniform-array-regexp/sources/osg/Program.js#L220 also fail as well.

jtorresfabra:
Also for me OLD expression is working Ok matching :
image

and new is not :

image

So I'm starting to think is a linux/windows issue

Kuranes:
I'm perhaps misunderstanding (or just me doing friday errors) ?

I think we need to get 'ubones' always disregardless of uniforme being an array or not ?

jtorresfabra:
Ok maybe there is a missunderstanding.
We need to get  'ubones [6];' As  https://github.com/cedricpinson/osgjs/pull/569/files#diff-0b7b4152a92b99864e6bba1c1299e821R226 needs only ubones; 
but we need to map with 'ubones [6];' here. https://github.com/cedricpinson/osgjs/blob/bugfix/uniform-array-regexp/sources/osg/Program.js#L229

Maybe changing the var name ( uniform per uniformName makes sense). Anyway I don't see how it could affect to animation example in any way.

Kuranes:
Here double tested linux with develop animation doesn't play

Our animation code (and sketchfab...) does expect uniform name without array notation.
for example it expects uBones and update uBones, never uBones[6].
https://github.com/cedricpinson/osgjs/blob/develop/sources/osgAnimation/SkinningAttribute.js#L48

imho it ismuch more flexible to use without [n] ? (independant of array size, like bones that are mesh dependant, so no need to make complex uniform naming in compilers)
it's also the way "uniform location" works, so it maps 1-1 ?

Do you have a use case shader / shadernode that expects an "array[n]" as uniform name instead of "array" and it's necessary to do so ?

jtorresfabra:
Ummm you are right animation is not working I did a bad merge ( sorry about it ).

> Our animation code (and sketchfab...) does expect uniform name without array notation.
> for example it expects uBones and update uBones, never uBones[6].

In fact the size of the array in animation example seems to vary, that could be the reason why it does not work with a "formal" definition with 6 elements. So in fact you don't want a 'fixed size' array in this case.

> imho it ismuch more flexible to use without [n] ? (independant of array size, like bones that are mesh dependant, so no need to make complex uniform naming in compilers)
> it's also the way "uniform location" works, so it maps 1-1 ?

Yes, now that I read the GLES shading languaje specification I see your point. Indeed there are differences between this and GLSL( 1.5 ) when using arrays. I was used to declared arrays in shaders in this way. In my defense... GLSL 1.5 is much more strict with declaring arrays size than GLES seems to be, so probably is my fault.

> Do you have a use case shader / shadernode that expects an "array[n]" as uniform name instead of "array" and it's necessary to do so ?

Some times it is more readable and intuitive if you have a fixed size array. There are a lot of reasons to declare a fixed size array.
Also from spec (https://www.khronos.org/registry/gles/specs/2.0/GLSL_ES_Specification_1.0.17.pdf): 
Arrays declared as formal parameters in a function declaration must specify a size.

I understand not declaring the size could be valid and flexible, but I don't like the idea of not letting the user to declare fixed size arrays, because it is something a lot of people do, and it is something the language allows to do. 

Say that, my solution seems to not be valid for the animation case.

Thoughts?

Kuranes:
I see your point. I think the func has to be similar to getActiveUnbiform location.
( see https://github.com/Kuranes/osgjs/blob/feature/shadows/js/osg/Program.js#L94 )

That said I have no problem against adding an alias on each array uniform, so that for instance uBones could be accessed by uniformMap['uBones'] or uniformMap['uBones[6]']

Some form of aliasing might be needed at some point anyway (for uniform packing like having 4 separate float value in a single float4 uniform, but 4 alias to them from js user code)

jtorresfabra:
@Kuranes, so I think I could misunderstood how to create uniform arrays, actually I think it is the case. And probably we did'nt need any change at all. Let me a while to study SkinningAttribute. 
@cedricpinson In the mean time you can apply @Kuranes patch or revert my commit.

Sorry for all this confusion, my bad.

PR #575, opened 2016-05-05
Added Autotransform Node

jtorresfabra:


sources/osgShader/node/billboard.glsl
```diff
@@ -0,0 +1,16 @@
...
+      vec3 scale;
+      scale.x = xs * sqrt( matrix[0][0] * matrix[0][0] + matrix[0][1] * matrix[0][1] + matrix[0][2] * matrix[0][2]);
+      scale.y = ys * sqrt( matrix[1][0] * matrix[1][0] + matrix[1][1] * matrix[1][1] + matrix[1][2] * matrix[1][2]);
+      scale.z = zs * sqrt( matrix[2][0] * matrix[2][0] + matrix[2][1] * matrix[2][1] + matrix[2][2] * matrix[2][2]);
+      return scale;
+}
+
+vec4 billboard( const in vec3 vertex, const in mat4 modelViewMatrix, const in mat4 projectionMatrix ) {
+      vec3 scale = getScale( modelViewMatrix );
+      return projectionMatrix * ( vec4( scale.x* vertex.x , scale.y * vertex.y, scale.z * vertex.z, 1.0 ) + vec4( modelViewMatrix[ 3 ].xyz, 0.0 ) );
```
Kuranes: Seems very costly method to me, first time I see that much complexity in billboard shader.
Not even sure to get what it does.
perhaps 
_projectionMatrix_ **+** _modelViewMatrix_ 
should be 
_projectionMatrix_ **x** _modelViewMatrix_ 

I recall other faster methods for billboard but perhaps doesn't fit your specifications/needs  ?
Perhaps add a comment on why this one needed vs all the other ones like the ones in [this list](http://www.opengl-tutorial.org/fr/intermediate-tutorials/billboards-particles/billboards/)

jtorresfabra:
Indeed I was not very happy with this, I added that because the current implementation of billboarding did not respect the transforms above it. https://github.com/cedricpinson/osgjs/blob/develop/sources/osgShader/Compiler.js#L1410.

I will have a look to try to improve it. Anyway it does not affect the autotransform stuff as billboarding in AutoTransfrom now is done via JS as @cedricpinson asked for. That code was just to have a ShaderNode, instead of an Inline node.

jtorresfabra:
@Kuranes I'm just reviewing that stuff, but I'm afraid the method I did could be the faster one. The links you send to me only works for "particle/quad" billboards, where all the vertices lays in the same plane.

My solution is for 3D geometry billboarding, so I don't see a faster way to do it. I simply get the scale from the modelViewMatrix, multiply for each vertex and then translate it to the position extracted from the MVM, so the trick here is to not apply the rotation. That's way Projection * scale * vertex + translation.

Other method could be to extract the rotation from the MVM in a quat and then apply the inverse rotation to the MVM, but it would be more expensive than my method. 

Also we can have two ways to do BB, one for quads and one for 3D geometry... But this is out of the scope of this PR.

 Thoughts?

Kuranes:
"3D geometry billboarding" 
Ha, ok didn't understood the case. Makes more sense in that case indeed
Interesting, First time I hear about those, Do you have an example ?

my xp is billboarding is always 2D, "3D billboard" makes sense only if need of lighting but perf wise makes more sense to solve it with a 2D + heightmap and associated maps for lighting

jtorresfabra:
I can make a quick example, the use case is to have a 3D model always looking to the camera, being this a particle or not.

Kuranes:
no need to add yourself additionnal work :)
I don't get it, but if you need it, then it  sure does exist ;)
If you have use case and code only trigger on that case, code does makes sense to me and is ok.

jtorresfabra:
> I don't get it, but if you need it, then it sure does exist ;)

hehehe, well I hope it's not a crazy idea to make a model looking always to the camera. Thanks for your review anyway! :)

sources/osg/AutoTransform.js
```diff
@@ -0,0 +1,310 @@
...
+
+    computeWorldToLocalMatrix: ( function () {
+        var neg = Vec3.create();
+        var rotInverse = Quat.create();
+        var scaleInverse = Vec3.create();
+        return function ( matrix /*, nodeVisitor */ ) {
+            if ( this.scale[ 0 ] === 0.0 && this.scale[ 1 ] === 0.0 && this.scale[ 2 ] === 0.0 ) {
+                return false;
+            }
+            scaleInverse[ 0 ] = 1 / this._scale[ 0 ];
```
cedricpinson: use 1.0 instead of 1 optimization to avoid the vm use int and cast to float after

sources/osg/Matrix.js
```diff
@@ -604,6 +640,24 @@ var Matrix = {
         };
     } )(),
 
+    preMultRotate: function ( matrix, quat ) {
+        if ( Quat.zeroRotation( quat ) )
```
cedricpinson: not sure that we want to test each time the zero rotationI would prefer raw code instead of checking each time here

sources/osg/Matrix.js
```diff
@@ -604,6 +640,24 @@ var Matrix = {
...
+            return matrix;
+        var r = Matrix.create();
+        Matrix.makeRotateFromQuat( r, quat );
+        Matrix.preMult( matrix, r );
+        return matrix;
+    },
+
+    postMultRotate: function ( matrix, quat ) {
+        if ( Quat.zeroRotation( quat ) )
```
cedricpinson: same comment

sources/osg/Matrix.js
jtorresfabra: I'm not sure I get you. 

Do you want me to  

1. Subsitute `Quat.zeroRotation( quat )`  by   ` quat[ 0 ] === 0.0 && quat[ 1 ] === 0.0 && quat[ 2 ] === 0.0 && quat[ 3 ] === 1.0 `  
2. Put r matrix in a closure. 
3. Both.

sources/osg/Matrix.js
```diff
@@ -640,23 +640,27 @@ var Matrix = {
         };
     } )(),
 
-    preMultRotate: function ( matrix, quat ) {
-        if ( Quat.zeroRotation( quat ) )
+    preMultRotate: ( function () {
+        var r = matrixCreate();
+        return function ( matrix, quat ) {
+            if ( quat[ 0 ] === 0.0 && quat[ 1 ] === 0.0 && quat[ 2 ] === 0.0 && quat[ 3 ] === 1.0 )
```
cedricpinson: I meant to check at all

sources/osg/Matrix.js
```diff
@@ -640,23 +640,27 @@ var Matrix = {
...
-    },
+        };
+    } )(),
 
-    postMultRotate: function ( matrix, quat ) {
-        if ( Quat.zeroRotation( quat ) )
+    postMultRotate: ( function () {
+        var r = matrixCreate();
+        return function ( matrix, quat ) {
+            if ( quat[ 0 ] === 0.0 && quat[ 1 ] === 0.0 && quat[ 2 ] === 0.0 && quat[ 3 ] === 1.0 )
```
cedricpinson: same here

PR #576, opened 2016-05-10
Bugfix/errors gl dev 7552

Kuranes:
Strenghtens Framebuffer and texture code against errors during binding and creation, allowing better handling of user errors

### DIFF
--- a/sources/osg/FrameBufferObject.js
+++ b/sources/osg/FrameBufferObject.js
@@ -4,19 +4,21 @@ var MACROUTILS = require( 'osg/Utils' );
 var GLObject = require( 'osg/GLObject' );
 var StateAttribute = require( 'osg/StateAttribute' );
 var Timer = require( 'osg/Timer' );
-
+var WebglCaps = require( 'osg/WebGLCaps' );
 
 /**
  * FrameBufferObject manage fbo / rtt
  * @class FrameBufferObject
  */
 var FrameBufferObject = function () {
+
     GLObject.call( this );
     StateAttribute.call( this );
     this._fbo = undefined;
     this._rbo = undefined;
     this._attachments = [];
     this._dirty = true;
+
 };
 
 FrameBufferObject.COLOR_ATTACHMENT0 = 0x8CE0;
@@ -216,7 +218,14 @@ FrameBufferObject.prototype = MACROUTILS.objectInherit( GLObject.prototype, MACR
         var gl = this._gl;
 
         // apply on unit 1 to init it
+        // make sure we do bind it whatever state stack
+        // texture is cached
         state.applyTextureAttribute( 1, texture );
+
+        if ( texture.isDirty() || !texture.getTextureObject() ) {
+            // image wasn't ready, texture not allocated due to lack of gpu MEM
+            return false;
+        }
 
         gl.framebufferTexture2D( gl.FRAMEBUFFER, attachment, textureTarget, texture.getTextureObject().id(), 0 );
 
@@ -227,6 +236,8 @@ FrameBufferObject.prototype = MACROUTILS.objectInherit( GLObject.prototype, MACR
             Notify.log( 'FBO: texture: ' + texture.getName() + ' : ' + texture.getTextureObject().id().trackedObject.defaultName + ' fbo: ' + this._fbo.trackedObject.defaultName );
         }
         /* develblock:end */
+
+        return true;
     },
 
     bindFrameBufferObject: function () {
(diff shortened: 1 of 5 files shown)


sources/osg/FrameBufferObject.js
```diff
@@ -244,6 +255,19 @@ FrameBufferObject.prototype = MACROUTILS.objectInherit( GLObject.prototype, MACR
 
     },
 
+    checkAllowedSize: function ( w, h ) {
+
+        var maxSize = WebglCaps.instance().getWebGLParameter( 'MAX_RENDERBUFFER_SIZE' );
+
+        if ( w === 0 || h === 0 || h > maxSize || w > maxSize ) {
+            Notify.error( 'width or height makes frame buffer not bindable' );
```
cedricpinson: Maybe we can have more information in the error message:
width (value) or height (value) makes frame buffer not bindable. Max RenderBuffer is 'value'

sources/osg/State.js
```diff
@@ -457,31 +468,25 @@ State.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( Objec
         }
     },
 
-    applyTextureAttribute: function ( unit, attribute ) {
+    applyTextureAttribute: function ( unit, attribute, force ) {
```
cedricpinson: why adding force ?
It looks like it's not understood or we are not what's happened.
I dont want to change api because of that

sources/osg/Texture.js
```diff
@@ -235,9 +236,20 @@ Texture.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( GLO
 
     setTextureSize: function ( w, h ) {
 
-        if ( w !== undefined ) this._textureWidth = w;
-        if ( h !== undefined ) this._textureHeight = h;
+        var maxSize = WebglCaps.instance().getWebGLParameter( 'MAX_TEXTURE_SIZE' );
 
+        if ( w !== undefined ) {
+            if ( w > maxSize ) {
+                Notify.error( 'Texture too big for GPU' );
```
cedricpinson: adds value in the log message

sources/osg/Texture.js
```diff
@@ -235,9 +236,20 @@ Texture.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( GLO
...
+            }
+            this._textureWidth = w;
+        }
+        if ( h !== undefined ) {
+            if ( h > maxSize ) {
+                Notify.error( 'Texture too big for GPU' );
```
cedricpinson: same here

sources/osg/State.js
```diff
@@ -751,10 +752,7 @@ State.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( Objec
             var attribute = attributePair.getAttribute();
 
             if ( attributeMap[ type ] === undefined ) {
-                attributeMap[ type ] = new Stack();
-                attributeMap[ type ].globalDefault = attribute.cloneType();
-
-                attributeMap.dirty();
+                this.createAttributeStack( attributeMap, type, attribute );
```
cedricpinson: typo ? it should be attribute.cloneType() ?

sources/osg/FrameBufferObject.js
```diff
@@ -216,7 +218,14 @@ FrameBufferObject.prototype = MACROUTILS.objectInherit( GLObject.prototype, MACR
         var gl = this._gl;
 
         // apply on unit 1 to init it
-        state.applyTextureAttribute( 1, texture );
+        // make sure we do bind it whatever state stack
+        // texture is cached
+        state.applyTextureAttribute( 1, texture, true );
```
cedricpinson: if it's not sync there is a bigger problem to solve. I would prefer we fix the bigger problem if there is.
Could you describe why you need to force a bind ?

sources/osg/State.js
Kuranes: No use case.
Framebuffer like VAO, if texture not really binded (not cache) just before, it's not framebuffer associated.
That's the simple way to make sure we do bind it.

sources/osg/FrameBufferObject.js
Kuranes: like above, Framebuffer like VAO, if texture not really binded (not cache) just before, it's not framebuffer associated.
That's the simple way to make sure we do bind it.

It's part of the enforce better log/sentry by forcefully following spec, not a case I reproduced.
Doesn't cost much to add, but allow to make sure we're doing the right thing

cedricpinson: I prefer to have code if it makes sense. The just in case could mask something else. Could you remove it plz

sources/osg/State.js
```diff
@@ -424,10 +424,25 @@ State.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( Objec
         }
     },
 
+    createAttributeStack: function ( attributeMap, key, globalDefault ) {
```
cedricpinson: probably use _createAttributeStack because it's more internal/private method

sources/osg/FrameBufferObject.js
```diff
@@ -244,6 +255,19 @@ FrameBufferObject.prototype = MACROUTILS.objectInherit( GLObject.prototype, MACR
 
     },
 
+    checkAllowedSize: function ( w, h ) {
```
cedricpinson: Also a internal method too so use _checkValidSize

PR #577, opened 2016-05-12
Improves FPS manipulator for WebVR support

stephomi:


sources/osgGA/FirstPersonManipulator.js
```diff
@@ -56,14 +57,17 @@ FirstPersonManipulator.prototype = MACROUTILS.objectInherit( Manipulator.prototy
...
-        this._target = Vec3.create();
         this._angleVertical = 0.0;
         this._angleHorizontal = 0.0;
 
         // tmp value use for computation
         this._tmpGetTargetDir = Vec3.create();
 
-        this._rotBase = Matrix.create();
+        // vr controls
+        this._vrEnable = false;
```
cedricpinson: We should not have two differents way to represent rotation inside the Manipulator. It would make more sense to move angleVertical/angleHorizontal in mouse device, same logic for vrRot/Trans in webVR device, and using only simple representation of translation / rotation inside the manipulator. That would be easier for the manipulator, not sure it would work without issue.
Did you think about that or is it too complicated for the current state of the manipulator ?

stephomi: The only way to have several controllers controlling a manipulator is:
- the manipulator only exposes a view matrix and it's the controllers responsability to handle it
- the manipulator exposes custom stuffs (like angleVertical/angleHorizontal for example), in that case all the controls should edit those

The second solution is limited because you can't do rool axis vr movement with only 2 angles.
The first solution is better imho, but it raises the question of the limit/constraints, it's not clear if it's the manipulator or controllers  reponsability to handle these.

I'm thinking it should be the controllers, for example in fps manipulator you don't want any constraints when you move your head in VR, but with the mouse controller you never want to go on the other side.

Also with the first solution, I'm not sure of the benefit of having several manipulator if they only exposes a viewMatrix to the controllers (in that case it will be the controllers that will do the heavy works).

cedricpinson: Agree we would probably needs something in the middle like simple internal representation but with helpers/config to set rotation/position in the manipulator.
I will merge it like that for now

PR #579, opened 2016-05-16
Fixes VAO picking

stephomi:


stephomi:
The changes of apply( this ) to call( this ) (so without arguments) is not about performance but about semantics.
call is more specific than apply so it should be the prefered version.

sources/osg/Geometry.js
```diff
@@ -328,7 +328,10 @@ Geometry.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( No
                 ];
                 for ( j = 0, m = listVABuff.length; j < m; j++ ) {
                     vaoSetup.push( '  va = vaList[ ' + j + '];' );
-                    vaoSetup.push( '  if ( va.isDirty() ) va.compile( gl );' );
+                    vaoSetup.push( '  if ( va.isDirty() ) {;' );
+                    vaoSetup.push( '    va.bind( gl );' );
```
stephomi: the buffer should be bound before re-compiling it.

stephomi:
maybe same issue here https://github.com/cedricpinson/osgjs/blob/develop/sources/osg/Geometry.js#L225

cedricpinson:
Yeah same issue, could you fix it too plz ?

stephomi:
k done

PR #581, opened 2016-05-19
Fixes tests with typed array

stephomi:


cedricpinson:
The behavior before was valid. it's better to add setInitialBoundingSphere in node, and call this method in sketchfab for our LightSource

stephomi:
k

PR #586, opened 2016-05-30
Fix image stream at end

cedricpinson:


sources/osg/ImageStream.js
```diff
@@ -25,12 +25,19 @@ ImageStream.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit(
...
-            this._status = ImageStream.PAUSE;
+            if ( !this._imageObject.loop )
+                this.stop();
         }.bind( this ), true );
 
         this.dirty();
     },
 
+    setLooping: function ( bool ) {
+        if ( this._imageObject.loop === bool ) return;
```
stephomi: really necessary? :no_mouth:

cedricpinson: probably not

PR #590, opened 2016-06-03
Precision picking

stephomi:


sources/osg/Matrix.js
```diff
@@ -658,60 +658,43 @@ var Matrix = {
         return result;
     },
 
-    transformVec3: ( function () {
-        var tmpVec = Vec3.create();
+    transformVec3: function ( matrix, vector, result ) {
+        if ( !result ) {
```
cedricpinson: maybe we could remove all result condition now

PR #592, opened 2016-06-07
Moves submitFrame in frame function for easier hook

stephomi:
makes it easier to hook a function so that we can mirror something on the screen

### DIFF
--- a/sources/osgViewer/Viewer.js
+++ b/sources/osgViewer/Viewer.js
@@ -114,6 +114,7 @@ var Viewer = function ( canvas, userOptions, error ) {
     this.setUpView( gl.canvas, options );
 
     this._hmd = null;
+    this._requestAnimationFrame = window.requestAnimationFrame.bind( window );
 
     this._contextLost = false;
 };
@@ -390,6 +391,10 @@ Viewer.prototype = MACROUTILS.objectInherit( View.prototype, {
         }
 
         this.endFrame();
+
+        // submit frame to vr headset
+        if ( this._hmd && this._hmd.isPresenting )
+            this._hmd.submitFrame( this._eventProxy.WebVR._lastPose );
     },
 
     setDone: function ( bool ) {
@@ -404,18 +409,8 @@ Viewer.prototype = MACROUTILS.objectInherit( View.prototype, {
         var self = this;
         var render = function () {
             if ( !self.done() ) {
-
-                var isPresentingVR = self._hmd && self._hmd.isPresenting;
-
-                if ( isPresentingVR )
-                    self._requestID = self._hmd.requestAnimationFrame( render );
-                else
-                    self._requestID = window.requestAnimationFrame( render, self.getGraphicContext().canvas );
-
+                self._requestID = self._requestAnimationFrame( render, self.getGraphicContext().canvas );
                 self.frame();
-
-                if ( isPresentingVR )
-                    self._hmd.submitFrame( self._eventProxy.WebVR._lastPose );
             }
         };
         render();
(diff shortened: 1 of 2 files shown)


sources/osgViewer/Viewer.js
```diff
@@ -404,18 +408,12 @@ Viewer.prototype = MACROUTILS.objectInherit( View.prototype, {
         var self = this;
         var render = function () {
             if ( !self.done() ) {
-
-                var isPresentingVR = self._hmd && self._hmd.isPresenting;
-
-                if ( isPresentingVR )
+                if ( self._hmd && self._hmd.isPresenting )
```
jtorresfabra: @stephomi as you are fixing this code ( and sorry for bothering )... would you mind to wrap the requestAnimationFrame in a function in a way that we don't do a comparison every frame? So if you set VR in the viewer self._hmd.requestAnimationFrame is called and by default window.requestAnimationFrame is called. 
I recall to speak about this with @cedricpinson some time ago.

stephomi: The signatures are a bit different (the vr requestAnimation doesn't take a html 2nd element in argument for example).
Also, if I'm not mistaken, isPresenting must be read at each frame as it's the only way to know if we have to render in vr or not.

What's the issue with the conditional? It takes 0 perf and should be completely transparent in non vr.

jtorresfabra: > The signatures are a bit different (the vr requestAnimation doesn't take a html 2nd element in argument for example).

In fact that second element is optional, and only for chrome webkit AFAIK( https://developer.mozilla.org/es/docs/Web/API/Window/requestAnimationFrame ). 

> Also, if I'm not mistaken, isPresenting must be read at each frame as it's the only way to know if we have to render in vr or not.

I don't know, it that is really the case then it does not make sense to avoid the conditional. But from what I've read : " When the VRDisplay is not presenting a scene, this is functionally equivalent to Window.requestAnimationFrame."
(https://developer.mozilla.org/en-US/docs/Web/API/VRDisplay/requestAnimationFrame)

> What's the issue with the conditional? It takes 0 perf and should be completely transparent in non vr.

It's just that I don't feel comfortable doing an extra conditional every frame if it is really not needed. It does not look like a best practice from the design point of view.

I could be wrong of course, but in my opinion we can avoid that conditional. 

Thoughts?

stephomi: > It's just that I don't feel comfortable doing an extra conditional every frame if it is really not needed.

^ that's really nothing, it's not even a function, no dynamic lookup or math, no allocation, just check the visitor/rendering code in osgjs and will see that it's literally nothing in comparison of what is already done.

Take for example the begin/end Frame that does many many useless tests... https://github.com/cedricpinson/osgjs/blob/develop/sources/osgViewer/Viewer.js#L326-Lundefined

Anyway I updated the PR, as long as the 2nd parameter doesn't hinder the vr function it's fine by me

jtorresfabra: > ^ that's really nothing, it's not even a function, no dynamic lookup or math, no allocation, just check the visitor/rendering code in osgjs and will see that it's literally nothing in comparison of what is already done.

Yes, I understand your point, clearly we won't experience a better perf. But the render function is somehow a key function, and my point is more from the design, not a perf improvement. 

> Take for example the begin/end Frame that does many many useless tests... https://github.com/cedricpinson/osgjs/blob/develop/sources/osgViewer/Viewer.js#L326-Lundefined

Ouch, this clearly could be rewrited. 

> Anyway I updated the PR, as long as the 2nd parameter doesn't hinder the vr function it's fine by me

Thank you! :+1:

PR #597, opened 2016-06-17
Adds documentation js

cedricpinson:
- Needs to add jsdoc comments on files to have a good documentation :)


sources/osg/BlendColor.js
```diff
@@ -3,10 +3,13 @@ var MACROUTILS = require( 'osg/Utils' );
 var StateAttribute = require( 'osg/StateAttribute' );
 var Vec4 = require( 'osg/Vec4' );
 
+/** @namespace osg */
```
jtorresfabra: It should not be here, it should be in sources/osg/osg.js . From the How To Document file: use @namespace tag to document the namespace in the file where the module is declared. Anyway it will be clearer when the template is applied.:).

PR #602, opened 2016-06-23
Fixed pre/postmultrotate methods

jtorresfabra:


sources/osg/Matrix.js
```diff
@@ -652,9 +653,9 @@ var Matrix = {
     postMultRotate: ( function () {
         var r = matrixCreate();
         return function ( matrix, quat ) {
-            Matrix.makeRotateFromQuat( r, quat );
+            Matrix.makeRotateFromQuat( quat, r );
             Matrix.postMult( matrix, r );
-            return matrix;
+            return r;
```
stephomi: r is a local temporary variable, shouldn't it be Matrix.postMult(r, matrix) ?

jtorresfabra: humm..Sorry I'm probably a bit slow today,  I don't get it:
The result of doing premult( a, b ) should be equal to postmult( b, a ). Isn't it? 
So if I do what you say,( postmult( r, matrix ) ) then is the same as premult(matrix, r ), which is not the intended way. 

What I'm missing?

stephomi: preMult( a, b ) -> a = a * b
postMult( b, a ) -> a = b * a
The "pre" and "post" simply tell where the result will be stored

jtorresfabra: So postmult(a,b) gives the same result as premult(a,b) ? The only thing that changes is the variable where you store the result? 

This is a bit confusing. 
Tomorrow I will review the maths. Thanks for your comments :+1:

stephomi: > So postmult(a,b) gives the same result as premult(a,b) ? The only thing that changes is the variable where you store the result?

Yes that's it.
You can use mult(a, b, result), which is not confusing and result can be either a or b.

jtorresfabra: I'm gonna test now, but matrices don't have conmutative property, so if what you say is true, then is somethign too wrong in osgjs.

jtorresfabra: Ok so what you say is true, but the maths in osgjs are wrong in my opinion.

jtorresfabra: @stephomi changed the PR to follow osgjs convention. I talked whith @cedricpinson  and all is clear.

PR #603, opened 2016-06-24
Makes the gizmo more flexible

stephomi:
The gizmo is more flexible for example you can set a flag on a node that only says "you can rotate me in x, and translate me in x and z".

### DIFF
--- a/sources/osgUtil/NodeGizmo.js
+++ b/sources/osgUtil/NodeGizmo.js
@@ -161,15 +161,22 @@ var NodeGizmo = function ( viewer ) {
 
 // picking masks
 NodeGizmo.NO_PICK = 1 << 0;
-NodeGizmo.PICK_ARC = 1 << 1;
-NodeGizmo.PICK_ARROW = 1 << 2;
-NodeGizmo.PICK_PLANE = 1 << 3;
+NodeGizmo.PICK_ARC_X = 1 << 1;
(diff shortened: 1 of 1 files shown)


sources/osgUtil/NodeGizmo.js
```diff
@@ -271,11 +278,27 @@ NodeGizmo.prototype = MACROUTILS.objectInherit( MatrixTransform.prototype, {
...
+
         var mask = this._attachedNode.editMask;
+
         this.setNodeMask( mask & NodeGizmo.PICK_GIZMO ? NodeGizmo.NO_PICK : 0x0 );
+
         this._translateNode.setNodeMask( mask & NodeGizmo.PICK_ARROW ? NodeGizmo.PICK_ARROW : 0x0 );
         this._rotateNode.setNodeMask( mask & NodeGizmo.PICK_ARC ? NodeGizmo.PICK_ARC : 0x0 );
         this._planeNode.setNodeMask( mask & NodeGizmo.PICK_PLANE ? NodeGizmo.PICK_PLANE : 0x0 );
+
+        this._translateNode.children[ 0 ].setNodeMask( mask & NodeGizmo.PICK_ARROW_X ? NodeGizmo.PICK_ARROW_X : 0x0 );
```
cedricpinson: dont access children in node, use an accessors

PR #606, opened 2016-06-28
Fix for coverity issues

jtorresfabra:


sources/osgDB/Input.js
```diff
@@ -504,6 +504,7 @@ Input.prototype = {
                 type = 'Uint8Array';
             } else {
                 Notify.warn( 'Typed Array ' + window.Object.keys( jsonObj.Array )[ 0 ] );
+                vb = jsonObj.Array.Float32Array;
```
cedricpinson: why ?

jtorresfabra: It seems I mislead the case. 
The point is that vb could potentially be undefined as it is created here https://github.com/cedricpinson/osgjs/blob/master/sources/osgDB/Input.js#L495 and then accesed here https://github.com/cedricpinson/osgjs/blob/master/sources/osgDB/Input.js#L510

Adding this could be enough.
`if ( vb !== undefined && vb.File ) {`

PR #619, opened 2016-07-15
Performance Improvement

cedricpinson:


sources/osg/State.js
```diff
@@ -1157,48 +1150,142 @@ State.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( Objec
...
+        if ( internalArray.length < 16 ) {
+            if ( internalArray instanceof Float32Array ) {
+                cacheData = new Float32Array( internalArray.length );
+            } else if ( internalArray instanceof Int32Array ) {
+                cacheData = new Int32Array( internalArray.length );
+            } else if ( internalArray instanceof Array ) {
+                cacheData = new Array( internalArray.length );
+            } else {
+
+                console.log( 'not found' );
```
jtorresfabra: Notify.log?

cedricpinson:
performance branch
screen shot 2017-04-05 at 18 46 51

master branch
screen shot 2017-04-05 at 16 52 29

stephomi:
On the performance example without complexStateSet=true the performance branch is a lot slower for me (items 4 and depth 3).

From left to right : for-in-experiment, master, performance-test
Test with multiple loading and it's consistent

sans titre

cedricpinson:
I merged the options for complex stateSet in master

sources/osg/Utils.js
```diff
@@ -167,6 +167,117 @@ Utils.createPrototypeClass = function ( Constructor, prototype, libraryName, cla
...
+
+// ===============================================
+
+
+var typeMemberIndex = 0;
+var textureTypeMemberIndex = 0;
+var stateAttributeTypeMember = {};
+var textureStateAttributeTypeMember = {};
+var attributeTypeIndex = 0;
+var stateAttributeType = {};
```
jtorresfabra: Not sure is better to use maps, to avoid duplication

sources/osg/CullVisitor.js
```diff
@@ -76,8 +76,19 @@ var CullVisitor = function () {
 
 };
 
+CullVisitor.registerApplyNodeType = function ( type, apply ) {
```
cedricpinson: agree

sources/osg/CullVisitor.js
```diff
@@ -477,7 +488,7 @@ CullVisitor.prototype = MACROUTILS.objectInherit( CullStack.prototype, MACROUTIL
 // Camera cull visitor call
 // ANY CHANGE, any change : double check in rendere Camera code
 // for the first camera
-CullVisitor.prototype[ Camera.typeID ] = function ( camera ) {
+var CameraApply = function ( camera ) {
```
cedricpinson: agree

sources/osg/Utils.js
cedricpinson: I dont understand {} there is no risk of duplication ?

jtorresfabra: forget it.

Kuranes:
👍 Love it.
++ utils for osg object creation
++ cullVisitorHelper.registerApplyFunction 

in the helper in utils to create prototype of classes, could even had some assert (deveblock) checking for missing mandatory methods. ( for StateAttribute I would the "clone" which I had problem with each time I made a new stateattribute )

cedricpinson:
@Kuranes Yeah that would be great to have cloneType also built by default in this helper, but we have some class that needs to do something specific  like:
```
    cloneType: function () {
        return new ShadowCastAttribute( true );
    },
```
It's not the only one, but it means we should change the default behavior to be sure default Constructor generate a default state. It's normal that in some case it needs something specific, but we will check that later

jtorresfabra:
Needs rebase now :(

PR #624, opened 2016-07-17
Enhance performance sample

Kuranes:
Adds gui and option to be able to test more performances related renderings

Mesh related:
- deepnees of tree
- number of item of branches
- use different geom at each item or only one item ("instancing")

Stateset/Texture
- with or without texture (which makes it stateset -less)
- number of texture
- texture size

### DIFF
--- a/examples/performance/index.html
+++ b/examples/performance/index.html
@@ -6,13 +6,20 @@
     <link rel="stylesheet" type="text/css" href="../templates/css/base.css">
 
     <!-- Javascript -->
+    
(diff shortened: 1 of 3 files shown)


jtorresfabra:
@Kuranes Is it possible that you want to merge against master and not develop?

Kuranes:
( unclear to me is the new workflow. will ask )

cedricpinson:
Workflow is the same just master is the main branch now 

Cedric

cedricpinson:
I am closing it to avoid confusion. should be rebased on master

PR #628, opened 2016-07-18
Adds udpate callback counter to avoid traversing

cedricpinson:


sources/osgUtil/NodeGizmo.js
```diff
@@ -535,10 +535,11 @@ NodeGizmo.prototype = MACROUTILS.objectInherit( MatrixTransform.prototype, {
 
     getTransformType: function ( node ) {
         var n = node;
-        while ( n.parents.length > 0 ) {
+        var parents = n.getParents();
```
stephomi: Infinite loop... should be ``while(n.getParents().length)`` and ``n = n.getParents()[0]``

sources/osg/UpdateVisitor.js
```diff
@@ -37,15 +37,18 @@ UpdateVisitor.prototype = MACROUTILS.objectInherit( NodeVisitor.prototype, {
         // handle callback in nodes
         var ncs = node.getUpdateCallbackList();
         var numUpdateCallback = ncs.length;
-        if ( numUpdateCallback )
-            for ( var j = 0, m = numUpdateCallback; j < m; j++ ) {
+
+        if ( numUpdateCallback ) {
```
stephomi: The condition is not necessary here.

sources/osg/StateSet.js
```diff
@@ -167,14 +176,40 @@ StateSet.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( Ob
...
+            var parents = this._parents;
+            for ( var i = 0, l = parents.length; i < l; i++ ) {
+                var parent = parents[ i ];
+                parent.setNumChildrenRequiringUpdateTraversal( parent.getNumChildrenRequiringUpdateTraversal() - 1 );
+            }
+        }
+    },
+
+    requiresUpdateTraversal: function () {
+        return Boolean( this._updateCallbackList.length );
```
stephomi: If the function is called many times, "!!" is prefered over "Boolean" for performance reason.
https://jsfiddle.net/m40qbpnf/ (jsperf is down...)

cedricpinson:
thx

PR #629, opened 2016-07-19
Improves performance sample

Kuranes:
add gui and option to be able to test more performances related renderings

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -84,8 +84,7 @@ module.exports = {
             // shaders
             test: /\.(frag|vert|glsl)$/,
             loader: 'raw-loader'
-        }
-                 ]
+        } ]
     },
     devtool: 'source-map',
     plugins: [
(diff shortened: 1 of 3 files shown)


Kuranes:
Rebased .

It's a really very practical and important PR for optimization.

You can reproduce a "problematic" situation in a few click, and profile it in the simplest context possible.

Previous perfomance test is very misleading, notably because of it's "instancing by default" which makes it a scene with ony one geom/texture....

Only thing not tweakable now is shader/program, but will do a second PR for that allowing to test separatly program switching, fragment shader load and texture fetch cost

webpack.config.js
```diff
@@ -84,8 +84,7 @@ module.exports = {
             // shaders
             test: /\.(frag|vert|glsl)$/,
             loader: 'raw-loader'
-        }
```
cedricpinson: could you remove this change from the PR

PR #630, opened 2016-07-19
[#9999] [] Adds glsl preprocessor & parser

Kuranes:
For now under url options:
 &enableShaderOptimizer=1&enableShaderCompilationTiming=1
- preprocessor to make it simple
- parser to allow optimisation
- 2 optimisations for now:
  *\* removes unused variables
  *\* removes unused function
- possible "obfuscation" in prod mode: var rename, comment remove, uneeded line feed, etc


cedricpinson:
What is the size impact including those deps ?

Kuranes:
gz compressed comparison:
before: 285kb
after: 294kb 

uncompressed:
before: 1 517kb
after:  1570kb

Kuranes:
TODO:
- add url option to enable disable preprocess, optimisation pass
- add url option to time compîlaton
- add better compilation log with shader name, and browse/gpu compilation time.

cedricpinson:
Is there a way to use like optimization like glsl optimizer ?

Kuranes:
for now it's just tree shaking, but who knows if some bugged drivers aren't helped by shaders going from > 1000 lines to 250lines...

Others optimization needs to be coded it :)

The order of things to be done to makes each optimization useful for the next :
- function inlining (using descope as a basis https://www.npmjs.com/package/glsl-token-descope)
- variables reuse (https://www.npmjs.com/package/glsl-token-assignments)
- texFetch dedup (exact same texFecth => reuse var)
- precision (https://www.npmjs.com/package/glsl-token-assignments + finding who ends in glFragColor)
- math simplifications (from simplifying constant expression https://github.com/silentmatt/js-expression-eval/tree/master to enable approximations)

Kuranes:
Usage: add &enableShaderCompilationTiming=1 to url

Internal refactoring to allow shader & program naming to be less of a hack and
more of a real part of osgjs.

Shade name allow easiear debugging, logging, timing and is used by
external tools like chrome shader editor.
In the same spirit of easier code handling there is now a
ShaderProgramBuilder, for faster code/recompilation/require and single entry point
instead of code repeated everywhere.

(both can be used, PR is breaking changes free)

Kuranes:
On second thought, probably preprocessor & optimizer should go into shaderprogrambuilder and be removed of shaderprocessor, would make much more sense:
- we would have  shadername in the preprocessor&optimizer timing/logging to too 
- CompilationTimingOn could be a parameter option of osg.shader and osg.program ctors 

(currently no shadername in preprocessor and optimizer is making skfb logs a mess, cannot be sure which is for which)

Kuranes:
TODO:
- find how to give shadername to preprocessor & optimizer
- add option to prevent trimming strings for processed/optimized but still readable shader code

Kuranes:
Now works everywhere with logging, and clearly separated from the shaderprocessor and shadercompiler, into th shaderprogrambuilder:

urloptions is:
&enableShaderOptimizer=1&enableShaderCompilationTiming=1

Kuranes:
rebased with better expression evaluator in preprocessor.
Allows for #ifdef code based on shader names
```glsl




#if SHADER_NAME == hereyoujustputthenameoftheshaderandHopnoemoretexture
#define _TEXTURELESS
#endif

#ifndef _TEXTURELESS 

#define textureRGB( tex, uv) texture2D(tex, uv.xy ).rgb
#define textureRGBA( tex,  uv) texture2D(tex, uv.xy ).rgba
#define textureIntensity(tex, uv) texture2D(tex, uv).r
#define textureAlpha( tex, uv) texture2D(tex, uv.xy ).a

#else

#define textureRGB( tex, uv) vec3(1.0*gl_FragCoord.x, 0.5, 1.0*gl_FragCoord.y)
#define textureRGBA( tex,  uv) vec4(1.0*gl_FragCoord.y, 0.5, 1.0*gl_FragCoord.x, 1.0)
#define textureIntensity(tex, uv) .5 + 0.1 * gl_FragCoord.x
#define textureAlpha( tex, uv) 1.0 - 0.1 * gl_FragCoord.y

#endif

```

sources/osg/Program.js
```diff
@@ -117,7 +126,12 @@ Program.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( GLO
     getProgram: function () {
         return this._program;
     },
-
+    getName() {
```
cedricpinson: already exist in Object.js Program is inherited from this

sources/osg/Program.js
```diff
@@ -6,15 +6,23 @@ var StateAttribute = require( 'osg/StateAttribute' );
...
+var defineShaderNameReg = /#define\sSHADER_NAME\s(\S+)/im;
+var unnamedProgrammNumber = 0;
+/*develblock:end*/
+
 /**
  * Program encapsulate an vertex and fragment shader
  * @class Program
  */
-var Program = function ( vShader, fShader ) {
+var Program = function ( vShader, fShader, name, doTiming ) {
```
cedricpinson: dont change the signature, name can be called with setName. doTiming can be computed at runtime that could depends on log level info.
Can check Notify level if it's <= info you can display timing. It will be display for debug and info level

sources/osg/Shader.js
```diff
@@ -78,10 +75,35 @@ Shader.prototype = MACROUTILS.objectInherit( GLObject.prototype, {
...
+    },
+    getName: function () {
+        return this._name;
+    },
+    getDefineShaderName: function () {
+        if ( !this._name ) return '';
+        return '\n#define SHADER_NAME ' + this._name + '\n';
+    },
+
+    getType: function () {
```
cedricpinson: rename this.type to this._type in the same time you touch this file.

sources/osgShader/ShaderGenerator.js
```diff
@@ -215,10 +214,13 @@ ShaderGenerator.prototype = {
                 return cache;
             }
 
+            this._osgShader = this._osgShader || require( 'osgShader/osgShader' );
+            var doTimeCompilation = this._osgShader.enableShaderCompilationTiming;
```
cedricpinson: same thing than the debug/info log level. if ?log=info is not enough then yeah maybe we need a global flag.

sources/osgShader/ShaderProgramBuilder.js
```diff
@@ -0,0 +1,76 @@
+'use strict';
```
cedricpinson: filename is not well named. ShaderProgramBuilder means it's a class. shaderProgramBuilder means it exposes functions / lib

cedricpinson: I dont understand this file, it seems that we dont need it. Why we can't just update Program

Kuranes: Program would become osgShader, (optimizer, preprocessor) dependant.

ShaderProgramBuilder  is at the corner of Shader, Shadergenerator/Compiler, Composer and ShaderPreprocessor/preprocessor/optimizer
(imho it should do the shaderprocessor thing too)

Kuranes:
Things that PR try to adress:

- Adding shader names as part of the lib rather than hack on the side

- Timing Shaders work process ( all substeps per shader/program, from shader node compiler to program linking)

- OSGJS internal code should use a single entry point in osgShader for shaders/program 

- Uses that entry point for logs, timing, pruning, and linking ( later optimize, transpiling to webgl2 etc). 

- All that not breaking  osgjs vanilla shader + program usage.

Things that could be done: all shader going through the single entry point should be going through shaderprocessor. (for now composer shader doesn't for instance.)

The single Entry Point is really really a big point, it adresses:
- Easier code debugging and reading (single place, all goes through same place)
- Easier Shader "processing" code evolution  (single place code change)
- DRY code. (many many many place do the 2 shader + program build + instance shaderprocessor etc)
- Clear sepearation  osg / osgShader usage of shader. (vanilla shader VS augmented shader)


Single entry point could be done differently, like done in Program, but would need lots of listener/callback inside that function to call all the osgShader parts (pre/post shader individual processing, pre/post shaders processing, pre/post program linking )

Kuranes:
not sure of "need fix" todoif I want to continue on the shader optimization ?

If move the shaderProgrambuilder into Program it's ok ?
(osg "core" will depend on "osgShader" then, but not very important indeed)

cedricpinson:
I feel that if we dont focus on it, it will take 6 month to have something working when a big amount of discussions/comments I prefer to focus on subject already open and that needs to be finished.
I am not against the idea it's just I dont have a solution to avoid this huge energy needed to converge faster. It will take too much energy to rebase / fix to follow master. So I prefer to stop thinking on this, and when we are ready we could do it and merge it in a shorter timerange.
If you have a better way to handle this, I am listening. 

I am still thinking it could be good especially for mobile.

Kuranes:
ok, I'll close this and save it in a separate repo then, as  "app side" code, and do my geeking there

PR #632, opened 2016-07-20
[#D3D-2390] Fixes continuous Move in FPS

Kuranes:
When user keep key down, the intepolator in FirstPersonManipulator reach
the "isDone()" state, with "_current === _target", making interpolator
useless.

Added a state "MovingAction" put by FirstPesonnManipulatorStandardMouseKeybardController, that when isDone && movingAction reset te current to 0.


Kuranes:
Must be done differently

PR #633, opened 2016-07-20
Fixes text traversal

jtorresfabra:
Text needs to do is own traversal in order to keep track of modelview matrix

### DIFF
--- a/sources/osgText/Text.js
+++ b/sources/osgText/Text.js
@@ -8,8 +8,9 @@ var MatrixTransform = require( 'osg/MatrixTransform' );
 var Shape = require( 'osg/Shape' );
 var Texture = require( 'osg/Texture' );
 var BlendFunc = require( 'osg/BlendFunc' );
-
-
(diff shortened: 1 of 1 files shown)


sources/osgText/Text.js
```diff
@@ -242,15 +246,89 @@ Text.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( AutoTr
...
+                    }
+                    this._previousWidth = width;
+                    this._previousHeight = height;
+                    Vec3.copy( position, this._previousPosition );
+                    Matrix.copy( projMat, this._previousProjection );
+                    Matrix.copy( modelViewMat, this._previousModelView );
+                }
+            }
 
+            Node.prototype.accept.call( this, visitor );
```
cedricpinson: why not autotransform ?

jtorresfabra: Because I'm overriding the method from Autotransform.

cedricpinson: ok

PR #638, opened 2016-07-23
Small optimizations in glsl code

jtorresfabra:
Just for test, I get some improvement with the example in https://github.com/cedricpinson/osgjs/pull/629 . Enabling the stats it's enough to see some FPS gain, at least in my old MBP. It could be nice if any of you can test this.

it seems the difference is not quite noticeable in my linux box.


jtorresfabra:
Closing for now, I didn't merge the master branch into de the perf one to do the comparison

PR #639, opened 2016-07-24
Fixed error in color encode

jtorresfabra:
It seems like a copy/paste error as 255^3 = 16581375 and not 160581375


stephomi:
Probably copy pasted from http://aras-p.info/blog/2008/06/20/encoding-

jtorresfabra:
Indeed http://aras-p.info/blog/2009/07/30/encoding-floats-to-rgba-the-final/ . In the comments of this post people also noticed the error.

PR #641, opened 2016-08-02
Adds optional url option to use glsl optimizer

cedricpinson:
uses ?GLSLOptimizer=1 to use glsl optimizer
if you want to see the shader results use &log=info


cedricpinson:
needs to refine logs....

PR #653, opened 2016-09-08
Adds integration of extDrawBuffers

jtorresfabra:


sources/osg/FrameBufferObject.js
```diff
@@ -252,6 +252,10 @@ FrameBufferObject.prototype = MACROUTILS.objectInherit( GLObject.prototype, MACR
         if ( status !== 0x8CD5 ) {
             this._reportFrameBufferError( status );
         }
+        if ( gl.checkFramebufferStatus( gl.FRAMEBUFFER ) !== gl.FRAMEBUFFER_COMPLETE ) {
```
jtorresfabra: I think this could be removed, as I recall to read it is a expensive call, but in https://www.khronos.org/opengles/sdk/docs/man/xhtml/glCheckFramebufferStatus.xml says that: 
"It is strongly advised, thought not required, that an application call glCheckFramebufferStatus to see if the framebuffer is complete prior to rendering. This is because some implementations may not support rendering to particular combinations of internal formats. In this case, GL_FRAMEBUFFER_UNSUPPORTED is returned."

jtorresfabra:
@cedricpinson @Kuranes @stephomi  I would like to know your opinion in the extension integration, now it is done at FrameBufferObject level only. But we may add a flag in Camera to enable drawBuffers exclusively in each camera object needing it, so the camera should be the responsible to enable drawBuffers in its FrameBufferObject.

sources/osg/FrameBufferObject.js
```diff
@@ -285,6 +289,14 @@ FrameBufferObject.prototype = MACROUTILS.objectInherit( GLObject.prototype, MACR
 
                 this.bindFrameBufferObject();
 
+
+                // Check extDrawBuffers extension
+                var extDrawBuffers = this._extDrawBuffers;
+                if ( extDrawBuffers === undefined ) { // will be null if not supported
+                    extDrawBuffers = gl.getExtension( 'WEBGL_draw_buffers' );
```
Kuranes: should be using webglcaps to get extension.
(as extension could be filtered out)

sources/osg/FrameBufferObject.js
Kuranes: you're adding exact same gl call as lines above no ?
the gl.checkFramebufferStatus then status !== 0x8Cd5 
cannot remember if do we call checkstatus at each call or only at creation ?

examples/drawbuffers/main.js
```diff
@@ -0,0 +1,271 @@
...
+                'uniform sampler2D Texture3;',
+                'uniform sampler2D DepthTexture;',
+                'void main (void)',
+                '{',
+                '  vec3 color = texture2D( Texture0, FragTexCoord0.xy).xyz;',
+                '  gl_FragData[0] = vec4( color * texture2D( DepthTexture, FragTexCoord0.xy).xyz, 1.0 );',
+                '}',
+                ''
+            ].join( '\n' );
+            var program = new osg.Program( new osg.Shader( 'VERTEX_SHADER', vertexShader ), new osg.Shader( 'FRAGMENT_SHADER', fragmentShader ) );
```
Kuranes: you added shaderPath to Example but do not load shader from glsl files :(

examples/drawbuffers/main.js
```diff
@@ -0,0 +1,271 @@
...
+                this._textureList[ i ]._name = 'texturecolor' + i;
+            }
+            var depthTexture = this.createDephTexture( width, height );
+            this._textureList.push( depthTexture );
+            // attach textures
+            camera.attachTexture( ext.COLOR_ATTACHMENT0_WEBGL, this._textureList[ 0 ], 0 );
+            camera.attachTexture( ext.COLOR_ATTACHMENT1_WEBGL, this._textureList[ 1 ], 0 );
+            camera.attachTexture( ext.COLOR_ATTACHMENT2_WEBGL, this._textureList[ 2 ], 0 );
+            camera.attachTexture( ext.COLOR_ATTACHMENT3_WEBGL, this._textureList[ 3 ], 0 );
+            camera.attachTexture( osg.FrameBufferObject.DEPTH_ATTACHMENT, depthTexture, 0 );
```
Kuranes: mmm would have expected API to be more "hidden" from user
`attachTexture('color', this._textureList[ 0 ], 0)`
and 
`attachTexture('depth', depthtexture, 0)`

Kuranes:
I don't get the "flag in Camera to enable drawBuffers exclusively in each camera object needing it"
For me it's like osgjs does it already, seems ok to me

examples/drawbuffers/main.js
jtorresfabra: Yep, but it is useful, in fact I'm using it in my defererred renderer example

examples/drawbuffers/main.js
jtorresfabra: Umm I don't get it, just tried to mimic a usual mrt setting in OSG https://github.com/openscenegraph/OpenSceneGraph/blob/master/examples/osgmultiplerendertargets/osgmultiplerendertargets.cpp#L349

sources/osg/FrameBufferObject.js
jtorresfabra: Ok noted.

sources/osg/FrameBufferObject.js
jtorresfabra: Yep you are right, but I don't get why to compare against the HexValue and not to the alias, which seems more readable. I will change it.

jtorresfabra:
@Kuranes Yeah, you are right, the camera stuff does not make sense. Thanks for the review so far.

jtorresfabra:
Updated to master and fixed @Kuranes comments. I think it is ready again.

cedricpinson:
I implemented mrt on the web2 branch https://github.com/cedricpinson/osgjs/tree/example/webgl2
It probably does not handle everything in a great way, but it could interesting to check if the api in webgl2 match the extension api.
https://github.com/cedricpinson/osgjs/blob/example/webgl2/sources/osg/FrameBufferObject.js

jtorresfabra:
It seems it is practically the same. 

The only thing I see different is that I needed to assure that the attachment.attachment was a color texture. Because it does not work well with depth textures. 

https://github.com/cedricpinson/osgjs/pull/653/files#diff-41cc010b04fecbc9cc51b95425d061b3R324

PR #655, opened 2016-09-12
Switched to gl-matrix npm module

jtorresfabra:


package.json
```diff
@@ -44,7 +45,7 @@
     "Matt Fontaine <tehqin@gmail.com>",
     "Stephane Ginier <stephane.ginier@gmail.com>"
   ],
-  "version": "0.2.7",
+  "version": "0.3.0",
```
cedricpinson: I think grunt release should update package.json

jtorresfabra: oh yes, I did not think about that

jtorresfabra: Fixed

jtorresfabra:
It seems I messed the commits

PR #658, opened 2016-09-14
[D3D-2515] shadow ps must return 0 when unlighted

Kuranes:
When a light is not ligthing a fragment we optimize by not computing
shadow (and avoid any acne shadow on those surface)

But we returned 1.0 which means "no shadow" which is not consistent
shadow receive function. (in fact it is "shadowed")


cedricpinson:
Did you check the rendering of shadow example with this code to validate there is no regression ?

Kuranes:
yes, in "normal" light are unaffected. 
It's only special light that make use of shadow values in spite of not being lighted that is impacted here

PR #665, opened 2016-09-21
Little hack for mac users on firefox enabling them to zoom on models

Utopiad:


sources/osgGA/OrbitManipulatorStandardMouseKeyboardController.js
```diff
@@ -73,6 +73,13 @@ OrbitManipulatorStandardMouseKeyboardController.prototype = {
                 } else {
                     this.setMode( OrbitManipulatorEnums.ROTATE );
                 }
+            }
```
cedricpinson: put the 'else if' on this line

cedricpinson: Could you check in sketchfab to be sure there is no side effect ?

Utopiad: I checked in Sketchfab with Chrome, Firefox, Opera and Safari and only Firefox triggers this elseIf block. :)

PR #667, opened 2016-09-26
Removes deprecated example shadow

cedricpinson:


cedricpinson:
Some others examples to removes/fix
* lod: Is the first example I did to demonstrate the lod capability, but it is shitty, we can get rid of it.
* menu3d : I've got a bunch of errors in the console, no image is shown in the quads ( 403 forbidden ). The example is useful, but it would be nice to fix the errors
* pointcloud/pokerscene: both of them only load scenes with the old method, not using osgDB, so they should be removed and maybe change them for a simple example loading a model with osgDB. Pointcloud scene makes somehow sense, to test if point geometry is renderered correctly. But the way it is loaded should be changed. Also it could make more sense to have something like osggeometry example, to avoid regressions and to test that rendering "normal" geometries work
https://github.com/openscenegraph/OpenSceneGraph/blob/master/examples/osggeometry/osggeometry.cpp BTW the tutorial loadModel should be removed too IMO
cc @jtorresfabra @Kuranes

PR #669, opened 2016-09-29
Refactor vertex compiler

stephomi:
@jtorresfabra @Kuranes 
Important changes concerning the vertex part of Compiler, with possibly breaking changes concerning vertex compiler.

The fragment is now computed before the vertex.
That way, the vertex simply have to see what was the requested varying and call the appropriate functions.

I moved some functions into 2 separates files : CompilerVertex and CompilerFragment.


jtorresfabra:
@stephomi thanks for the heads up. Yeah, it probably break my compilers. 

I have only had a look to the PR but if I understood ok the purpose of this refactor ( from your words ) is to create the right varyings and try to get a cleaner vertex shader, isn't it?. Doesn't it collide somehow with https://github.com/cedricpinson/osgjs/pull/630? Maybe not, I'm guessing. Maybe you could explain a bit more on the motivation of the PR if I'm wrong.

Also as the API changes, can I ask for a dev release before applying this PR @cedricpinson?

Finally, I will wait til the PR is reviewed to test in my end so I can give you some feedback :).  

Thanks in advance.

stephomi:
Yes, the purpose of this PR is to clean the vertex compiler shader.

I removed all the "declareVertexTransform..." functions because they weren't needed.
If you wanted to access a normal/tangent/worldPosition varying in the fragment, you had to change the vertex shader to somehow make it happens.
But the logic behind these declarations was confusing and not flexible.
For example, [here](https://github.com/cedricpinson/osgjs/blob/develop/sources/osgShader/Compiler.js#L1496-L1507), it states that you need to be "shadowed" to declare worldPosition or that you need to be "lighted" to declare the normal.
Another issue is if you need the "view normal" in your vertex shader, it was automatically created as a varying (when in fact, you might only need it for the vertex shader only).

We don't need all this logic (function such as "needTangent", "isShadowed" were simply removed and it works just as good), as we simply need to declare what the fragment needs.
You probably won't have to do much in case it broke some old compilers.

#630 occurs after the shader compiler, so it's safe.

jtorresfabra:
> We don't need all this logic (function such as "needTangent", "isShadowed" were simply removed and it works just as good), as we simply need to declare what the fragment needs.
> You probably won't have to do much in case it broke some old compilers.

Understood, thanks.

> #630 occurs after the shader compiler, so it's safe.

Yeah, I think I did not express myself very well. I know it occurs later
What I meant is that #630 is also aiming to remove not used variyngs, uniforms, etc. So that is addresed also in that PR. 

Anyway thanks for the detailed explanations, they makes sense.

stephomi:
@cedricpinson Also I'm think we could rename some matrices or adding some missing ones to osgjs for consistency : 
- ProjectionMatrix (present)
- ViewMatrix (present)
- ModelViewMatrix (present)
- ModelViewNormalMatrix (present, but under the name of NormalMatrix)
- ModelMatrix (present, but currently named ModelWorldMatrix)
- ModelNormalMatrix (can be useful for triplanar, world space displacement, etc...)

Not very important but the compiler vertex could take advantage of those in simple case (earlyz, depth pass, outline, ...) :
- ViewProjectionMatrix (mostly for optimization)
- ModelViewProjectionMatrix (mostly for optimization)

cedricpinson:
Agree to change matrix name for consistency, check if it breaks some samples. It's possible because of old custom shader that used static matrix name.

PR #675, opened 2016-10-11
Cleans a little bit orbit code

stephomi:


sources/osgGA/OrbitManipulator.js
```diff
@@ -129,16 +129,14 @@ OrbitManipulator.prototype = MACROUTILS.objectInherit( Manipulator.prototype, {
...
         this._scaleMouseMotion = 1.0;
 
         this._inverseMatrix = mat4.create();
 
-        // if we hit the min distance and can't zoom anymore, maybe we still want to move on
-        // with a very low _minDistance, it's like a fps manipulator as long as you don't unzoom
-        this._autoPushTarget = true;
+        // distance at which we start pushing the target (so that we can still zoom)
+        // with a very low _limitZoomIn, it's like a fps manipulator as long as you don't unzoom
+        this._autoPushTarget = 1e-4;
```
jtorresfabra: The API suggest that autoPushTarget is a bool
(https://github.com/cedricpinson/osgjs/blob/1306248f28349a54c30438d1d2c4907433d72e01/sources/osgGA/OrbitManipulator.js#L326).

Also you compare like a bool:   if ( this._autoPushTarget && newValue < this._limitZoomIn )

Could be an error here?

stephomi: Oops, good catch indeed, it's a bool

PR #677, opened 2016-10-13
[D3D-2585] Adds debug visualisation shadow

Kuranes:
Show both target bbox as computed, and light shadow frustum as computed.


sources/osg/Shape.js
```diff
@@ -730,6 +730,145 @@ var createGridGeometry = function ( cx, cy, cz, wx, wy, wz, hx, hy, hz, res1, re
...
+
+
+    g.getPrimitives().push( new DrawElements( PrimitiveSet.LINES, new BufferArray( 'ELEMENT_ARRAY_BUFFER', indexes, 1 ) ) );
+
+    return g;
+};
+// 
+
+/*
+ * allow for bounding boxes dynamically updated client side
```
cedricpinson: Rename and move updateBoundingBoxGeometryWithPerspectiveFrustum to osgUtil/debug.js

Could be named this way:

createFrustrumDebugGeometry: function( projection, dr ) {...}
You can create the debug geometry in the function and create a updateCallback that will update the geometry from the projection matrix

Kuranes: osgUtils/DebugBoundingBox.js ?


 "createFrustrumDebugGeometry"  :  frustum is a boundinbox geometry just "updated" ?
I use  createBoundingBoxGeometry for BoundinBox debug AND frustum debug.

More correct term would be "createLineBoxGeometry" 
that can be updated as a boundingbox or as a frustum ?

Should move all the additions  in "osgUtils/DebugBoundingBox.js" including  the create ?
or just the 2 updates ?

Kuranes:
Heeded the comment :

- moved everything in osgUtils/Debug.js 
- "createBoundingBoxGeometry" named 'createBoxDebugLineGeometry'
- added createFrustrumDebugGeometry: function( projection, dr ) {...}


( the ability to "autoupdate" doesn't work in shadowmap though because matrix are reused/zero'ed elsewhere  )

cedricpinson:
it's osgUtils/debug.js not Debug, upper case first letter is for class.

Because there is createAxisGeometry, createGridGeometry  in shape it's better to have createBoundingBoxGeometry in shape to be consistent.

I dont want to expose updateBoundingBoxGeometry and updateBoundingBoxGeometryWithPerspectiveFrustum in debug.js . If you need to update the debug element without reference, then add an update methodGeometry function:

var updateFrustrum =  function( args ) {
do your stuff
};

var createDebugFrustrumGeometry = function() {
var geom = shape.createBoundingBoxGeometry();
geom.updateGeometry = updateFrustrum;
}

could be a simple function assigned to the debug geometry or a prototype switch. Probably we can start only with a function assigned to the geometry.

Also the naming is not consistent
createBoxDebugLineGeometry is not consistent with createFrustrumDebugGeometry.
could simply:
- createDebugFrustrumGeometry

Kuranes:
ok, will do.

Do I rename Shape.js to shape.js then ?
(maybe I could do a pass on all osgjs file to check consistency)

jtorresfabra:
Hey @Kuranes 

> Do I rename Shape.js to shape.js then ?
> (maybe I could do a pass on all osgjs file to check consistency)

That could break a lot of user code. If you plan to do that please send an advice to the list.

Kuranes:
will rename all "functions files" to lowercase then

Kuranes:
- rebased
- made the Renaming

(tested everything; found things broken but not by this PR (tutorial arealight and loadmodel.))

sources/osgShadow/ShadowMap.js
```diff
@@ -1035,6 +1083,37 @@ ShadowMap.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( S
         //this._shadowReceiveAttribute = undefined;
         this._texture = undefined;
         this._shadowedScene = undefined;
+    },
+
+    setDebug: function ( a ) {
```
cedricpinson: fix the argument name. I guess a is a boolean ?

sources/osgUtil/debug.js
```diff
@@ -0,0 +1,91 @@
...
+
+};
+/*
+ * Given a Camera, create a wireframe representation
+ *  of its view frustum
+ */
+var createDebugFrustrumGeometry = function () {
+
+    var g = Shape.createBoundingBoxGeometry();
+    g.updateGeometry = updateFrustum.bind( g );
```
cedricpinson: go for me

sources/osgUtil/osgUtil.js
```diff
@@ -20,6 +21,7 @@ var WebVR = require( 'osgUtil/WebVR' );
 var osgUtil = {};
 
 osgUtil.Composer = Composer;
+osgUtil.Debug = Debug;
```
cedricpinson: not fixed

Kuranes:
- fixed renaming and osgUtil mixin

Shouldn't https://github.com/cedricpinson/osgjs/blob/master/sources/osgUtil/GizmoGeometry.js expose too it's internal debug geoms ? ( plan, line, quad, circle geometrY )

Kuranes:
- added near/far fix for spotlight shadow bugs

sources/osgUtil/osgUtil.js
```diff
@@ -1,10 +1,12 @@
 'use strict';
 var Composer = require( 'osgUtil/Composer' );
+var Debug = require( 'osgUtil/debug' );
```
cedricpinson: should be var debug =  and not Debug.
First upper letter is for class in the code too

Kuranes:
- renamed Debug to debug in js code.

PR #685, opened 2016-11-10
Adds GLTF support

DavidPeicho:
Adds glTF file loading.
Updates the pbr example to use the glTF loader.


sources/osgDB/Registry.js
```diff
@@ -0,0 +1,31 @@
...
+        if ( !Registry.instance ) {
+            Registry.instance = this;
+            Registry.instance.plugins = new Map();
+        }
+        return Registry.instance;
+    },
+
+    // We register directly a plugin for a extension.
+    addReaderWriter: function ( extension, plugin ) {
+        if ( Registry.instance.plugins.get( name ) !== undefined )
```
jtorresfabra: Shiuld be extension, not name

sources/osgDB/Registry.js
```diff
@@ -0,0 +1,31 @@
...
+            console.log( 'the plugin already exists' );
+        Registry.instance.plugins.set( name, plugin );
```
jtorresfabra: Same as above

sources/osgPlugins/ReaderWriterGLTF.js
```diff
@@ -0,0 +1,1060 @@
...
+            reader.readAsArrayBuffer( file );
+        else if ( fileType === 'text' )
+            reader.readAsText( file );
+        else
+            reader.readAsDataURL( file );
+
+        return defer.promise;
+    },
+
+    requestImageURL: function ( url ) {
```
cedricpinson: It looks very similar to Input.readImageURL, so probably we could use the same code. Could need a bit of refactoring but I think it would be better to have only one method to fetch data type

sources/osgPlugins/ReaderWriterGLTF.js
```diff
@@ -0,0 +1,1060 @@
...
+        this._visitedNodes[ nodeId ] = true;
+
+        return Promise.all( promises );
+    },
+
+    readNodeURL: function ( files ) {
+
+        var self = this;
+
+        this._files = files;
```
cedricpinson: If we call twice the function with the same instance of GLTFLoader promised cached will be not reset.
It's ok if it's too complicated to reuse the class like that, We can use use another class that will be registered in the plugin system and that will instanciate GLTFLoader for each loaded file.

DavidPeicho: I made something for this, just using an init procedure. So it should work by instanciating the loader once.
All the loader's attributes are useless once the loading is over, they are just useful for caching data.

examples/glTF/index.html
```diff
@@ -0,0 +1,45 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <title>Gizmo</title>
```
cedricpinson: typo it's not gizmo example :)

sources/osgPlugins/ReaderWriterGLTF.js
```diff
@@ -0,0 +1,1068 @@
...
+
+    loadGLTFJson: function ( urlOrFiles ) {
+
+        var glTFFileOrUrl = null;
+        if ( typeof ( urlOrFiles ) === 'string' ) {
+
+            var ext = urlOrFiles.split( '.' ).pop();
+            if ( ext !== 'gltf' ) {
+
+                console.warn( 'The given URL does not point toward a valid glTF file' );
```
jtorresfabra: Please use osg.Notify  ( all the file )

sources/osgDB/Registry.js
```diff
@@ -0,0 +1,31 @@
...
+            Registry._instance = Registry;
+            Registry._instance.plugins = new window.Map();
+        }
+        return Registry._instance;
+    },
+
+    // We register directly a plugin for a extension.
+    addReaderWriter: function ( extension, plugin ) {
+        if ( Registry.instance().plugins.get( extension ) !== undefined )
+            console.log( 'the plugin already exists' );
```
jtorresfabra: Use osg.Notify instead of console.log() ( My fault :) )

PR #689, opened 2016-11-24
Feature/multi viewport d3 d 2684

Kuranes:
Multiple viewport makes necessary to have scissor state to prevent drawing outside of viewport
and renderbuffer getting size from texture instead of viewport (renderbuffer and texture attachment must be same size)


sources/osg/Scissor.js
```diff
@@ -0,0 +1,63 @@
...
+
+    attributeType: 'Scissor',
+
+    cloneType: function () {
+        return new Scissor();
+    },
+
+    apply: function ( state ) {
+        var gl = state.getGraphicContext();
+        if ( this._x === -1 ) {
```
cedricpinson: did you mean !== ?

Kuranes: yes true indeed.
(PR still "needs fixing", have to real test it before making it ready, (mutliple cam/viewport) obviously)

Kuranes:
Main question is about the Scissor added to renderStage.

https://github.com/cedricpinson/osgjs/blob/64e4224853cbdaa9a3c90397caebce20e0127c2e/sources/osg/RenderStage.js#L271

It has to be here to be applied between framebuffer binding and the glClear, otherwise it's applied too late (on leaves)
Currently a state in a stateset, could be like viewport.

cedricpinson:
I dont think it's a good place, mixing inter state attribute should be probably avoid. Why not after https://github.com/cedricpinson/osgjs/blob/example/webgl2/sources/osg/RenderStage.js#L270 ?

Kuranes:
It was to be sure it was done prior to any other Fragment operation possible. but I indeed cannot think of any just now.
ok with me, will move it there

cedricpinson:
ok thx

Kuranes:
not sure what that is that case https://github.com/cedricpinson/osgjs/blob/64e4224853cbdaa9a3c90397caebce20e0127c2e/sources/osg/RenderStage.js#L209 
framebuffer without camera ?
Are we supposed to handle it in the scissor code ?
like getting State attribute stack and find a scissor there ?

cedricpinson:
I am not sure this code can be called, check with c++ impl

Kuranes:
mmm doesn't seem possible in osg https://github.com/openscenegraph/OpenSceneGraph/blob/master/src/osgUtil/RenderStage.cpp#L1164

Albeit I think pbo copy or blitbuffer is the use case that could fit in (but we don't do that until webgl2)

PR #696, opened 2016-12-13
[#D3D-270] Fixes shadow shader non uniform flow derivatives usage

Kuranes:
Removed using derivating inside a branch as it makes for undefined behavior
So the early return in the shadowmaping code in now done after
derivatives computations
see http://teknicool.tumblr.com/post/77263472964/glsl-dynamic-branching-and-texture-samplers

### DIFF
--- a/sources/osg/WebGLCaps.js
+++ b/sources/osg/WebGLCaps.js
@@ -190,21 +190,6 @@ WebGLCaps.prototype = {
     // inevitable bugs per platform (browser/OS/GPU)
     initBugDB: function () {
 
-        var p = this._webGLPlatforms;
-        var ext = this._webGLParameters;
-
-        // derivatives gives strange results on Shadow Shaders
-        if ( p.Apple ) {
-
-            if ( !ext.UNMASKED_VENDOR_WEBGL || ext.UNMASKED_VENDOR_WEBGL.indexOf( 'Intel' ) !== -1 ) {
-                // bug is on INTEL GPU on APPLE
-                // we disable the ext on Apple if we cannot get GPU info
-                this._bugsDB[ 'OES_standard_derivatives' ] = true;
-
-            }
-
-        }
-
     },
     initPlatformSupport: function () {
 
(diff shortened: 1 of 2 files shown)


sources/osg/WebGLCaps.js
```diff
@@ -190,20 +190,21 @@ WebGLCaps.prototype = {
     // inevitable bugs per platform (browser/OS/GPU)
     initBugDB: function () {
 
-        var p = this._webGLPlatforms;
-        var ext = this._webGLParameters;
-
-        // derivatives gives strange results on Shadow Shaders
-        if ( p.Apple ) {
+        // example of blocking gpu extension: apple intel derivatives
```
cedricpinson: remove all the code instead of commenting

sources/osgShadow/shaders/shadowsReceiveMain.glsl
```diff
@@ -1,30 +1,65 @@
...
     if (!lighted)
-        return 0.;
+    {
+        shadow = 0.0;
+        earlyOut = true;
+    }
 
     if (depthRange.x == depthRange.y)
-        return 1.;
+      earlyOut = true;
```
cedricpinson: indentation is wrong, plz pass the file to a glsl mode in emacs or something similar, also check the how bracket are in other files like https://github.com/cedricpinson/osgjs/blob/9c9b8a7fff32bcc9241ac5db3e6dabecc02d576c/sources/osgShader/node/lights.glsl use the same coding style plz

Kuranes:
- removed comment
- reformatted glsl

sources/osgShadow/shaders/shadowsReceiveMain.glsl
Kuranes: light.glsl file is not very consistent 
```c++
func ()
{
}
```
and 
```c++
if(){
}
```
?

cedricpinson:
missing braces like the previous file I linked

Kuranes:
> like the previous file I linked
those are one line conditional, so braces too for one line conditional then ?

Please note that light.glsl file is not in itself very consistent, and no osgjs glsl code style exists, and skfb code style says "like js" which isn't consistent with current code.

Light.glsl has
```cpp
func ()
{
}
```
and
```cpp
if(){
}
```
which is the good one ?

cedricpinson:
Like not in light.glsl, like in the link I previously linked

PR #698, opened 2016-12-16
[#D3D-2731] Uses Mesh Normal for shadow bias

Kuranes:
ended up using normal for normal map to compute slope for bias, now uses specifically
Normal from mesh, and use it to do exploding offset normal shadow acne
mitigation


cedricpinson:
dead end ?

Kuranes:
it works, but need time to test on lots of models as it's rather touchy change.

sources/osgShader/node/lights.glsl
```diff
@@ -76,65 +71,64 @@ vec3 computeSpotLightShading(
     // compute attenuation
     float attenuation = getLightAttenuation(dist, lightAttenuation);
     if (attenuation != 0.0)
+    {
```
cedricpinson: coding style it should be if (...) { on the same line

cedricpinson:
It would be good to first have a commit that remove all old shadow method that are not used. I am talking about esm ....

Kuranes:
I started in another PR for that after this one, really easier for me that way.
( other one ia quite big, with  the light glsl style fix, conversion to Example template, etc. )

the clean PR raise two three question too, rhat might take some 'talkloop' removing other algo easy, but not sure waht you want to keep on the pcf algos (pcf, fakepcf, rotated, poisson, max number of taps?)

Kuranes:
moved over https://github.com/cedricpinson/osgjs/pull/719 with everything bleached

PR #702, opened 2017-01-02
Improves pan vs zoom mobile manipulator

stephomi:


sources/osgGA/OrbitManipulatorHammerController.js
```diff
@@ -123,19 +124,31 @@ OrbitManipulatorHammerController.prototype = {
             Notify.debug( 'zoom end, ' + dragCB( event ) );
         };
 
+        var smoothstep = function ( a, b, x ) {
```
cedricpinson: It would make sense to add this function in our glMatrix namespace and probably we will need to move later osgAnimation ease in osg core glMatrix, it's more math than animation actually

PR #707, opened 2017-01-21
Adds SSAO example based on Scalable Ambient Obscurance

DavidPeicho:
Implements, as an example, the SSAO method described in [Scalable Ambient Obscurance](http://graphics.cs.williams.edu/papers/SAOHPG12/McGuire12SAO.pdf )

### DIFF
--- a/examples/ssao/main.js
+++ b/examples/ssao/main.js
@@ -1,0 +1,701 @@
+( function () {
+    'use strict';
+
+    var ExampleOSGJS = window.ExampleOSGJS;
+
+    var OSG = window.OSG;
+    var osg = OSG.osg;
+    var osgViewer = OSG.osgViewer;
+    var osgShader = OSG.osgShader;
+    var osgUtil = OSG.osgUtil;
+    var Texture = osg.Texture;
+
+    var $ = window.$;
+    var P = window.P;
+
+    // Generates MAX_MIP_LEVEL depth mipmaps.
+    // Do not forget to change the code in ssaoFragment.glsl
+    // because the mipmaps are passed manually using texture units
+    var MAX_MIP_LEVEL = 5;
+
+    var shaderProcessor = new osgShader.ShaderProcessor();
+
+    // Loads a glTF model and update
+    // the current displayed scene in the Example
+    var loadFiles = function ( example, files, sceneName ) {
+
+        var promise = OSG.osgDB.Registry.instance().getReaderWriterForExtension( 'gltf' )
+            .readNodeURL( files );
+
+        promise.then( function ( scene ) {
+
+            if ( !scene ) {
+
+                example._config.scene = 'box';
+                example.updateScene();
+
+                return;
+
+            }
+
+            var mt = new osg.MatrixTransform();
+
+            var mtr = new osg.MatrixTransform();
+            osg.mat4.fromRotation( mtr.getMatrix(), Math.PI / 2, [ 1, 0, 0 ] );
+
+            mtr.addChild( scene );
+            mt.addChild( mtr );
+
+            example.addScene( sceneName, mt );
+        } );
+
+    };
+
+    // inherits for the ExampleOSGJS prototype
+    var Example = function () {
+
+        ExampleOSGJS.call( this );
+
+        this._config = {
+            scale: 1.0,
+            radius: 0.3,
+            bias: 0.01,
+            intensity: 0.8,
+            crispness: 0.8,
+            scene: 'box'
+        };
+
+        // The two following attributes are use for sliders
+        // normalization according to the scene bounding sphere
+        this._baseSlidersBounds = {
+            radius: [ 0.005, 0.25 ],
+            bias: [ 0.01, 0.08 ],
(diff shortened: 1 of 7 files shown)


examples/ssao/main.js
```diff
@@ -0,0 +1,701 @@
...
+    var loadFiles = function ( example, files, sceneName ) {
+
+        var promise = OSG.osgDB.Registry.instance().getReaderWriterForExtension( 'gltf' )
+            .readNodeURL( files );
+
+        promise.then( function ( scene ) {
+
+            if ( !scene ) {
+
+                example._config.scene = 'box';
```
cedricpinson: Instead of using only a box when we dont have a model we could create a scene like this http://marcinignac.com/experiments/ssao/v01/index.html programatically

examples/ssao/main.js
```diff
@@ -0,0 +1,701 @@
...
+    // because the mipmaps are passed manually using texture units
+    var MAX_MIP_LEVEL = 5;
+
+    var shaderProcessor = new osgShader.ShaderProcessor();
+
+    // Loads a glTF model and update
+    // the current displayed scene in the Example
+    var loadFiles = function ( example, files, sceneName ) {
+
+        var promise = OSG.osgDB.Registry.instance().getReaderWriterForExtension( 'gltf' )
```
jtorresfabra: Hey let me chime in. The normal usage for loading nodes should be 
`osgDB.readNodeURL(` 'nodeURL' );`

The reader parse calls the registry for you: 

https://github.com/cedricpinson/osgjs/blob/master/sources/osgDB/readerParser.js#L34

examples/ssao/main.js
DavidPeicho: Sure, I will fix it soon.

examples/ssao/main.js
```diff
@@ -0,0 +1,701 @@
...
+            var self = this;
+
+            this.initDatGUI();
+            this.createViewer();
+            self.createScene();
+
+            this.readShaders().then( function () {
+
+                // Loads the glTF Sponza scene by default if exists
+                loadFiles( self, '../media/models/sponza.gltf', 'sponza' );
```
stephomi: doesn't exist :)

DavidPeicho: Yes, the media folder is read-only now :)

cedricpinson: It would be better to use the same system like for PBR, drag and drop some predefined url

examples/ssao/shaders/ssaoFragment.glsl
```diff
@@ -0,0 +1,282 @@
...
+ * The HIGH_QUALITY macro allows to smooth the AO
+ * and spread darken corners.
+ */
+float fallOffMethod0(float vv, float vn, vec3 normal) {
+
+    #   if HIGH_QUALITY
+
+        float invRadius2 = 1.0 / uRadius2;
+        float f = max(1.0 - vv * invRadius2, 0.0);
+        return f * max((vn - uBias) * inversesqrt(EPSILON + vv), 0.0);
```
stephomi: After digging seriously, this line is not scene scale independent at all.
vv and vn *both* depends of the scene bounds (camera view space).
uBias and EPSILON should be dependent somehow of the scale bounding box to work consistently.

For uBias range [0.0, radius * 0.01] are good (0.001*radius is a good starting point).
For EPSILON, I'd remove it completely imho for now (vv at 0 shouldn't happen)

DavidPeicho: So uBias is depends on the **world radius**?
Epsilon is effectively not really mandatory, but it prevents division by zero.

stephomi: yes because :

```
vec3 v = occludingPoint - cameraSpacePosition;
float vv = dot(v, v);
float vn = dot(v, normal);
```

v is a view space vector so both vv and vn are dependent of the scene size (near/far).
An epsilon of 0.001 for a small model breaks the ssao easily for example.

examples/ssao/main.js
DavidPeicho: For now, readNodeURL uses:
` var extension = url.substr( url.lastIndexOf( '.' ) + 1 );`
But the glTF Reader/Writer can handle files. We should either add a new method, or modify this one (but modifying it without changing its name does not make sense)

jtorresfabra: I see. Is there a good reason for breaking the API? I mean, that's the way OSG works so, why don't you call the readNodeURL sequentially for each file?

DavidPeicho: We needed a way to load glTF files from a drag'n'drop. Apparently, it is possbile to retrieve the URL from the File object that has been dropped. I do not have a lot of time at this moment, but I will take a look at this as soon as I can.

examples/ssao/main.js
DavidPeicho: I added something similar, creating random spheres and rotating the whole.

DavidPeicho:
What is the status of this? Should I fix something?

stephomi:
Good for me

PR #708, opened 2017-01-21
Adds support for ANGLE_InstancedArrays extension

jtorresfabra:
Adds ANGLE_InstancedArrays extension extension support.
Comments:
- The way that the vertex attribute divisor has been integrate differs form the one in OSG. In OSG it is integrated as a StateSet so to enable divisor teh code looks like:
` osg::ref_ptr<osg::Program> program = createProgram(); // creates a glsl program. `
    `program->addBindAttribLocation( "vertex", 0 ); // vertex is an attribute`
   `program->addBindAttribLocation( "instancesPosition", 1 ); // should use VertexAttribDivisor `

   ` // Set attribute divisor for instances Position`  
`ss->setAttributeAndModes( new osg::VertexAttribDivisor( 1, 1 ),osg::StateAttribute::ON); // Note that you need to know the attrib location to apply the divisor`

In osgjs we don't know the attrib location before the generateDrawCommand occurs. So I decided to add it to the buffer. So the code looks like:

https://github.com/cedricpinson/osgjs/pull/708/files?diff=unified#diff-467cd15ceb04dca8d459596984d71903R135

Otherwise we can still have an StateAttribute, but it would need some hacks. We can discuss it further if my solution does not make sense.

- Added methods to support user defined computeBoundingBox/computeBoundingSphere callbacks

- The example is a bit naive, I could improve it later. For the first review I think it is enough.


cedricpinson:
look here I think it's possible to set a location https://github.com/cedricpinson/osgjs/blob/master/sources/osg/Program.js#L216

jtorresfabra:
Umm not sure, here you don't know the name of the attribute. It could be whatever the user want.( https://github.com/cedricpinson/osgjs/pull/708/commits/26594759b2f75af8db3bd5ac1a4d5fe1d5e7df0a#diff-467cd15ceb04dca8d459596984d71903R137 ).

sources/osg/DrawArraysInstanced.js
```diff
@@ -0,0 +1,39 @@
...
+var DrawArrays = require( 'osg/DrawArrays' );
+var MACROUTILS = require( 'osg/Utils' );
+
+/**
+ * DrawElements manage rendering of indexed primitives
+ * @class DrawElements
+ */
+var DrawArraysInstanced = function ( mode, first, count, numInstances ) {
+    DrawArrays.call( this, mode, first, count );
+    this.numInstances = numInstances;
```
cedricpinson: coding styles, member variable should be _

cedricpinson:
it's possible this PR needs a rebase, also we should improve the coding style to use _ for internal class member

jtorresfabra:
Ok. 
Should I update also member variables in https://github.com/cedricpinson/osgjs/blob/master/sources/osg/DrawArrays.js and
https://github.com/cedricpinson/osgjs/blob/master/sources/osg/DrawElements.js ?

cedricpinson:
yes but again be careful about the old reader/parser

jtorresfabra:
Fixed, ready for review. Also be careful because I'm not sure the change of the name of the private variables could affect skfb.

sources/osg/DrawArraysInstanced.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
+var WebGLCaps = require( 'osg/WebGLCaps' );
+var DrawArrays = require( 'osg/DrawArrays' );
+var MACROUTILS = require( 'osg/Utils' );
+var notify = require( 'osg/notify' );
+
+/**
+ * DrawElements manage rendering of indexed primitives
```
stephomi: DrawArrays

sources/osg/DrawArraysInstanced.js
```diff
@@ -0,0 +1,41 @@
...
+ * @class DrawElements
```
stephomi: same

sources/osg/DrawArraysInstanced.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+var DrawArraysInstanced = function ( mode, first, count, numInstances ) {
+    DrawArrays.call( this, mode, first, count );
+    this._numInstances = numInstances;
+    this._extension = undefined;
+};
+
+/** @lends DrawElements.prototype */
```
stephomi: same

jtorresfabra: ahh copy/paste. Shame on me...

jtorresfabra: Fixed

sources/osg/DrawArraysInstanced.js
```diff
@@ -0,0 +1,41 @@
...
+DrawArraysInstanced.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( DrawArrays.prototype, {
+
+    draw: function ( state ) {
+        if ( this._count === 0 )
+            return;
+        var gl = state.getGraphicContext();
+        if ( this._extension === undefined )
+            this._extension = WebGLCaps.instance( gl ).getWebGLExtension( 'ANGLE_instanced_arrays' );
+        if ( !this._extension ) notify.error( 'Your browser does not support instanced arrays extension' );
+        this._extension.drawArraysInstancedANGLE( this._mode, this._first, this._count, this._numInstances );
```
stephomi: Not robust, it will crash.
Should be:

```
if ( !this._extension ){
    this._extension = WebGLCaps.instance( gl ).getWebGLExtension( 'ANGLE_instanced_arrays' );
    if ( !this._extension ){
        notify.error( 'Your browser does not support instanced arrays extension' );
        return;
    }
}
this._extension.drawArraysInstancedANGLE( this._mode, this._first, this._count, this._numInstances );
```

jtorresfabra: good catch

jtorresfabra: Fixed

stephomi:
@cedricpinson 
For the callbacking thingy, not really convinced it should be in core osgjs.

Also, there's missing checks for callback Sphere/Box for :
- MorphGeometry, RigGeometry
- Light, LightSource, Lod, Transform

jtorresfabra:
Upated in https://github.com/cedricpinson/osgjs/pull/968

PR #709, opened 2017-01-25
Fixed scope issue in mat4.getLookAt method.

jtorresfabra:


stephomi:
Nice catch! I guess this function was never called...

Not really related to the fix but... :

The function ``transformMat4R`` is not returning out, we should be consistent with glMatrix (also glmatrix avoids multiple lookup in same array by declaring once the variable on the top http://glmatrix.net/docs/vec3.js.html#line540, probably not a big deal but consistency again...)
Not sure if the function name is clear enough to be exposed like that too (transformMat4Rotate?)

v1 could be replaced by ``vec3.ZERO`` to remove ambiguity.
Can also be replaced by mat4.getTranslation(out, eye) too. (the /out[15] is not necessary I guess as matrix is expected to be a view matrix).

jtorresfabra:
@stephomi you are right indeed. Closing this PR as #710 fixes it with your comments. 
Thank you!

PR #711, opened 2017-02-01
WIP: Backwards compatibility for WebGL1 extensions

jtorresfabra:
Not for merge. Just to see the strategy for backwards compatibility.
Also some changes in order to compile without errors.


sources/osg/WebGLCaps.js
```diff
@@ -363,6 +383,7 @@ WebGLCaps.prototype = {
             }
 
             ext[ sup ] = gl.getExtension( sup );
+            this.getAndApplyExtension( gl, sup );
```
cedricpinson: Does it really works for all extension ? If signature between extension and core of webgl2 it will not work

cedricpinson:
what happens if on a page you have a webgl1 context and a webgl2, how webglcaps support that ?

jtorresfabra:
Good question. I'm going to implement test example with this configuration to assure all works as expected. 

BTW: draw buffers also works with minor changes.

jtorresfabra:
@cedricpinson just modified the multipleviewers example to have each viewer with a different webGL version. It fails, but it seems it is related to Uniforms, not to webGLCaps. that seems to work. I'll keep digging.

cedricpinson:
Maybe we can also say that if you have webgl2 you just have to use two same webgl context. Not sure mixing is a real usecase

jtorresfabra:
Ok I found the error:
https://github.com/cedricpinson/osgjs/blob/master/sources/osg/Uniform.js#L57

We cannot cache when different type of context comes in. Without cache it works with different contexts.

sources/osg/Uniform.js
```diff
@@ -54,15 +54,25 @@ Uniform.prototype = {
 
     apply: function UniformApply( gl, location ) {
 
-        if ( !this._cache )
+        if ( !this._cache || this._cacheDirty === true ) {
             this._cache = gl[ this._glCall ];
-
+            this._dirtyCache = false;
```
stephomi: this._cacheDirty = false;

Also why exposing gl call signature? I think it's supposed to be kept private

jtorresfabra: > this._cacheDirty = false;
arg, well spotted!
> Also why exposing gl call signature? I think it's supposed to be kept private
I guess I did it for convenience so each time glSignature changes it dirties the cache.
https://github.com/jtorresfabra/osgjs/blob/653a94d5a8f160be5fa6aa2756ccc543325ec33b/sources/osg/Uniform.js#L169
 But I can keep it private. I'll fix it.

jtorresfabra:
Closing this as https://github.com/cedricpinson/osgjs/pull/765 has been merged

jtorresfabra:
Reopen the PR just to review it.

jtorresfabra:
Closing it, does not make sense anymore.

PR #713, opened 2017-02-07
Removes Geometry Proxy and setUpdateCallback function

jtorresfabra:


stephomi:
Maybe now we can simply keep `_primitives` and `_attributes`.

jtorresfabra:
Fixed.

stephomi:
+1

cedricpinson:
@jtorresfabra needs to check the old deprecated parse scenegraph if we dont break something https://github.com/cedricpinson/osgjs/blob/performance-test/sources/osgDB/readerParser.js#L186
I wanted to remove the code but I believe we still have old files

jtorresfabra:
oups, missed these. Fixed now.

sources/osgUtil/IntersectionVisitor.js
```diff
@@ -156,7 +156,7 @@ IntersectionVisitor.prototype = MACROUTILS.objectInherit( NodeVisitor.prototype,
 
     applyNode: function ( node ) {
         if ( !this.enter( node ) ) return;
-        if ( node.primitives ) {
+        if ( node.getPrimitiveSetList ) {
```
cedricpinson: it's probably better to use 'instanceof Geometry' it makes more sense to read

cedricpinson: my bad it's faster to check property https://jsperf.com/instanceof-performance/25

PR #714, opened 2017-02-08
Adds error logging to video example and change video URL

jtorresfabra:


stephomi:
The commit message is too long, missing EOL after "video example." :).

jtorresfabra:
I have seen much longer commit messages, but it is ok. Fixed. :+1:

PR #715, opened 2017-02-08
Updated/Fixed tutorials

jtorresfabra:


cedricpinson:
Actually I wanted to remove ParameterVisitor, it's only used in those tutorials now not sure if we could change it for dat.gui ?

jtorresfabra:
> Actually I wanted to remove ParameterVisitor, it's only used in those tutorials now not sure if we could change it for dat.gui. Could you check if it's easy or not ?

Yep, I'm doing it.

> I also wanted to move example more like how they do in threejs, it's lighter example but at least easier to test them.

Yes, and I hesitate about the point on having separate example/tutorial folders. Also a couple of tutorials could be removed ( like https://github.com/cedricpinson/osgjs/blob/master/tutorials/06_Advanced_Custom_Shader/main.js )

jtorresfabra:
Fixed.

@cedricpinson I will be merging tutorials in the examples folder in a different PR as we discussed.

PR #716, opened 2017-02-09
Removes inline webpack loaders shortcuts

gvergnaud:
Webpack 2 stopped supporting the '-loader' shortcut. When trying to require a source package of osgjs in a project using wepback 2, the build currently fails. This PR fixes that.

### DIFF
--- a/sources/osgNameSpace.js
+++ b/sources/osgNameSpace.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var pkg = require( 'json!../package.json' );
+var pkg = require( 'json-loader!../package.json' );
 
 module.exports = {
     name: pkg.name,


cedricpinson:
Does it mean it will not work anymore with webpack 1 ? in this case does it mean we should update package.json to use webpack2 as developer dependency ?

gvergnaud:
It will continue to work with Webpack 1. Being able to use a loader by omitting the `-loader` in its package name was just a feature to avoid writing to much text ahead of the file name. They removed it in Webpack 2 in favor of a more 'programmatic' approach.

More info here :
https://webpack.js.org/guides/migrating/#automatic-loader-module-name-extension-removed

PR #719, opened 2017-02-13
Feature/shadow-normal-clean_D3D-2731

Kuranes:
- Remove all but pcf shadow algo
- format lights.glsl
- move shadowmap to Example
- Fixes Normal Bias
- Adds Normal exploding offset


Kuranes:
moved to https://github.com/cedricpinson/osgjs/pull/720

PR #720, opened 2017-02-13
Feature/shadow normal clean d3 d 2731

Kuranes:
Remove all but pcf shadow algo
format lights.glsl
move shadowmap to Example
Fixes Normal Bias
Adds Normal exploding offset


examples/shadowmap/main.js
```diff
@@ -704,13 +626,46 @@
             this.updateShadowFormat();
             this.updateShadowMapSize();
 
-            this.updateDebugRTT();
+            var l, numLights;
+            numLights = ~~this._config.lightNum;
+            l = numLights;
+            while ( l-- ) {
```
cedricpinson: please use regular loop like in all example, I know it's faster but it's not really important here. Better to have consistent over performance for this kind of code

cedricpinson:
Very cool, if you can just fix some comment
+425 −1,638 👍

Kuranes:
- removed all while
- rebased

examples/templates/Example.js
```diff
@@ -95,7 +96,17 @@
             var keys = Object.keys( queryDict );
             for ( var i = 0; i < keys.length; i++ ) {
                 var property = keys[ i ];
-                this._config[ property ] = queryDict[ property ];
+                try {
```
cedricpinson: could you explain this please

examples/templates/Example.js
```diff
@@ -206,6 +217,32 @@
             if ( optionalArgs )
                 osg.extend( optionsDebug, optionalArgs );
 
+            if ( !this._programRTT ) {
```
cedricpinson: could you also explain why adding this ?

examples/templates/Example.js
Kuranes: The idea is to get typed result from url, here in shadow in order to get numbers in _config directly from url (many shadow url parameter from shadows are numbers)
( content of _config has is used by dat.gui directly for slider/input etc)

examples/templates/Example.js
Kuranes: it makes rendering of the debug quad RTT independent of the scene tree, so that it's drag and drop code, without needing user care.
(without that code I got black quads)

cedricpinson: but it used to work without it before ?

Kuranes: shadow wasn't using example, now it does, but shadow has its own version of debug RTT which was using the shader program.

cedricpinson: It's because your texture are 0 so textureNull, so not used in the compiler to build the program. I am not against having a specific program for that, but I dont want you workaround stuff if you dont understand why it does not work.

cedricpinson:
screen shot 2017-02-17 at 15 20 14
could you clean the message when there is no option in the url, it says cannot parse url option, but useless

examples/shadowmap/main.js
```diff
@@ -704,21 +626,54 @@
             this.updateShadowFormat();
             this.updateShadowMapSize();
 
-            this.updateDebugRTT();
+            var i, l, numLights;
+            numLights = ~~this._config.lightNum;
```
cedricpinson: not a fan of this, If you fixed the content with the url parser in Example I would prefer we remove this kind of tricks

examples/templates/Example.js
```diff
@@ -95,7 +96,17 @@
...
+                try {
+
+                    var n = JSON.parse( queryDict[ property ] );
+                    if ( !isNaN( parseFloat( n ) ) && isFinite( n ) ) {
+                        n = parseFloat( n );
+                    }
+                    this._config[ property ] = n;
+
+                } catch ( e ) {
+                    console.log( 'cannot parse url option: ' + property );
```
cedricpinson: dont use console.log use osg.notify

cedricpinson:
there was a conflict with the fix about earlyz, I recheck and made a few comments. Would be great if you could fix them at the same time you resolve conflicts

examples/templates/Example.js
Kuranes: "texture are 0 so textureNull"
Not sure to understand, code does
"stateSet.setTextureAttributeAndModes( 0, texture );"
So no null nor texturenull here ?

It does set the texture (otherwise special shader wouldn't work?).
Withtout special shader, we get black quad as shader compiler outputs "shaders with shadows"
even with the debugNodeRTT isn't under the shadowed scene node, the one with shadow receiver attributes.

I dont want you workaround stuff if you dont understand why it does not work.
It's code from old shadow demo from master, already reviewed once, that's why I didn't think twice about spending time to double question and digg it.
( I think we even did had a discussion about that exact thing, shader compiler getting attributes default of material/light/etc and doing a "material" shader where it shouldn't as debugNode is a sibling of the node containing shadowedscene )
I can dig more and make a repro case easier to debug ?

Kuranes:
@cedricpinson rebased and cleaned last remains.

slightly OT: I confirm I need to make an issue on the leftovers of sibling nodes and lights/material vs shader compiler

PR #721, opened 2017-02-15
Adds mat3 normal matrix support

stephomi:
Adds modelNormalMatrix and switch *normalMatrix to mat3.
With hindsight I'm not 100%, sure it's better or not. @cedricpinson 

Otherwise it works fine, except examples needs to be updated and normalFromMat4 return value should be checked too.


sources/osg/State.js
```diff
@@ -285,111 +287,16 @@ State.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( Objec
...
-                    mat4.transpose( normal, normal );
-
-                    mu.setMatrix4( normal );
+                    mat3.normalFromMat4( normal, matrix );
+                    mu.setMatrix3( normal );
                     mu.apply( gc, mul );
                 }
-            }
-
-            this._modelViewMatrix = matrix;
```
cedricpinson: Why ? it looks like the check on line https://github.com/cedricpinson/osgjs/pull/721/files#diff-fe8561621c75d077d3789cdfa9f0088aL254 will never be valid

cedricpinson:
I dont see any reason why not using mat3. About performance this PR increase time in gl probably because of the check I put in the comment. It's possible to check it with performance example

sources/osg/State.js
stephomi: Hmm I don't know, maybe it's in the case applyModelViewMatrix is called several time in a row?

PR #722, opened 2017-02-17
Adds drag'n drop and ReaderWriterZIP

jtorresfabra:
Also small refactor of ReaderWriterGLTF


DavidPeicho:
Nice refactoring!
I should update the others examples with the new loading code whenever it is merged.

sources/osgPlugins/ReaderWriterGLTF.js
```diff
@@ -124,218 +121,29 @@ GLTFLoader.prototype = {
...
-
-            } );
-            this._cachepromise[ uri ] = promise;
-
-            return promise;
-
-        }
-
-        var promiseFile = this.requestFileFromReader( urlOrFile, fileType );
+        var promiseFile = this.requestFileFromReader( uri, fileType );
```
cedricpinson: this method does not exist

cedricpinson:
I thought about something like so I think it's good. We still need to discuss about some details.
About the zip plugin, I dont want the deps to be in the osgjs build, also be sure that if this code path is used, the user will be warned he has not the deps loaded in his page

sources/osgPlugins/ReaderWriterGLTF.js
jtorresfabra: Oups, bug. But nope it should use the requestFile method ( https://github.com/cedricpinson/osgjs/pull/722/files#diff-29f7b843057ce77ec71c63808ab0c91bR57 ). 

The fileHelper is to read a list of files, like you have when drag'n dropping.

jtorresfabra:
@cedricpinson Just to be sure, do you want JSZip in the vendors folder or just importing it as script tag pointing to a url in the index.html like https://github.com/cedricpinson/osgjs/blob/master/examples/pbr/index.html#L36

examples/dragndrop/index.html
```diff
@@ -0,0 +1,45 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <title>Gizmo</title>
```
cedricpinson: example name :)

PR #729, opened 2017-02-21
Cleans a few stuffs in polytope code

stephomi:
Reduce vector allocations.
Fixes the buggy use case when there is no matrix transform in the picking path (uninitialized _iPolytope/_iReferencePlane)
Adds the cullingActive check for node intersection (like the segmentIntersector)


cedricpinson:
could you remove yarn.lock plz

PR #732, opened 2017-02-23
Adds Easier Shader Debug

Kuranes:
The usueal pink shader is still activated, but real but failed shader is
inside a  _DEBUG so that you can debug it in  "shader editor"

### DIFF
--- a/sources/osg/Program.js
+++ b/sources/osg/Program.js
@@ -246,25 +246,21 @@ Program.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( GLO
             }
 
             if ( !compileClean ) {
+
                 // Any error, Any
                 // Pink must die.
-                if ( !Program.prototype._failSafeCache ) {
 
-                    var program = gl.createProgram();
-                    this._vertex.failSafe( gl );
-                    this._fragment.failSafe( gl );
+                var program = gl.createProgram();
+                this._vertex.failSafe( gl, this._vertex.getText() );
+                this._fragment.failSafe( gl, this._fragment.getText() );
 
-                    gl.attachShader( program, this._vertex.shader );
-                    gl.attachShader( program, this._fragment.shader );
-                    gl.linkProgram( program );
-                    gl.validateProgram( program );
+                gl.attachShader( program, this._vertex.shader );
+                gl.attachShader( program, this._fragment.shader );
+                gl.linkProgram( program );
+                gl.validateProgram( program );
 
-                    // cache to compile and allocate only once
-                    // not polluting the inspector
-                    Program.prototype._failSafeCache = program;
-                }
                 Notify.warn( 'FailSafe shader Activated ' );
-                this._program = this._failSafeCache;
+                this._program = program;
             }
 
             this._uniformsCache = new CustomMap();
--- a/sources/osg/Shader.js
+++ b/sources/osg/Shader.js
@@ -23,12 +23,10 @@ Shader.VERTEX_SHADER = 0x8B31;
 Shader.FRAGMENT_SHADER = 0x8B30;
 
 // Debug Pink shader for when shader fails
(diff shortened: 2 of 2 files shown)


cedricpinson:
I dont understand how it could work well. 
If your _DEBUG program declare some uniforms, let's say materialDiffuse. At compile time because the _DEBUG will not use those uniforms, the function cacheUniformList will not list them. So when you enable the _DEBUG, materialDiffuse will not apply because not listed in the program to apply.
It's the same thing with cacheAttributeList.

It's possible it could work, but I prefer we improve first the way we report shader error

Kuranes:
pr title not clear sorry:

shader debug => shader syntax error debug.

When a shader compilation fails, it's always *syntax* problem, not a *render* problem.

Here, the goal is not to debug visually the render, so no interest at all in attributes and uniform.

The goal here with this _minimal change_ is to get the shader in shader editor chrome extension so we can fix syntax errors fast and easy *on the spot*

you can spot and fix  the *_syntax_* painlessly. 
(shader editor does compile shader just for syntax check and points errors with colors in its own context...)

cedricpinson:
In this case why not just fixing that in the report log of the shader, there is some code to handle this

Kuranes:
much much easier in shader editor:
- you have the whole shader
- shader syntax code coloring
- you can fix edit directly on the whole shader
- much mire readable
- no need to tweak log level.
etc

adding the same level of feature means code bloat in shader.js code

cedricpinson:
I dont feel that it's the general use case, Why not instead disabling failsafe instead, would that be simpler and do the same thing ?

Kuranes:
When shader compilation fails, the program is not linked and thus not
showing in the shader editor.

imho the change is really very benign here, hard to do a simpler change, no
side effect, so the cost/benefits, even if considereing "it's not the
general case" is quite positive.
( Myself I would even remove all the shader error logging code in favor of
that, as I won't ever use it again if this PR is merged.)

cedricpinson:
shader editor does not work on all platform, the log must stay and be improved. The point I dont like is that there are more effort to hack on top of thing, instead of just improving what needs to be improved. It's probably as easy to copy/paste the shader from the log into shader editor. Not need to create multiple failsafe shader with original shader inside.

Kuranes:
Ok, currently we don't log no more the whole shader when it fails.
( shader error log is messy since changes in logs level and fold )

OT:  I would recommand using that shader editor I do use
https://github.com/Kuranes/ShaderEditorExtension/tree/feature/upgrade that
works even on mobile, including ios ( using vorlonjs.)( Still some case
where attributes/uniform not properly rebinded on edit, but shader editor
syntax debug always works for me.

cedricpinson:
It does not work here, I click on one program everything disappear

Kuranes:
yes, quite possible shader *render* debugging may not work on your os but
shader syntax error debugging  works :)

cedricpinson:
Ho yes I am sure it helps for that, my point is that log must be clean and fixed, you will probably able to copy/paste a shader from the log.
I will be open to add extra code to improve that. We should also talk with other devs to know their workflow and what's their pain.

stephomi:
I agree with @Kuranes on this point.
When writting shader the log system in console log is a pain.
Copy-paste the error in a glsl linter (http://shdr.bkcore.com/, chrome shader extension)  works but it's always a pain to copy long string from the chrome console output.

For me, the worst case is when I don't know which shader failed, when you have easily 10-20 shaders it's not funny (I usually check the shader_name define or the code itself).

Ideally I'd like to have the failed shader in the shader editor extension (with its original name, and ideally append with "Failed" or something like that).

cedricpinson:
ok then if you have both the same issue, @Kuranes use @stephomi inputs to update your PR

jtorresfabra:
Just to add my 2 cents

ShaderEditor is not working for me from long time ago, so I didn't rely in it at all. I have had hard times to get shaders form the chrome debug tools, as with long text strings it's pretty awful as @stephomi said.
It keeps creating shaders objects, attached a screenshot.
image

I ended up using webGL Insight chrome extension to copy paste shaders.

I get the point of @cedricpinson and also I think it would be nice to have something in case you are not using chrome and chrome extensions. Said that I think adding this PR won't hurt, and sure it will help to improve the workflow for people using these extensions. 

I didn't test the @Kuranes mod of shaderEditor, but I will.

PR #733, opened 2017-02-23
Removes uniform gl method cache

Kuranes:
Firefox android on samsung S6 doesn't allow js "function().call(this, ...args...)"
on webglcontextRenderingObject

### DIFF
--- a/sources/osg/Uniform.js
+++ b/sources/osg/Uniform.js
@@ -8,7 +8,6 @@ var Uniform = function ( name ) {
     this._data = undefined;
     this._transpose = false;
     this._glCall = '';
-    this._cache = undefined;
     this._name = name;
     this._type = undefined;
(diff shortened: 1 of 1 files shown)


cedricpinson:
seems more an issue on their side, I dont remember the improvement about this, if we talk about small % maybe it's better to make it work, if it's more then we should fix it

stephomi:
Did you try with `this._cache = gl[ this._glCall ].bind(gl);` and `this._cache( location, this._transpose, this._data );`? (which is the fastest method on paper btw..)

Kuranes:
( for the bug, I did test, but 'bind' not allowed either )
didn't test speed/bench gains of eithers

cedricpinson:
I feel that it's a firefox bug on a specific mobile platform. I close it, and we will see later if pop again

PR #734, opened 2017-02-23
Show Light State Bleeding Bug

Kuranes:
Special case with node above and after, above and before, and sibling
to shadowedSceneNode.
Shows how light is applied to node where no light should be, triggering complex compiler shader instead of just applying texture in a shadeless shader

As in only the left bottom cube should be lighted in red (and with a complex shader (light+material))

image 
```
root 
      \--cube
      \--mainnode
                        \-- cube
                        \-- lights
                        \-- shadowedscene 
                                                      \-- ground 
                                                       \---cube
     \--cube
```

### DIFF
--- a/tutorials/09_Shadows/index.html
+++ b/tutorials/09_Shadows/index.html
@@ -1,0 +1,19 @@
+<html><body>
+
+
+    <script type="text/javascript" src="../../examples/vendors/core.js"></script>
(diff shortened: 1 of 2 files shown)


Kuranes:
image

cedricpinson:
I dont understand the issue. A light is a positional state attribute it means that its position in the scene graph will define its position position in space, but the attribute will affect globaly. until there is a light that disable a part of the scenegraph.
Is it your issue ?

PR #740, opened 2017-02-27
Feature/shadow atlas no scissor d3 d 9999

Kuranes:


Kuranes:
- need bench on s6
- need fix on render order

Kuranes:
bench on s6:  same perf tax as with scissor
 osgjs/examples/shadowmap/?atlas=1&lightNum=4&noScissor=true&stats=1&overrideDevicePixel
atlas=&noScissor=true  => 19fps
atlas=&noScissor=false => 19fps
atlas=0 => 29fps

Still suspect that no glCear of the whole framebuffer has huge impact on "tiled gpu"
(as they have to copy back the tiles to the gpu prior to rendering)

cedricpinson:
Could you report the issue. Don't remember if we the good contact in Samsung.

Kuranes:
https://bugs.chromium.org/p/chromium/issues/detail?id=681835

cedricpinson:
If it's only at the generation we can merge it, and we will activate it only when we will have no shadow on camera on skfb

Kuranes:
imho This PR is just testing the "quad clear", no need to merge as it's same exact perf in the end on samsung.
( the glClear/Scissor codepath is better, as it's a good way for their tiled-gpu driver can detect the "only use that part of the framebuffer for vertex and pixel shading" and fix perf on their side )

( imho We can close the PR and delete the branch)

cedricpinson:
Ok thx

Kuranes:
things to check:
 - a 'whole clear' comparison.(all shadow map at once as use case is animation vs static rather than some  dynamic, some statics)

PR #741, opened 2017-02-27
[#2945] Fixes model Normal

Kuranes:
need to be normalized and use special code path to avoid adreno bugs


Kuranes:
@jtorresfabra Would be great if you could test it asap on the xiaomi adreno ?

Kuranes:
vModelNormal is a bit misleading, perhaps could we rename to vWorldNormal ?

jtorresfabra:
I agree.

PR #743, opened 2017-02-27
Renames compiler shader functions

stephomi:
Some getOrCreate were just create. Also renamed a few functions.

As a side note :
In osgjs the separation between CompilerVertex and CompilerFragment is just implicit, they are part of the same Compiler class.

Maybe it'd be nice to have a system where some functions could *only* be called from certain shader stage.

For example it's confusing that `getOrCreateModelVertex`, `getOrCreateLocalVertex`, etc... (same for normal, tangent) should only be called from the vertex stage (in the fragment we query directly the varying).
Alternatively, functions named `getOrCreateFrontBlabla` shouldn't be called from CompilerVertex.

### DIFF
--- a/sources/osgShader/Compiler.js
+++ b/sources/osgShader/Compiler.js
@@ -50,6 +50,16 @@ var Compiler = function ( attributes, textureAttributes, shaderProcessor ) {
 
 Compiler.prototype = MACROUTILS.extend( {}, CompilerVertex, CompilerFragment, {
 
+    createFragmentShader: function () {
+        this._fragmentShaderMode = true;
+        return this._createFragmentShader();
+    },
+
+    createVertexShader: function () {
+        this._fragmentShaderMode = false;
+        return this._createVertexShader();
+    },
+
     getOrCreateProjectionMatrix: function () {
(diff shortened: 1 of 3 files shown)


Kuranes:
can add  some those kinds of checks/assert with osgshader:compiler::_fragmentShaderMode ?
https://github.com/cedricpinson/osgjs/blob/master/sources/osgShader/CompilerVertex.js#L9

cedricpinson:
// =============== To Call From Vertex Shader ONLY ==================
blabla vertex

// =============== To Call From Fragment Shader ONLY ==================
blabla fragment

stephomi:
@Kuranes could be done automatically with
```
var fns = Object.keys( CompilerVertex );
var nbFunc = fns.length;

var wrapperVertexOnly = function ( fn ) {
    return function () {
        if ( this._fragmentShaderMode ) console.error( 'This function should not be called from fragment shader.' );
        return fn.apply( this, arguments );
    };
};

for ( var i = 0; i < nbFunc; ++i ) {
    CompilerVertex[ fns[ i ] ] = wrapperVertexOnly( CompilerVertex[ fns[ i ] ] );
}
```

PR #750, opened 2017-03-09
Adds 3DTiles support

jtorresfabra:
Also adds b3dm and glb ReaderWriters. 
At this moment only discrete LOD supported.


cedricpinson:
I dont want to add binary sample files anymore on the repository could we add them like in the dropbox ?

jtorresfabra:
@cedricpinson Dropbox is not well suited to host 3DTiles databases. We need to have a common database path for the files, so the database pager could do the right requests.

jtorresfabra:
Waiting for https://github.com/cedricpinson/osgjs/pull/757 to be merged.

jtorresfabra:
We will need to wait for https://github.com/AnalyticalGraphicsInc/3d-tiles-samples/issues/15

As we only support glTF 2.0 
cc:/ @pjcozzi

PR #751, opened 2017-03-13
For in experiment

stephomi:
1st commit : Replaces object.keys + for/forEach with ``for..in`` statement
Safe change, easier to read and should be fine performance wise (more benchmark/osgjs example would be great though).

2nd commit : Stops using Map for uniforms/texture and rely only on native {}
Needs more benchmark

I started using native window.Map for uniforms/textures but didn't do it because : 
- [] accessor doesn't work and it requires set/get so lots of change
- it requires `for..of` in order to iterate on it which isn't supported in android <=50 and IE11 (https://kangax.github.io/compat-table/es6/), also I have no idea about the performance of `for..of`

The 2 commits are separate (1st one can be merged without the second).


stephomi:
@jtorresfabra @cedricpinson @Kuranes

stephomi:
Can be merged imo

cedricpinson:
Did not notice performance improvement but agree if it's not a problem it can simplify some code. Did you check on firefox / mobile ?

stephomi:
yes same

PR #753, opened 2017-03-15
Changes to support new glTF2 updates

jtorresfabra:


jtorresfabra:
@cedricpinson envs works for me too.

PR #759, opened 2017-03-24
Improves error and warning for shader compiler

stephomi:
sans titre


stephomi:
Also improves shader name in extension by displaying material name in it

PR #764, opened 2017-03-29
Renames intersector

stephomi:
@jtorresfabra @cedricpinson

### DIFF
--- a/sources/osg/KdTreeSphereIntersector.js
+++ b/sources/osg/KdTreeSphereIntersector.js
@@ -1,17 +1,15 @@
 'use strict';
 var MACROUTILS = require( 'osg/Utils' );
 var vec3 = require( 'osg/glMatrix' ).vec3;
-var KdTreeRayIntersector = require( 'osg/KdTreeRayIntersector' );
+var KdTreeLineSegmentIntersector = require( 'osg/KdTreeLineSegmentIntersector' );
 var TriangleSphereIntersector = require( 'osgUtil/TriangleSphereIntersector' );
 
 
 var KdTreeSphereIntersector = function () {
-
     this._intersector = new TriangleSphereIntersector();
-
 };
 
-KdTreeSphereIntersector.prototype = MACROUTILS.objectInherit( KdTreeRayIntersector.prototype, {
+KdTreeSphereIntersector.prototype = MACROUTILS.objectInherit( KdTreeLineSegmentIntersector.prototype, {
 
     init: function ( intersections, center, radius, nodePath ) {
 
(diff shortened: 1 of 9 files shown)


sources/osgUtil/TriangleSphereIntersector.js
```diff
@@ -1,21 +1,26 @@
 'use strict';
 var MACROUTILS = require( 'osg/Utils' );
 var vec3 = require( 'osg/glMatrix' ).vec3;
-var TriangleIntersector = require( 'osgUtil/TriangleIntersector' );
+var TriangleLineSegmentIntersector = require( 'osgUtil/TriangleLineSegmentIntersector' );
 
-var TriangleIntersection = function ( normal, i1, i2, i3 ) {
+// implicitly match with TriangleLineSegment intersection
+var TriangleSphereIntersection = function ( normal, i1, i2, i3, r1, r2, r3 ) {
```
jtorresfabra: I think we just could reuse TriangleIntersection, and not have two different TriangleIntersection containers which basically have exactly the same.

sources/osgUtil/TriangleLineSegmentIntersector.js
```diff
@@ -5,7 +5,7 @@ var mat4 = require( 'osg/glMatrix' ).mat4;
 var TriangleIndexFunctor = require( 'osg/TriangleIndexFunctor' );
 var ComputeMatrixFromNodePath = require( 'osg/computeMatrixFromNodePath' );
 
-var TriangleIntersection = function ( normal, i1, i2, i3, r1, r2, r3 ) {
+var TriangleLineSegmentIntersection = function ( normal, i1, i2, i3, r1, r2, r3 ) {
```
jtorresfabra: Don't agree in changin that name. I would stick to TriangleIntersector, I don't see any good reason to chnage the OSG standard. Also TriangleIntersector is a utility class, users shouldn't be using this directly. In fact in OSG is inside a cpp and not extensible at all.

sources/osg/KdTreeSphereIntersector.js
```diff
@@ -1,17 +1,15 @@
 'use strict';
 var MACROUTILS = require( 'osg/Utils' );
 var vec3 = require( 'osg/glMatrix' ).vec3;
-var KdTreeRayIntersector = require( 'osg/KdTreeRayIntersector' );
+var KdTreeLineSegmentIntersector = require( 'osg/KdTreeLineSegmentIntersector' );
```
jtorresfabra: For this I agree, because users could be using directly this intersector, and it is not a ray is a linesegment. So OK.

PR #765, opened 2017-04-05
Adds webgl2 support

stephomi:
Needs polishing


sources/osg/Texture.js
```diff
@@ -712,15 +801,36 @@ Texture.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( GLO
...
+        if ( !WebglCaps.instance().isWebGL2() ) {
+            return internalFormatGl2ToGl1[ internalFormat ] || internalFormat;
+        }
+
+        // gl2
+        var map = internalFormatGl1ToGl2[ this._type ];
+        return ( map && map[ internalFormat ] ) || internalFormat;
+    },
+
+    getTypeGL: function () {
```
cedricpinson: If used only internally maybe it should be prefixed by _

sources/osg/WebGLCaps.js
```diff
@@ -273,13 +260,13 @@ WebGLCaps.prototype = {
...
     hasLinearHalfFloatRTT: function ( gl ) {
-        return this._webGLExtensions[ 'OES_texture_half_float_linear' ] && this.checkSupportRTT( gl, Texture.HALF_FLOAT, Texture.LINEAR );
+        return this._webGLExtensions[ 'OES_texture_half_float_linear' ] && this.checkSupportRTT( gl, Texture.HALF_FLOAT_OES, Texture.LINEAR );
     },
     hasLinearFloatRTT: function ( gl ) {
         return this._webGLExtensions[ 'OES_texture_float_linear' ] && this.checkSupportRTT( gl, Texture.FLOAT, Texture.LINEAR );
     },
     hasHalfFloatRTT: function ( gl ) {
-        return this._webGLExtensions[ 'OES_texture_half_float' ] && this.checkSupportRTT( gl, Texture.HALF_FLOAT, Texture.NEAREST );
+        return this._webGLExtensions[ 'OES_texture_half_float' ] && this.checkSupportRTT( gl, Texture.HALF_FLOAT_OES, Texture.NEAREST );
```
stephomi: The function name already implied that is it gl1/gl2 agnostic, it just tells you if can render to a h/float texture.

sources/osg/Geometry.js
```diff
@@ -315,8 +318,7 @@ Geometry.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( No
         if ( !this._primitives.length ) return;
 
         if ( this._extVAO === undefined && Geometry.enableVAO ) { // will be null if not supported
-            var extVAO = WebGLCaps.instance( state.getGraphicContext() ).getWebGLExtension( 'OES_vertex_array_object' );
-            this._extVAO = extVAO;
+            this._extVAO = WebGLCaps.instance( state.getGraphicContext() ).getWebGLExtension( 'OES_vertex_array_object' );
```
cedricpinson: you could probably also set this._extVAO to gl if in webgl2. Then in the release it will work as expected. Not sure if we need to rename the variable '_extVAO'

sources/osg/KdTree.js
```diff
@@ -121,7 +121,7 @@ BuildKdTree.prototype = {
             if ( mode === PrimitiveSet.TRIANGLES )
                 totalLenArray += prim.getCount();
             else if ( mode === PrimitiveSet.TRIANGLE_STRIP || mode === PrimitiveSet.TRIANGLE_FAN )
-                totalLenArray += ( prim.getCount() - 2 ) * 3;
+                totalLenArray += Math.abs( ( prim.getCount() - 2 ) ) * 3;
```
cedricpinson: Would have be good to have this in a separate commit for the history because it's clearly not related to webgl2

sources/osg/State.js
```diff
@@ -787,7 +787,8 @@ State.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( Objec
 
         if ( this._currentVAO !== vao ) {
 
-            this._extVAO.bindVertexArrayOES( vao );
+            var gl = this._graphicContext;
```
cedricpinson: we dont need gl variable it's not re used

sources/osg/WebGLCaps.js
```diff
@@ -187,26 +185,11 @@ WebGLCaps.prototype = {
     isWebGL2: function () {
         return this._isGL2;
     },
+
     // inevitable bugs per platform (browser/OS/GPU)
     initBugDB: function () {
 
-        var p = this._webGLPlatforms;
```
cedricpinson: Why removing this part ?

sources/osg/Geometry.js
cedricpinson: Probably getWebGLExtension should be renamed because we dont do the difference between supporting an extension and having the feature in the core (webgl2).

stephomi: webgl1 extension (that are in core in gl2) are added to gl context automatically, so this._extVAO can simply be a boolean here

sources/osg/KdTree.js
stephomi: k

sources/osg/State.js
stephomi: k

sources/osg/WebGLCaps.js
stephomi: Adding it back

sources/osg/Options.js
```diff
@@ -1,4 +1,5 @@
 'use strict';
+var Notify = require( 'osg/notify' );
```
cedricpinson: should be notify and not Notify

sources/osg/Geometry.js
```diff
@@ -63,12 +63,13 @@ Geometry.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( No
...
 
-        if ( !this._extVAO ) return;
+        if ( !this._useVAO ) return;
 
         for ( var prgID in this._vao ) {
             if ( this._vao[ prgID ] ) {
                 var vao = this._vao[ prgID ];
-                this._extVAO.deleteVertexArrayOES( vao );
+                // TODO not the best way to get the context
+                WebGLCaps.instance().getContext().deleteVertexArray( vao );
```
cedricpinson: we can't do that, we need to store the gl context, else if you have two context on the same page it will not work.

stephomi: k

sources/osg/Options.js
stephomi: k

PR #771, opened 2017-04-18
Adds webgl2 disjoint timer query support

stephomi:
Timer query is **still** an extension in webgl2, but it shares most of the constant with the gl1 extension.

Exception is the `getQueryObjectExt` that is now `getQueryParamter` (and not `getQueryObject`)

### DIFF
--- a/sources/osg/TimerGPU.js
+++ b/sources/osg/TimerGPU.js
@@ -20,7 +20,7 @@ var TimerGPU = function ( gl ) {
 
     if ( gl ) {
 
-        var ext = WebGLCaps.instance().getDisjointTimerQuery();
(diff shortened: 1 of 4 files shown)


sources/osg/TimerGPU.js
```diff
@@ -19,14 +20,17 @@ var TimerGPU = function ( gl ) {
 
     if ( gl ) {
 
-        var ext = gl.getExtension( 'EXT_disjoint_timer_query' );
+        var ext = WebGLCaps.instance().getDisjointTimerQuery();
```
cedricpinson: to be consistent it would be better to use the gl instance to be sure if two osgjs context exist there is no mismatch between instance.

```
WebGLCaps.instance(gl).getDisjointTimerQuery();
```

PR #777, opened 2017-05-10
Switch node

spotscale:


spotscale:
Removing unintended package.json commit and fixing lint stuff, creating new PR...

PR #781, opened 2017-05-12
Allows silent fail webgl2 context try

Kuranes:
Otherwise old compute with no webgl2 supports failed because the
previous webgl detection was relying on a non standardized error message

### DIFF
--- a/sources/osgViewer/webgl-utils.js
+++ b/sources/osgViewer/webgl-utils.js
@@ -96,6 +96,11 @@ var WebGLUtils = function () {
         "It doesn't appear your computer can support WebGL.<br/>" +
         '<a href="http://get.webgl.org">Click here for more information.</a>';
 
+    /*
+     * Allows to prevent the global canvas error code on webgl2
+     * as we still have to check if we can fallback on webgl1
+     */
+    var doReportCreationError = true;
     /**
      * Creates a webgl context. If creation fails it will
      * change the contents of the container of the <canvas>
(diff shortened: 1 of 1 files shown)


sources/osgViewer/webgl-utils.js
```diff
@@ -110,7 +115,7 @@ var WebGLUtils = function () {
         /** function:(msg) */
         opt_onError ) {
         function handleCreationError( msg ) {
-            if ( msg.indexOf( 'WebGL2' ) !== -1 ) return;
+            if ( !doReportCreationError ) return;
```
cedricpinson: It will not work because setupWebGL call create3DContext and create3DContext set the global variable doReportCreationError to false, the message will never be displayed, unless I missed something.
Maybe just add a optional parameter to function instead of global variable

Kuranes: I tested it and it works :)

"doReportCreationError"  is "true" by default and if we test WebGL2, we set it back to "true" once we unsuccessfully tried WebGL2 (https://github.com/cedricpinson/osgjs/blob/c588a09269cee85ef9a4d3264a9b5e8cf2916287/sources/osgViewer/webgl-utils.js#L182)

Parameter to function won't work because handleCreationError is a callback

Kuranes: To be clear the intention is that the message is NOT displayed when trying to get a Webgl2 context.
We only display the message if we cannot get a Webgl1 context

cedricpinson: Yeah I get it, just handling state with global this way it not very convenient especially when there are different code path. but ok

PR #786, opened 2017-05-17
Fixes unwanted culling for lights by default

jtorresfabra:


andek714:
This fixes my problem with strange light behavior

PR #790, opened 2017-05-23
Optimizes shadow compute caster scene bound

Kuranes:
Do it once per shadowed scene instead of every shadowmap

### DIFF
--- a/sources/osgShadow/ShadowMap.js
+++ b/sources/osgShadow/ShadowMap.js
@@ -2,7 +2,6 @@
 var Camera = require( 'osg/Camera' );
 var CullVisitor = require( 'osg/CullVisitor' );
 var BlendFunc = require( 'osg/BlendFunc' );
-var ComputeBoundsVisitor = require( 'osg/ComputeBoundsVisitor' );
 var Depth = require( 'osg/Depth' );
 var FrameBufferObject = require( 'osg/FrameBufferObject' );
 var mat4 = require( 'osg/glMatrix' ).mat4;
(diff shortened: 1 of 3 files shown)


Kuranes:
Perf improvemvent of doing computeBounds once instead of N times per shadow counts for a (N-1)*50% of shadow js pass improvement.
For example https://sketchfab.com/models/9df1dc92bcbd4ca2a911c878e212e88c/edit?debug3d=1

the CullShadowCasting block code goes from:
desktop ~3ms => ~0.88 
mobile   ~10ms => ~3.5ms

sources/osgShadow/ShadowMap.js
```diff
@@ -930,14 +873,6 @@ ShadowMap.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInherit( S
             //console.log( 'shadow early OUT' );
```
cedricpinson: you can remove the console.log comment

cedricpinson: the comment in aimShadowCastingCamera https://github.com/cedricpinson/osgjs/pull/790/files?diff=split#diff-8cf8f17e8132314edcfa57e87eda7ab4R686 says that it's a frustrum, but the variable is bbox, is it accurate or should be a renamed of variable and comments ?

sources/osgShadow/ShadowedScene.js
```diff
@@ -99,24 +178,70 @@ ShadowedScene.prototype = MACROUTILS.objectLibraryClass( MACROUTILS.objectInheri
...
+                if ( !st || !st.valid() ) continue;
+
+                if ( st.isDirty() ) {
+                    isDirty = true;
+                    st.init();
+                }
+
+                if ( st.isEnabled() || !st.isFilledOnce() ) {
+                    isDirty = true;
+
```
cedricpinson: remove the blank line

sources/osgShadow/ShadowMap.js
Kuranes: Will rename as bbox 

( The bbox is used in as the frustumBounds, doesn't inmply it's a frustum, but that was when I was doing ops on it in that function, which not the case anymore)

PR #792, opened 2017-05-31
Clears the whole atlas at the same time

Kuranes:
Clears whole texture at once for shadowAtlas Avoid samsung bug on scissor+glClear
 Fixes shadowAtlas recent light and compiler refactors broke it.

### DIFF
--- a/sources/osg/RenderStage.js
+++ b/sources/osg/RenderStage.js
@@ -264,13 +264,6 @@ MACROUTILS.createPrototypeObject(
             }
             state.applyAttribute(this.viewport);
 
-            // fragment clipping
-            if (this.camera) {
-                var scissor =
-                    this.camera.getStateSet() && this.camera.getStateSet().getAttribute('Scissor');
-                if (scissor) state.applyAttribute(scissor);
-            }
-
             /*jshint bitwise: false */
             if (this.clearMask !== 0x0) {
                 if (this.clearMask & gl.COLOR_BUFFER_BIT) {
--- a/sources/osgShadow/ShadowMapAtlas.js
+++ b/sources/osgShadow/ShadowMapAtlas.js
@@ -1,14 +1,15 @@
 'use strict';
 var notify = require('osg/notify');
-var StateAttribute = require('osg/StateAttribute');
-var Texture = require('osg/Texture');
-var Uniform = require('osg/Uniform');
-var MACROUTILS = require('osg/Utils');
-var Scissor = require('osg/Scissor');
 var vec4 = require('osg/glMatrix').vec4;
+var Camera = require('osg/Camera');
+var MACROUTILS = require('osg/Utils');
 var ShadowTechnique = require('osgShadow/ShadowTechnique');
 var ShadowTextureAtlas = require('osgShadow/ShadowTextureAtlas');
 var ShadowMap = require('osgShadow/ShadowMap');
+var Texture = require('osg/Texture');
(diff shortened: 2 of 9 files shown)


sources/osg/RenderStage.js
```diff
@@ -265,10 +265,21 @@ MACROUTILS.createPrototypeObject(
             state.applyAttribute(this.viewport);
 
             // fragment clipping
+
             if (this.camera) {
```
cedricpinson: All this the code in this.camera is useless, it was necessary if we cleared by sub regions but if we dont do that we dont need to apply it here

sources/osgShadow/ShadowMapAtlas.js
```diff
@@ -314,10 +382,60 @@ MACROUTILS.createPrototypeObject(
...
+                if (this._shadowMaps.indexOf(shadowMap) !== -1) return;
+            }
+
+            this._shadowMaps.push(shadowMap);
+
+            var light = shadowMap.getLight();
+            if (this._lights.indexOf(light) === -1) {
+                this._lights.push(light);
+            } else {
+                console.assert(false);
```
cedricpinson: uncool

sources/osgShadow/ShadowMapAtlas.js
```diff
@@ -314,10 +382,60 @@ MACROUTILS.createPrototypeObject(
...
             }
+
+            if (this._shadowedScene) shadowMap.setShadowedScene(this._shadowedScene);
+
+            var lightNumberArray = this._texture.getLightNumberArray();
+            if (lightNumberArray.indexOf(shadowMap.getLightNumber()) === -1) {
+                lightNumberArray.push(light.getLightNumber());
+            } else {
+                console.assert(false);
```
cedricpinson: double uncool

sources/osgShadow/ShadowTextureAtlas.js
```diff
@@ -20,10 +20,10 @@ var ShadowTextureAtlas = function() {
     this._uniforms = {};
     this._lightNumberArray = []; // default for a valid cloneType
 
-    this._viewMatrices = [];
-    this._projectionMatrices = [];
-    this._depthRanges = [];
-    this._mapSizes = [];
+    this._viewMatrices = {};
```
cedricpinson: why switching [] to {}

sources/osgShadow/ShadowMapAtlas.js
Kuranes: ha double sorry 
that debug code to double check error in lightnumber/index was helpful, but definitely not needed now

sources/osgShadow/ShadowTextureAtlas.js
Kuranes: As we're now indexing by lightnumbers now, not using lightArrayIndex, so it's a sparse array
if user put lights 8,9,10, better use an associative array ?

Kuranes:
- Fixed all comments

sources/osgShadow/ShadowTextureAtlas.js
cedricpinson: no idea what is better in this case, but simpler is probably better

cedricpinson: I would probably keep array because size will not be more up to 8 I guess

Kuranes:
- rebased with latest
- use array for textureAtlas uniforms

PR #795, opened 2017-06-06
Fixes css in examples index

jtorresfabra:


examples/templates/css/sidebar.css
```diff
@@ -114,10 +121,48 @@ body {
...
+    margin-left: auto;
+    height: inherit;
+}
+.row {
+    margin-right: -15px;
+    margin-left: -15px;
+    height: inherit;
+}
+
+.btn {
```
arthurjamain: There is a lot of stuff in there :) Are you sure it's all useful ?

jtorresfabra: Just copied the bootstrap properties but you are right, removed not needed properties.

jtorresfabra:
Added index.html to the beautifier.

PR #798, opened 2017-06-13
Metric outputs

spotscale:
Modifies notify levels and moves timing logs to info level.

### DIFF
--- a/sources/osg/Utils.js
+++ b/sources/osg/Utils.js
@@ -177,18 +177,17 @@ var times = {};
 
 // we bind the function to Notify.console once and for all to avoid costly apply function
 
-Utils.time = ( Notify.console.time || function ( name ) {
+Utils.time = ( function ( name ) {
     times[ name ] = Timer.instance().tick();
 } ).bind( Notify.console );
 
-Utils.timeEnd = ( Notify.console.timeEnd || function ( name ) {
-
+Utils.timeEnd = ( function ( name ) {
     if ( times[ name ] === undefined )
         return;
 
     var duration = Timer.instance().deltaM( times[ name ], Timer.instance().tick() );
 
-    Notify.debug( name + ': ' + duration + 'ms' );
+    Notify.info( name + ': ' + duration + 'ms' );
     times[ name ] = undefined;
 
 } ).bind( Notify.console );
--- a/sources/osg/notify.js
+++ b/sources/osg/notify.js
@@ -2,12 +2,15 @@
 
 var Notify = {};
 
-// must be uppercase and match loggers
-Notify.DEBUG = 0;
-Notify.INFO = 1;
-Notify.NOTICE = Notify.LOG = 2;
-Notify.WARN = 3;
-Notify.ERROR = 4;
+// Range of notify levels from DEBUG through to FATAL
+// ALWAYS is reserved for forcing the absorption of all messages.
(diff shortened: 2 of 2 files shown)


sources/osg/Utils.js
```diff
@@ -177,18 +177,17 @@ var times = {};
...
-Utils.timeEnd = ( Notify.console.timeEnd || function ( name ) {
-
+Utils.timeEnd = ( function ( name ) {
     if ( times[ name ] === undefined )
         return;
 
     var duration = Timer.instance().deltaM( times[ name ], Timer.instance().tick() );
 
-    Notify.debug( name + ': ' + duration + 'ms' );
+    Notify.info( name + ': ' + duration + 'ms' );
```
jtorresfabra: Maybe we should keep this as Notify.log to not change the default behaviour

spotscale: Changed back to debug. Will change behavior compared to using console.timeEnd like before that was always printed, but more correct to put this at a debug level

jtorresfabra: Agree, for me this should be at debug/info level.

cedricpinson:
I dont think we should change it this way, timer is not something that must describe the level of log. For example with this modification we will not be able to do that anymore:

```
processModelJSON : 0.86ms

Download file.osgjs.gz (Async)
  |  createAndInitFeatures : 99.57ms
Download file.osgjs.gz (Async) : 251.75ms

JSONparse file.osgjs.gz : 0.47ms

parseSceneGraph file.osgjs.gz (Async) : 5.38ms

DownloadAndLoadBuffer file.osgjs.gz (Async) : 1031.27ms

initSceneGraph - create saved materials : 10.71ms

initSceneGraph - processSceneGraph : 1.89ms

initSceneGraph - apply materials : 3.39ms

optimizer
  |  optimizer - materials : 0.71ms
  |  optimizer - nodes : 80.86ms
  |  Geometry count : before 9, after 7
  |  Material count : before 3, after 1
optimizer : 81.93ms

featureFinished
  |  compute KdTree : 612.30ms
  |  drawFirstFrame : 4.80ms
  |  drawSecondFrame : 271.10ms
featureFinished : 943.48ms

Bound center : [ -0.0001373291015625, -0.00005340576171875, 89.64857196807861 ]
Bound radius : 177.71846491219168
Bbox min : [ -99.92720031738281, -99.92990112304688, -18.1125545501709 ]
Bbox max : [ 99.92692565917969, 99.92979431152344, 197.40969848632812 ]
```
It's in notice log level we use it

jtorresfabra:
@cedricpinson is right. 
I did not notice that we loose some features by not using console.time like the folding:
if you call time(x) , time(y),  timend(y) timeend(x) , it is in a folder. So a better approach could be to wrap console.time in a function and only call it if the desired log level matches the with the call, something like:

osg.time( 'featureFinished' );
osg.timeEnd( ‘featureFinished’, notify.NOTICE );

Other approach is to mimic the same behaviopur that console.time.

spotscale:
I understand....all I wanted right now was to remove the vast console output when moving around our PagedLOD models. I'm afraid I don't have time to change it in the way you suggest, so I'd appreciate any suggestion for a quick fix in the meantime!

cedricpinson:
this is fixed by this PR https://github.com/cedricpinson/osgjs/pull/819

PR #804, opened 2017-06-26
Changes testing framework to karma + chrome headless

jtorresfabra:


jtorresfabra:
Won't apply til chrome-headless gets more mature and can be installed via npm.

PR #812, opened 2017-07-02
Adds node testing, replaces phantom.js

jtorresfabra:


webpack.config.js
```diff
@@ -21,6 +21,10 @@ module.exports = {
         libraryTarget: 'umd',
         library: 'OSG'
     },
+    node: {
+       fs: "empty"
+    },
+    target: 'node',
```
jtorresfabra: Not sure if it impacts other code, see https://webpack.js.org/concepts/targets/ for info. I did this to be able to access to filesystem while testing in node. 

Maybe @arthurjamain can shed some light.

cedricpinson: We should probably have a target for node and web using something like:
https://github.com/TheLarkInn/compare-webpack-target-bundles
https://stackoverflow.com/questions/35903246/how-to-create-multiple-output-paths-in-webpack-config
Then we can use both web and node

.eslintrc.json
```diff
@@ -17,6 +17,9 @@
         "mocha": true
     },
     "globals": {
+        "process": false,
```
cedricpinson: I dont want we add tis globally, there is some syntax to enable or disable some rules with eslint. It's better to use that instead of enabling globals names on the whole project.
check here http://eslint.org/docs/rules/no-undef#node

index.html
```diff
@@ -113,6 +113,24 @@
         li.appendChild( link );
         sidebar.appendChild( li );
 
+        // And benchmarks
+        var li = document.createElement( 'li' );
```
cedricpinson: this part must be really simplified. I will have a look

cedricpinson:
Last thing is that the deploy script maybe should be adjusted if mocha is used from .node_modules directory. Did you test it ?

.eslintrc.json
jtorresfabra: Fixed.

index.html
jtorresfabra: I really don't want this to interfere in the PR I just added it for having in the main index, but if it is going to be a problem I rather remove it.

package.json
```diff
@@ -8,12 +8,15 @@
     "eslint-plugin-node": "^5.1.0",
     "eslint-plugin-promise": "^3.5.0",
     "eslint-plugin-standard": "^3.0.1",
-    "grunt-contrib-connect": "~0.9",
     "grunt-documentation": "*",
-    "grunt-mocha": "^1.0.2",
+    "grunt-contrib-connect": "~0.9",
+    "grunt-execute": "^0.2.2",
     "grunt-plato": "~0.2.1",
+    "bluebird": "*",
```
jtorresfabra: Needed to add bluebird as a dev dependency for node execution, otherwise it cannot resolve bluebird, weird as it is being added as external, but does not seem to work.

jtorresfabra:
> Last thing is that the deploy script maybe should be adjusted if mocha is used from .node_modules directory. Did you test it ?

Umm not sure I get you, sorry.

PR #814, opened 2017-07-03
Separate lighting

stephomi:


cedricpinson:
Please adds a bit of documentation how DECLARE works and what is supported, no need to write a lot but give an overview of usage and the special case of 'result' .
Could be good to add it in the PR description or better in the file that handle declare with a link in the PR description

sources/osgShader/node/lightCommon.glsl
```diff
@@ -0,0 +1,77 @@
...
+        const in vec3 lightViewPosition,
+        const in float lightSpotCutOff,
+        const in float lightSpotBlend,
+                          
+        out float attenuation,
+        out vec3 eyeLightDir,
+        out float dotNL) {
+
+    attenuation = 0.0;
+    if (lightSpotCutOff <= 0.0) return;
```
Kuranes: mmm  dotNL/attenuation/eyelightDir being undefined if we return here no ?

(no more "lighted" var that would help there)

sources/osgShader/node/lightCommon.glsl
```diff
@@ -0,0 +1,77 @@
...
+
+    eyeLightDir = lightViewPosition - viewVertex;
+    float dist = length(eyeLightDir);
+
+    eyeLightDir = dist > 0.0 ? eyeLightDir / dist : vec3( 0.0, 1.0, 0.0 );
+    float cosCurAngle = dot(-eyeLightDir, lightViewDirection);
+    if (cosCurAngle < lightSpotCutOff) return;
+
+    float spot = 1.0;
+    if (lightSpotBlend > 0.0) {
```
Kuranes: perhaps we could take that refactor occasion to prevent in js code that lightSpotBlend is never near 0 and get rid

sources/osgShader/node/lightCommon.glsl
```diff
@@ -0,0 +1,77 @@
...
+        spot = cosCurAngle * smoothstep(0.0, 1.0, (cosCurAngle - lightSpotCutOff) / lightSpotBlend);
+        if (spot < 0.0) return;
```
Kuranes: mmm  dotNL/attenuation being undefined if we return here no ?

sources/osgShader/node/lightCommon.glsl
```diff
@@ -0,0 +1,77 @@
...
+        const in vec3 normal,
+        const in vec3 lightViewDirection,
+  
+        out float attenuation,
+        out vec3 eyeLightDir,
+        out float dotNL) {
+
+    attenuation = 1.0;
+    eyeLightDir = -lightViewDirection;
+    dotNL = dot(-lightViewDirection, normal);
```
Kuranes: not sure it even saves a cycle but resuse eyeLightDir coudl do it

sources/osgShader/node/textures.glsl
```diff
@@ -1,15 +1,19 @@
+#pragma DECLARE_FUNCTION
 vec3 textureRGB(const in sampler2D tex, const in vec2 uv) {
     return texture2D(tex, uv.xy ).rgb;
```
Kuranes: not very interesting, but just noted strange spacing here and below if you happen to pass by

sources/osgShader/utils.js
```diff
@@ -107,23 +111,279 @@ var callFunction = function(funcName, output, inputs) {
...
-    if (osgShader.debugShaderNode) return debug.join('\n') + '\n' + callString;
+    if (osgShader.debugShaderNode) {
+        return debug.join('\n') + '\n' + callString;
+    }
 
     return callString;
 };
 
+var extractMapVariable = function(option) {
+    var res = option.match(/([^\s]+):([^\s]+)/g);
```
Kuranes: regexp really takes some time to decrypt/read when you read  code, 
I love it very much when there is a sample of what it's supposed to match.
```
// ie myNodeName:myResult
 var res = option.match(/([^\s]+):([^\s]+)/g);
```

Kuranes: ( same for other regexp below )

sources/osgShadow/shaders/shadowLinearSoft.glsl
```diff
@@ -1,23 +1,45 @@
...
+    float rb = texture2DCompare(depths, centroidUV + size.zw * vec2(1.0, 0.0), compare, clampDimension);
+    float rt = texture2DCompare(depths, centroidUV + size.zw * vec2(1.0, 1.0), compare, clampDimension);
+
+#ifdef _OUT_DISTANCE
+    outDistance = mix(mix(lb, lt, f.y), mix(rb, rt, f.y), f.x);
+#endif
 
-    vec2 f = fract(uv*size.xy+0.5);
-    vec2 centroidUV = floor(uv*size.xy+0.5)*size.zw;
+    lb = step(lb, 0.0);
```
Kuranes: if we  change "lb,lt,rb,rt" to a "vec4" we could do a vectorized step against "vec4 = step(vec4,vec4(0))" there
and then perhaps get early out at 1.0 or 0.0 with "all(equals(0.0))" and "all(equals(1.0))" or "dot(vec4,vec4)==1"
(most common case of shadowmaping is totally outside/inside of penumbra so early out makes sense)

sources/osgShader/node/lightCommon.glsl
stephomi: will do

sources/osgShader/node/lightCommon.glsl
stephomi: will do

sources/osgShader/node/lightCommon.glsl
stephomi: will do

sources/osgShader/node/textures.glsl
stephomi: will do

sources/osgShader/utils.js
stephomi: k

stephomi: (added a few comments)

sources/osgShadow/shaders/shadowLinearSoft.glsl
stephomi: can't tell if it makes a difference or not, I did the step vec4 but not the early out (dunno if it's worth it)

stephomi:
Added comments and removed useless conditional in spot code

sources/osgShadow/shaders/shadowLinearSoft.glsl
Kuranes: np, I'll do it in my shadow task round to bench it on multiple cases

sources/osgShader/node/lightCommon.glsl
```diff
@@ -0,0 +1,68 @@
...
+                          
+        out float attenuation,
+        out vec3 eyeLightDir,
+        out float dotNL) {
+
+    eyeLightDir = lightViewPosition - viewVertex;
+    float dist = length(eyeLightDir);
+    eyeLightDir = dist > 0.0 ? eyeLightDir / dist : vec3( 0.0, 1.0, 0.0 );
+
+    float cosCurAngle = dot(-eyeLightDir, lightViewDirection);
```
Kuranes: mmm yes, without the if it might need perhaps a 
```cosCurAngle = max(0.0, dot(-eyeLightDir, lightViewDirection);```

stephomi: smoothstep below will clamp to 0

PR #821, opened 2017-07-12
Adds hammer disable option

stephomi:


sources/osgGA/CADManipulator.js
```diff
@@ -146,7 +146,7 @@ MACROUTILS.createPrototypeObject(
             this._orientation = quat.create();
             this._pivotPoint = vec3.create();
 
-            this._eye = undefined;
+            this._eye = vec3.create();
```
jtorresfabra: I think you will broke the setHomePosition method if you do that

jtorresfabra: Something is brokenin the CADManipulator. The pivot point is not working as expected and computeHomePosition is throwing NaN. So it really does not make a difference to add/remove this as the manipulator needs to be fixed. I will need to bisect to see what happened.

stephomi: ok!

PR #825, opened 2017-07-18
Removes undefined call to noDraw

Kuranes:


cedricpinson:
Could you explain a bit more why you removed those lines ? I miss the context

Kuranes:
Model where first frame is without 3D model in camera view => segfault during load because calling undefined method.

because noDraw is a shadowMap method, not a shadowedScene method.
- The shadowmapp::noDraw call are done just below in shadowed scene
- The remove mask visitor is needed only once, and done by shadowed scene once we know it's a "noDraw" case.

removing those doesn't change any functionnality, just prevent segfault and makes it faster (no remove node mask "per light/shadow")

PR #831, opened 2017-07-20
Removes deprecated promise API

jtorresfabra:


cedricpinson:
Did you check performance with those changes ?

jtorresfabra:
Yes I did, but better to test in other devices.

jtorresfabra:
For me it was the same performance.

cedricpinson:
The PR fails the tests actually see 
https://travis-ci.org/cedricpinson/osgjs/builds/255721797#L3213
I think there is an issue with the PR that ensured unittest fails if no test ran

jtorresfabra:
That's good. It means the PR I did about make the node process fail when a test fail is working. There was indeed a test fail and a subtle bug that is now fixed. ready for review now!

jtorresfabra:
Need to add some performance tests to make sure we don't loose a ms :)

cedricpinson:
@jtorresfabra ping, what's the status ?

PR #836, opened 2017-07-27
Adds ability to enable/disable tracing to Notify

Kuranes:
helpful to get full stack of the log call


cedricpinson:
is the url parameter not enough ?

Kuranes:
painful to setup on each url when testing lots of differents models.
Though for consistency, both should be settable from URL and from setNotifyLevel (or a separate setter)
Setting logLevel using URL options would be nice

PR #837, opened 2017-08-07
Adds earlyout on datapager sort

Kuranes:
Gain few cpu time ms on old mobile and lowend desktop cpu as it's called each frame.
With that commit, the function disappear from the profiler,  as it seems the native "this._pendingNodes.sort" call is not that optimized, And as it's the most common case, I think it's worth it for such a small change.

### DIFF
--- a/sources/osgDB/DatabasePager.js
+++ b/sources/osgDB/DatabasePager.js
@@ -208,7 +208,7 @@ MACROUTILS.createPrototypeObject(
         },
 
         addLoadedDataToSceneGraph: function(frameStamp, availableTime) {
-            if (availableTime <= 0.0) return 0.0;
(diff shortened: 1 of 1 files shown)


Kuranes:
@jtorresfabra in the case of paged, not sure about that, but Maybe the sort comparator func could be declared once and for all, instead of at each call.
Not sure if JS compiler optimizer will optimize that away as a "const func" or do the func allocate.
(many does a warning about "do not declare function in loop" and it did appear in profiler so...)

jtorresfabra:
>@jtorresfabra in the case of paged, not sure about that, but Maybe the sort comparator func could be >declared once and for all, instead of at each call.

Ok will do, thx.

PR #840, opened 2017-08-08
Adds error when using non declared uniform

stephomi:


tests/osgShader/ShaderGenerator.js
```diff
@@ -86,7 +87,7 @@ module.exports = function() {
             stateSet.setAttributeAndModes(getShader());
 
             state.pushStateSet(stateSet);
-            state.apply();
+            state.applyStateSet(stateSet);
```
cedricpinson: replace:
state.pushStateset(stateSet);
state.applyStateSet(stateSet);

by:
state.applyStateSet(stateSet);

PR #841, opened 2017-08-09
Adds option to filter Stats

Kuranes:
Alleviates performance impact of stats, alleviates readability, makes it usable on phones.
All stat is 
?stats=1
but filtered stat is 
?stats=frame
?stats=culllightsource
etc.


cedricpinson:
I merged it but it will probably be different soon, I am refactoring stats to use gl text instead and removing rstats deps

PR #842, opened 2017-08-09
Adds Pooled dualStack to reduce allocations

Kuranes:


Kuranes:
Needs heavy review, as it's really basic code change, to make allocation reduce on a single test model, but without deep understanding of the code.

cedricpinson:
did you test performance example before after with and without complex stateset ? I am interested about the number on differents browser.

Kuranes:
tried on both chrome and firefox. 
?stats=1&deep=3&complexStateSet=1
Chrome => no difference whatsoever.
Firefox => small difference. 130-150ms before vs 90-120ms after.

sources/osg/Stack.js
```diff
@@ -1,22 +1,23 @@
...
+        this._values.push(value);
+        this._back = value;
     },
     pop: function() {
-        var values = this.values;
+        var values = this._values;
+        if (!values.length) return undefined;
         var value = values.pop();
-        this.back = values[values.length - 1];
+        this._back = values.length ? values[values.length - 1] : undefined;
```
cedricpinson: Not sure but maybe it would make sense to have the same api than the pool and maybe transform this stack to StackPool

sources/osg/State.js
```diff
@@ -821,15 +810,15 @@ MACROUTILS.createPrototypeObject(
             if (_attributeArray[index] === undefined) {
                 this._createAttributeStack(_attributeArray, index, attribute);
             } else {
-                _attributeArray[index].globalDefault = attribute;
+                _attributeArray[index]._globalDefault = attribute;
```
cedricpinson: you renamed one var upper in the code _attributeArray to attributeArray . If you change one change them all, be consistent

sources/osg/State.js
```diff
@@ -821,15 +810,15 @@ MACROUTILS.createPrototypeObject(
...
             }
         },
 
         getGlobalDefaultAttribute: function(typeMember) {
             var _attributeArray = this._attributeArray;
             var index = MACROUTILS.getIdFromTypeMember(typeMember);
             if (index === undefined) return undefined;
-            return _attributeArray[index] ? _attributeArray[index].globalDefault : undefined;
+            return _attributeArray[index] ? _attributeArray[index]._globalDefault : undefined;
```
cedricpinson: same here _attributeArray attributeArray

sources/osg/State.js
Kuranes: reverted the change. too many things at all.

cedricpinson:
replaced by #847

PR #843, opened 2017-08-10
Fixes non filtered stats

Kuranes:


Kuranes:
Sorry for that.

FYI => learned the hard way that firefox devtools open doesn't fill the browser cache, so if you open a new tab without dev tools, it opens an OLD page even if you loaded the new page in the tab with devtools open

PR #847, opened 2017-08-13
Removes stack's allocations in State

cedricpinson:
tried on both chrome and firefox.
?stats=1&deep=3&complexStateSet=true
Chrome => no difference whatsoever.
Firefox => small difference. 130-150ms before vs 90-120ms after

same PR #842 just cleaned a bit


cedricpinson:
The only worry I have is that probably on pagedlod items we would need to clear stateSet reference from and stateAttribute reference from pool stack, otherwise it will not be de allocated from the browser

jtorresfabra:
If I get it right, the problem would be that pagedLOD children StateAttributes would be always in the Stack. 

DatabasePager as it is now only release GL objects via ReleaseVisitor:
https://github.com/cedricpinson/osgjs/blob/master/sources/osgDB/DatabasePager.js#L368

The rest of information of the childs is removed using a node.removeChildren(), so I think the problem resides in the side of Node more than in PagedLOD front. So when a Node removes its children we should remove also StateSet/StateAttributes about them.

cedricpinson:
Even if we remove StateAttribute in StateSet of node it will be referenced in State in stack polled so. The only thing we will need would be to clear 'sometime' all content of stack so release javacscript StateAttribute. But it's maybe not a big problem because gl resources are released correctly

jtorresfabra:
> The rest of information of the childs is removed using a node.removeChildren(), so I think the problem resides in the side of Node more than in PagedLOD front. So when a Node removes its children we should remove also StateSet/StateAttributes about them.

I meant also from stack.

PR #850, opened 2017-08-21
Improves gc

cedricpinson:
Reuses objects for each frame


sources/osg/CullVisitor.js
```diff
@@ -375,49 +378,21 @@ MACROUTILS.createPrototypeObject(
...
-                for (var i = 0, j = this._reserveCullSettingsStackCurrent; i <= j; i++) {
-                    this._reserveCullSettingsStack[i].reset();
+                if (!this._pooledRenderStages[type]) {
+                    var classCtor = !classInstance ? RenderStage : classInstance.constructor;
+                    var createRenderStage = function() {
+                        return new classCtor();
+                    };
+                    this._pooledRenderStages[type] = new PooledResource(createRenderStage);
                 }
+                return this._pooledRenderStages[type].getOrCreateObject().init();
```
jtorresfabra: I think it could be better if the PooledResource always return clean Objects when getOrCreate(). 
Is there any case when we don't want a clean/reseted object? 
Simply forcing to every poolable object to have a reset method does the trick to call reset in getOrCreateObject(), and removes the hack of doing twice init() things.

Otherwise we should state very clear in the documentation that returned objects are not clean, and user is responsible to reset/init them

cedricpinson: agree but for performance reason it's the user who needs to clean the object he gets
I will add a comment in the source code

PR #853, opened 2017-08-28
Avoids memory alloc using private funcs

jtorresfabra:


sources/osgDB/DatabasePager.js
```diff
@@ -431,6 +425,21 @@ MACROUTILS.createPrototypeObject(
             });
             availableTime -= elapsedTime;
             return availableTime;
+        },
+
+        // Helper functions to avoid undesired memory allocation
+        _sortByTimeStamp: function(r1, r2) {
```
cedricpinson: thoses functions could even be outside of the object, unless the user could overwrite it. It could be full static function

jtorresfabra: yeah the point is letting the user to implement it's own priority functions

jtorresfabra:
Fixed

PR #857, opened 2017-08-30
Optimizes hammer listeners

Kuranes:
disables unecessary recoginzers
remove unecessary checks

### DIFF
--- a/sources/osgGA/OrbitManipulatorHammerController.js
+++ b/sources/osgGA/OrbitManipulatorHammerController.js
@@ -23,11 +23,67 @@ OrbitManipulatorHammerController.prototype = {
         this._transformStarted = false;
         this._dragStarted = false;
 
+        this._cbPanStart = this.panStart.bind(this);
+        this._cbPanMove = this.panMove.bind(this);
+        this._cbPanEnd = this.panEnd.bind(this);
+
+        this._cbPinchStart = this.pinchStart.bind(this);
+        this._cbPinchEnd = this.pinchEnd.bind(this);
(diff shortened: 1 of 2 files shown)


sources/osgGA/OrbitManipulatorHammerController.js
```diff
@@ -56,29 +60,36 @@ OrbitManipulatorHammerController.prototype = {
...
+            return;
+        }
+        this._eventProxy = hammer;
+        if (this._manipulator) this.setListeners();
+    },
+
     _computeTouches: function(event) {
+        // should not happen as we test mouse each time ?
         if (event.pointers !== undefined) return event.pointers.length;
+        console.log('compute touches mouse');
```
Kuranes: never ever got into this log...

cedricpinson: hammer was also used on touchscreen not only mobile, it's possible it's an old stuff from the time touch was touch api was not stable.
If the code is useless, then remove it, be sure to test it on different device

sources/osgGA/OrbitManipulatorHammerController.js
```diff
@@ -30,18 +30,22 @@ OrbitManipulatorHammerController.prototype = {
         this._isValid = valid;
     },
 
-    setEventProxy: function(hammer) {
-        if (!hammer || hammer === this._eventProxy) {
-            return;
-        }
+    setListeners: function() {
```
cedricpinson: if it's almost for internal use, then it should be named _setListeners

PR #861, opened 2017-09-05
Reuses controllers when initializing manipulators

rafagaitan:
…ng multiple times the events of the controllers

These fixes also allow the users of the manipulators to override the event callbacks, otherwise new callbacks were created and some inconsistencies could happen.

### DIFF
--- a/sources/osgGA/CADManipulator.js
+++ b/sources/osgGA/CADManipulator.js
@@ -155,7 +155,11 @@ MACROUTILS.createPrototypeObject(
 
             CADManipulator.ControllerList.forEach(function(value) {
                 if (CADManipulator[value] !== undefined) {
-                    self._controllerList[value] = new CADManipulator[value](self);
+                    if (self._controllerList[value]) {
+                        self._controllerList[value].init();
+                    } else {
+                        self._controllerList[value] = new CADManipulator[value](self);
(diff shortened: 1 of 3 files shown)


rafagaitan:
Ops, I didn't know that! Do you want me to change it?

BTW, I feel we could move to a common part the registering process, since it's the same code in the three manipulators, maybe a function in the base class? what do you think?

stephomi:
Also, our commit message always starts with 3rd person so e.g `Reuses controllers when initializing manipulators`.
I'm not sure how much we want to enforce these 2 rules on osgjs (most of the commiters on osgjs work at sketchfab where we follow these rules so we simply apply them back in osgjs).
It's up to @cedricpinson , but if we want to enforce them we should probably add it to the README.

As for sharing code, sure.
I'm not sure what the `AvailableControllerList` is used for too, it's identical to `AvailableControllerList`, maybe we can remove it too.

cedricpinson:
@rafagaitan yes please could cleanup commit message ? thx

rafagaitan:
I've cleaned up the commit message. The changes remain the same.

I will try to take a look to the code and avoid code duplications on another pull request, I need to study it more carefully, but for now these fixes some problems we had in our application.

PR #862, opened 2017-09-07
Improves PostProcess example

stephomi:
- Preload postprocess json and glsl so that we don't have to drag n drop files by default
- integrate ssao inside the new example


examples/postprocess/main.js
```diff
@@ -238,12 +479,20 @@
             this._gui = new window.dat.GUI();
 
             var cb = this._gui.add(this._params, 'composer', this._composerConfigNames).listen();
-            var self = this;
```
cedricpinson: You removed this line and it broke colorCorrection. add bind(this) in colorCorrection callback or keep a self

stephomi:
fixed

PR #863, opened 2017-09-11
Makes stacked name node optional

stephomi:


sources/osgWrappers/serializers/osgAnimation.js
```diff
@@ -351,7 +351,7 @@ osgAnimationWrapper.UpdateMatrixTransform = function(input, umt) {
 
 osgAnimationWrapper.StackedTranslate = function(input, st) {
     var jsonObj = input.getJSON();
-    if (!jsonObj.Name || !jsonObj.Translate) return P.reject();
+    if (!jsonObj.Translate) return P.reject();
```
cedricpinson: it requires a warning because it's very hard to detect same for other.

PR #864, opened 2017-09-13
Adds webglRestore support

Kuranes:
if lostcontext user callback return true we try to restore,
if user do need to re-provide data, he has to do by using a callback on their GLObject

### DIFF
--- a/examples/performance/main.js
+++ b/examples/performance/main.js
@@ -53,7 +53,19 @@
                 that._viewer
                     .getCamera()
                     .setEnableFrustumCulling(!that._viewer.getCamera().getEnableFrustumCulling());
-            }
+            },
+            loseContext: function() {
+                var gl = this._viewer.getGraphicContext();
+                var ext = gl.getExtension('WEBGL_lose_context');
+                if (!ext) {
+                    osg.log('missing WEBGL_lose_context extension');
+                    return;
+                }
+                ext.loseContext();
+                window.setTimeout(function() {
+                    ext.restoreContext(gl);
+                }, 0);
+            }.bind(this)
         };
 
         this._mediaPath = '';
@@ -248,6 +260,7 @@
             //button
             this._gui.add(this._config, 'update');
             this._gui.add(this._config, 'frustumCulling');
+            this._gui.add(this._config, 'loseContext');
             //this._gui.add( this._config, 'postProc' );
             //this._gui.add( this._config, 'scissor' );
 
--- a/sources/osg/BufferArray.js
+++ b/sources/osg/BufferArray.js
@@ -102,6 +102,12 @@ BufferArray.flushAllDeletedGLBufferArrays = function(gl) {
     }
 };
 
+BufferArray.onLostContext = function(gl) {
+    if (!BufferArray._sDeletedGLBufferArrayCache.has(gl)) return;
+    var deleteList = BufferArray._sDeletedGLBufferArrayCache.get(gl);
+    deleteList.length = 0;
+};
+
 MACROUTILS.createPrototypeObject(
     BufferArray,
     MACROUTILS.objectInherit(GLObject.prototype, {
@@ -124,11 +130,17 @@ MACROUTILS.createPrototypeObject(
             return false;
         },
 
+        invalidate: function() {
+            this._buffer = undefined;
+            this.dirty();
+        },
+
         releaseGLObjects: function() {
             if (this._buffer !== undefined && this._buffer !== null && this._gl !== undefined) {
                 BufferArray.deleteGLBufferArray(this._gl, this._buffer);
+                GLObject.removeObject(this._gl, this);
             }
-            this._buffer = undefined;
+            this.invalidate();
         },
 
         setNormalize: function(normalize) {
--- a/sources/osg/FrameBufferObject.js
+++ b/sources/osg/FrameBufferObject.js
@@ -17,6 +17,7 @@ var FrameBufferObject = function() {
     this._fbo = undefined;
     this._rbo = undefined;
     this._attachments = [];
+    this._buffers = [];
     this._dirty = true;
     this._hasMRT = WebglCaps.instance().getWebGLExtension('WEBGL_draw_buffers');
 };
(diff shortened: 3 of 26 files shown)


Kuranes:
todo:
- stats disappears
- test about the "renderer" by camera
- demo with callback

Kuranes:
TODO:
- make it Default to restore the context
- samples testing using url option ?SimulateWebGLLostContext=4040 (4040 is the number of calls where the webglcontext will be lost)
- add button on pagedlod and performance only samples

Kuranes:
Done:

make it Default to restore the context
samples testing using url option ?SimulateWebGLLostContext=4040 (4040 is the number of calls where the webglcontext will be lost)
add button on pagedlod and performance only samples

sources/osg/VertexArrayObject.js
```diff
@@ -0,0 +1,112 @@
...
+
+// static method to delete Program
+VertexArrayObject.deleteGLVertexArrayObject = function(gl, buffer) {
+    if (!VertexArrayObject._sDeletedGLVertexArrayObjectCache.has(gl))
+        VertexArrayObject._sDeletedGLVertexArrayObjectCache.set(gl, []);
+    VertexArrayObject._sDeletedGLVertexArrayObjectCache.get(gl).push(buffer);
+};
+
+// static method to flush all the cached glPrograms which need to be deleted in the GL context specified
+VertexArrayObject.flushDeletedGLVertexArrayObjects = function(gl, availableTime) {
```
jtorresfabra: You need to add the call here https://github.com/cedricpinson/osgjs/blob/master/sources/osgViewer/View.js#L318. Otherwise memory is not being released.

sources/osg/VertexArrayObject.js
```diff
@@ -0,0 +1,112 @@
...
+    var numBuffers = deleteList.length;
+    for (var i = numBuffers - 1; i >= 0 && elapsedTime < availableTime; i--) {
+        gl.deleteVertexArray(deleteList[i]);
+        deleteList.splice(i, 1);
+        elapsedTime = Timer.instance().deltaS(beginTime, Timer.instance().tick());
+    }
+    return availableTime - elapsedTime;
+};
+
+VertexArrayObject.flushAllDeletedGLVertexArrayObjects = function(gl) {
```
jtorresfabra: Same as above but here https://github.com/cedricpinson/osgjs/blob/master/sources/osgViewer/View.js#L330

sources/osg/FrameBufferObject.js
```diff
@@ -43,12 +44,13 @@ FrameBufferObject.flushDeletedGLFrameBuffers = function(gl, availableTime) {
     if (availableTime <= 0.0) return availableTime;
 
     if (!FrameBufferObject._sDeletedGLFrameBufferCache.has(gl)) return availableTime;
+    var deleteList = FrameBufferObject._sDeletedGLFrameBufferCache.get(gl);
+    if (deleteList.length === 0) return availableTime;
 
     var elapsedTime = 0.0;
     var beginTime = Timer.instance().tick();
-    var deleteList = FrameBufferObject._sDeletedGLFrameBufferCache.get(gl);
+    deleteList = MACROUTILS.arrayUniq(deleteList);
```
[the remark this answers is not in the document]
Kuranes: Do you mean for FBO only or all the gl resource delete list ?
Would make sense to me to use Set to prevent multiple shared GL resources delete
Need perf test (pagedlod sample I think)
Can change.
@jtorresfabra any advice ?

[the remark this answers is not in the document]
jtorresfabra: @cedricpinson @Kuranes  I don't think we even need that. deleting something that have already been deleted has no effect. (for instance: https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/deleteFramebuffer ). So I don't think it is worth to change this.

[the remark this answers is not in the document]
Kuranes: Interesting didn't know, thanks, so I remove the uniq changes then and believe in the power of the spec ?
@cedricpinson ok with that ?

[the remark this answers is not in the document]
jtorresfabra: I did :), and it worked so far...

sources/osg/BufferArray.js
```diff
@@ -102,6 +102,12 @@ BufferArray.flushAllDeletedGLBufferArrays = function(gl) {
     }
 };
 
+BufferArray.lostContext = function(gl) {
```
Nehon: Naming here is not clear. lostContext doesn't sound like an action, if it's something called as a callback, maybe call it onContextLost, else, maybe call it, clearContext, discardContext or something like that.

sources/osg/BufferArray.js
```diff
@@ -124,11 +130,17 @@ MACROUTILS.createPrototypeObject(
             return false;
         },
 
+        invalidate: function() {
```
Nehon: invalidate sounds good to me. another option could be "clear", as it seems to discard the underlying data.

sources/osg/FrameBufferObject.js
```diff
@@ -127,7 +140,26 @@ MACROUTILS.createPrototypeStateAttribute(
                 return new FrameBufferObject();
             },
 
+            invalidate: function() {
```
Nehon: same remark as before, could be "clear", but invalidate sounds ok.

examples/performance/main.js
```diff
@@ -53,7 +53,19 @@
...
+                var gl = this._viewer.getGraphicContext();
+                var ext = gl.getExtension('WEBGL_lose_context');
+                if (!ext) {
+                    osg.log('missing WEBGL_lose_context extension');
+                    return;
+                }
+                ext.loseContext();
+                window.setTimeout(function() {
+                    ext.restoreContext(gl);
+                }, 0);
```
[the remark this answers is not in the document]
Kuranes: it makes sure the call is async (not in the middle of a frame)

sources/osg/BufferArray.js
Kuranes: ok for onLostContext then ?
("LostContext" is name of the event in webgl)

sources/osg/VertexArrayObject.js
Kuranes: Fixed

sources/osg/VertexArrayObject.js
Kuranes: fixed

PR #866, opened 2017-09-18
Fixes PolytopeIntersector function names

davehg:
The intersector function name was misspelled, causing two functions to have the same name.


jtorresfabra:
Tests are failing, but not due to this PR.

stephomi:
Yes my bad.

I also fixed it in https://github.com/cedricpinson/osgjs/pull/867, IE11 crashed because of duplicate names (as expected because of use strict).

However, ES6 browser will not fail because apparently it's valid in es6 syntax :(
https://stackoverflow.com/questions/29936845/use-strict-now-allows-duplicated-properties

PR #868, opened 2017-09-19
Removes old Composer

Kuranes:
removes deprecated Composer in favor of new one


examples/split-rendering-rtt/main.js
```diff
@@ -395,22 +396,38 @@
...
-            });
-
-            blendPass.getOrCreateStateSet().setTextureAttributeAndModes(1, transparent);
-            blendPass.getOrCreateStateSet().setTextureAttributeAndModes(0, opaque);
-            composer.addPass(blendPass);
-            composer.renderToScreen(this._canvas.width, this._canvas.height);
-
-            this._root.addChild(composer);
+            var blendRTT = true;
+            if (blendRTT) {
```
jtorresfabra: Always true? no need for the if/else block

Kuranes: was in the demo before, I think it's for debug or smth. 
didn't want change to change behavior

examples/split-rendering-rtt/main.js
```diff
@@ -3,17 +3,18 @@
...
     var osgUtil = OSG.osgUtil;
     var ExampleOSGJS = window.ExampleOSGJS;
 
     var composeFragmentShader = [
-        osgUtil.Composer.Filter.defaultFragmentShaderHeader,
-        'uniform sampler2D Texture1;',
-        'void main (void)',
+        'uniform sampler2D TextureOpaque;',
+        'uniform sampler2D TextureTransparent;',
+        'vec4 blend ()',
```
jtorresfabra: error here?

examples/frustum-culling/main.js
```diff
@@ -487,7 +488,7 @@ bs.getOrCreateStateSet().setTextureAttributeAndModes( 0, new osg.Texture(), osg.
                 'uniform mat4 uModelViewMatrix;',
                 'uniform mat4 uProjectionMatrix;',
                 'void main(void) {',
-                '  gl_Position = uProjectionMatrix * uModelViewMatrix * vec4(Vertex,1.0);',
+                '  gl_Position = uProjectionMatrix * (uModelViewMatrix * vec4(Vertex,1.0));',
```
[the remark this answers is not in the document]
Kuranes: Those parenthesis tranforms a  mat4*mat4 into a  mat4*vec4.
It's needed everywhere we does this to prevent further copy and past of the unoptimized code (mat4*mat4).

examples/split-rendering-rtt/main.js
jtorresfabra: @Kuranes I guess is better to remove it. If it does not make sense, better to get rid.

PR #869, opened 2017-09-19
Avoids sparse array in StateSet and State

cedricpinson:
It does not seems to have a big effect on performance sample even if it seems a bit faster on chrome and similar in firefox, but on safari it seems more interesting

this the timing of 'render' in stats
chrome:
3.6/3.7 (master)
3.3/3.45 (avoid-sparse-array)

firefox:
4.7/4.9 (master)
4.5/4.8 (avoid-sparse-array)

safari:
6.7/6.9 (master)
5.5/5.8 (avoid-sparse-array)

Also adds a better handling of Scissor

### DIFF
--- a/sources/osg/State.js
+++ b/sources/osg/State.js
@@ -391,14 +391,20 @@ MACROUTILS.createPrototypeObject(
                 _attributeArray.length > stateSetAttributeArray.length
                     ? _attributeArray.length
                     : stateSetAttributeArray.length;
+
             for (var i = 0, l = max; i < l; i++) {
                 var attribute;
                 var attributeId = i;
-                var attributeStack = _attributeArray[attributeId];
 
-                var stateSetAttributePair = stateSetAttributeArray[attributeId];
+                var attributeStack =
+                    attributeId < _attributeArray.length ? _attributeArray[attributeId] : undefined;
 
-                var hasStateAttributeStack = attributeStack !== undefined;
+                var stateSetAttributePair =
+                    attributeId < stateSetAttributeArray.length
+                        ? stateSetAttributeArray[attributeId]
+                        : undefined;
+
+                var hasStateAttributeStack = !!attributeStack;
                 var hasStateAttributeStackChanged =
                     hasStateAttributeStack && attributeStack._changed;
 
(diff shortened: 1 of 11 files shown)


sources/osg/State.js
```diff
@@ -480,11 +490,18 @@ MACROUTILS.createPrototypeObject(
...
                 for (var j = 0; j < lt; j++) {
                     var attributeId = j;
-                    var attributeStack = _textureAttributeArray[attributeId];
-                    var stateSetAttributePair = stateSetTextureAttributeArray
-                        ? stateSetTextureAttributeArray[attributeId]
-                        : undefined;
-                    var hasStateAttributeStack = attributeStack !== undefined;
+
+                    var attributeStack =
+                        attributeId < _textureAttributeArray.length
```
Kuranes: tested caching _textureAttributeArray.length and stateSetTextureAttributeArray.length as var outside of the loop ?

sources/osg/State.js
```diff
@@ -803,58 +822,65 @@ MACROUTILS.createPrototypeObject(
...
 
         pushAttributeMap: function(_attributeArray, stateSetAttributeArray, validAttributeArray) {
             /*jshint bitwise: false */
             var attributeStack;
-
             for (var i = 0, l = validAttributeArray.length; i < l; i++) {
                 var index = validAttributeArray[i];
-                var attributePair = stateSetAttributeArray[index];
+                var attributePair =
+                    index < stateSetAttributeArray.length
```
Kuranes: ( same note about caching length )

sources/osgShader/ShaderGenerator.js
```diff
@@ -100,7 +101,7 @@ ShaderGenerator.prototype = {
 
             for (var i = 0; i < cacheType.length; i++) {
                 var type = cacheType[i];
-                var attributeStack = textureUnit[type];
+                var attributeStack = type < textureUnit.length ? textureUnit[type] : undefined;
```
Kuranes: (same note about textureUnit.length and _attributeArrayForUnit.length )

cedricpinson: fixed

sources/osg/State.js
cedricpinson: fixed

sources/osg/State.js
cedricpinson: fixed

PR #870, opened 2017-09-21
Caches Blend/Depth/CullFace/ColorMask gl state

cedricpinson:
Stats on differents examples
shadowmap:
before 446 gl commands
after 405 gl commands

PBR example with model lieutenant:
before 329
after 308

performance with complex stateset:
before 13345
after 10262

sketchfab https://sketchfab.com/models/49f9821876b744fab3453048185c04b6?stats=1
before 4011
after 2169

Mobile on https://sketchfab.com/models/2afd4091d395418c9107b84ad5b09cf2?stats=1 I get from
90-100ms on glframe w/o cache => final 8fps
60-70ms on glrame with cache => final 10fps

### DIFF
--- a/sources/osg/StateCache.js
+++ b/sources/osg/StateCache.js
@@ -1,0 +1,598 @@
+'use strict';
+
+var Depth = require('osg/Depth');
+var BlendFunc = require('osg/BlendFunc');
+var CullFace = require('osg/CullFace');
+
+var createStateBlendFunc = function() {
+    return {
+        buffer: {
+            separate: false,
+            enable: false,
+            sourceFactor: BlendFunc.ONE,
+            destinationFactor: BlendFunc.ZERO,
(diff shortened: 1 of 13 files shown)


cedricpinson:
@Kuranes @jtorresfabra it could be good to re review, I changed some stuffs

sources/osg/StateCache.js
```diff
@@ -0,0 +1,598 @@
...
+};
+
+StateCache.prototype = {
+    clearColor: function(array) {
+        var data = this._stateClearColor;
+        var buffer = data.buffer;
+        var state = data.state;
+        data.changed = false;
+
+        if (
```
Kuranes: note that for for multi boolean and/or even four 255 value, you could use bit mask.
One value comparisiong would make it fast.
Attribute could store bitmask and value, as bitmask usually setted not often, but comparison are done a lot of time.
(valable for state without float/integer values)

PR #872, opened 2017-10-02
Improves a bit promise warnings

stephomi:
- reject a promise with a reason to avoid "undefined" reason
- Removes new P() when we can simply use chaining instead

### DIFF
--- a/sources/osgWrappers/serializers/osg.js
+++ b/sources/osgWrappers/serializers/osg.js
@@ -1,6 +1,15 @@
 'use strict';
 var P = require('bluebird');
+var Notify = require('osg/notify');
 var TransformEnums = require('osg/transformEnums');
+
+var rejectObject = function(msg, node) {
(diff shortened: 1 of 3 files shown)


stephomi:
Note sure where to put the helper `rejectObject` (for now copy-pasted in 3 files)

sources/osgWrappers/serializers/osg.js
```diff
@@ -1,7 +1,16 @@
 'use strict';
 var P = require('bluebird');
+var Notify = require('osg/notify');
```
jtorresfabra: If I recall @cedricpinson prefer notify with no capital letter at the beginning from other PR's.

jtorresfabra: About where to put the rejectObject I'm also not clear, but if you don't find a better place maybe you can add it to the osg/Utils hodgepodge.

PR #885, opened 2017-10-16
Updates prettier

stephomi:


cedricpinson:
Currently the configuration was used from the eslintrc.json in grunt for the commands
grunt prettier/check
https://github.com/cedricpinson/osgjs/blob/master/Gruntfile.js#L222-L239

So we needs to update config in eslintrc.json to use .prettierrc instead. We should not have two configuration in the project.

PR #890, opened 2017-10-20
Adds extra check on primitives before trying to render geometry

cedricpinson:


sources/osg/Geometry.js
```diff
@@ -341,10 +342,27 @@ utils.createPrototypeNode(
...
-
-            if (!this._primitives.length) return;
+            var invalidPrimitives = true;
+            if (this._primitives.length) {
+                invalidPrimitives = false;
+                for (var i = 0; i < this._primitives.length; i++) {
+                    var primitive = this._primitives[0];
+                    if (primitive instanceof DrawElements) {
+                        if (!primitive.getIndices() || !primitive.getIndices().isValid()) {
+                            invalidPrimitives = true;
```
stephomi: you don't need the invalidPrimitives variable, you can just issue the warning and return directly

cedricpinson: yes my bad

PR #893, opened 2017-10-24
Adds custom code shader node support

Kuranes:
demos it with 2 shader in demo shadow

examples/shadowmap?customShadow=debug
examples/shadowmap?customShadow=pcss

For now doesn't work *after* instanciation of nodes (here using the dat.gui list customShadow)

### DIFF
--- a/sources/osg/Program.js
+++ b/sources/osg/Program.js
@@ -214,6 +214,63 @@ utils.createPrototypeStateAttribute(
                 this._program = undefined;
             },
 
+            _rebuildProgramFromSpector: function(
+                vertexShaderText,
+                fragmentShaderText,
+                onCompiled,
+                onError
+            ) {
+                this._dirty = true;
+                this._vertex.invalidate();
+                this._fragment.invalidate();
+                this.invalidate();
+                this._vertex.setText(vertexShaderText);
+                this._fragment.setText(fragmentShaderText);
+                this._spectorOnCompiled = onCompiled;
+                this._spectorOnError = onError;
+            },
+
+            _onErrorToSpector: function(errLink) {
+                if (!this._spectorOnError) return false;
+                this._spectorOnError(errLink);
+                return true;
+            },
+
+            _onCompilationToSpector: function() {
+                if (!this._spectorOnCompiled) return;
+                this._spectorOnCompiled(this.program);
+            },
+
+            _bindProgramToSpector: function() {
+                if (!window || !window.spector || this._program.__SPECTOR_rebuildProgram) return;
+                this._program.__SPECTOR_rebuildProgram = this._rebuildProgramFromSpector.bind(this);
+            },
+
+            _logDebugShader: function(gl, errLink) {
+                if (errLink !== 'Failed to create D3D shaders.\n') return;
+                // rawgl trick is for webgl inspector
+                var debugShader = gl.rawgl !== undefined ? gl.rawgl : gl;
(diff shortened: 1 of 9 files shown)


sources/osgShader/utils.js
```diff
@@ -428,21 +428,35 @@ var createNode = function(res, fileName) {
     return NodeCustom;
 };
 
+var objOutGlobal = {};
```
cedricpinson: It seems a bit hacky the way to replace a node by another.

Why not having changing the node used by the compiler ?

```
    createShadowingLight: function(light, lighted) {
        // ....
        this.getNode('ShadowReceive').inputs(inputs).outputs(outputs).addDefines(defines);
        return shadowedOutput;
    },
```
Instead of having this.getNode('ShadowReceive') we could have in compiler
this._shadowReceive = 'osgShader::ShadowReceive';
and having using the following code
```
    createShadowingLight: function(light, lighted) {
        // ....
        this.getNode(this._shadowReceive).inputs(inputs).outputs(outputs).addDefines(defines);
        return shadowedOutput;
    },
```

And if you want to change the node you could do something like
```
Compiler.setShadowReceiveNode('CustomNameSpace::ShadowRecive');
```
And it would use this node registered in the factory

It's an idea but it seems cleaner to me even if it's probably more verbose

Kuranes: At Compiler level it's indeed much more verbose and less flexible.
( It means to code compiler(s), code the handle/check override for each override(s).  etc)

Unless we have need to replace a shader core node in one compiler only, it doesn't make much sense to code that now, as it's really another feature, not the one needed here (but we could do it if need arise, one doesn't prevent the other). 

"we want to change osgjs shader" => "replace the shader node"
!==
"we wnat to change that shader node in that compiler" => override getNode in the Compiler

Shader node are shared by all compiler, so replacing it a this level seems logical to me, as we want to replace it everywhere. (in all compiler)

sources/osg/Program.js
```diff
@@ -261,45 +318,27 @@ utils.createPrototypeStateAttribute(
...
-                                    debugShader.getTranslatedShaderSource(this._fragment.shader)
-                                );
-                            }
-
+                            if (this._onErrorToSpector(errLink)) return;
+                            this._logDebugShaders(gl, errLink);
                             compileClean = false;
+                        } else {
+                            this._onCompilationToSpector(this.program);
+                            // TODO: better usage of validate.
```
cedricpinson: This block of comment is useless, we could just remove it.

sources/osgShader/utils.js
cedricpinson: In this case it could be in prototype, I am not saying it's the best solution there is maybe better but what you did is create two nodes and merge partially one to the other, I am not comfortable with this approach.
I still think it makes sense to customize shader or change/replace a type of node.
I continue to think about it

cedricpinson:
@Nehon I added you to review, you could have some idea

sources/osgShader/utils.js
Kuranes: there is no merge, it's a total replace of the shader node.

Nehon: To me, it's ok. It provides an easy way to override a node in a global scope from a "user" perspective. 
The alternative would require more work on the user side, specially when several compilers are meant to work together in the same scene.
Also the implementation seems to follow the same paradigm as the original implementation, so I found it ok.

PR #894, opened 2017-10-25
Fixes VAO with non VAO geometry

cedricpinson:
When using Morph geometry with BufferProxy the geometry does not use vao
so when we had something like:
Geometry VAO
Geomtry Non-VAO

The second geometry was changing the VAO of the first geometry. The fix
disable VAO when drawing the second geometry.

### DIFF
--- a/sources/osg/Geometry.js
+++ b/sources/osg/Geometry.js
@@ -325,15 +325,13 @@ utils.createPrototypeNode(
             state.drawGeometry(this);
             var cachedDraw = this._cacheDrawCall[prgID];
 
-            if (this._useVAO) {
-                if (!this._vao[prgID]) {
-                    state.setVertexArrayObject(null);
-                } else {
-                    if (this._vao[prgID].isDirty()) {
-                        // need vertex array recreation
-                        // ie: lost context
-                        cachedDraw = undefined;
-                    }
+            if (!this._vao[prgID]) {
+                state.setVertexArrayObject(null);
+            } else {
+                if (this._vao[prgID].isDirty()) {
+                    // need vertex array recreation
+                    // ie: lost context
+                    cachedDraw = undefined;
                 }
             }
 


sources/osg/Geometry.js
```diff
@@ -325,15 +325,13 @@ utils.createPrototypeNode(
...
-            if (this._useVAO) {
-                if (!this._vao[prgID]) {
-                    state.setVertexArrayObject(null);
-                } else {
-                    if (this._vao[prgID].isDirty()) {
-                        // need vertex array recreation
-                        // ie: lost context
-                        cachedDraw = undefined;
-                    }
+            if (!this._vao[prgID]) {
```
Kuranes: doesn't it means that on GPU without VAO state will do a bindVertexArray and it will fail ?
 
https://github.com/cedricpinson/osgjs/blob/cd3737da30daeb63471129edea8009644bc7c60e/sources/osg/State.js#L1084

Kuranes: ha, perhaps not already call to that later in the same func https://github.com/cedricpinson/osgjs/blob/cd3737da30daeb63471129edea8009644bc7c60e/sources/osg/Geometry.js#L371

so drawGeometry is a drawGeemetryVAO or bindVertexArray(null) doesn't call VAO functions

stephomi: test 

https://github.com/cedricpinson/osgjs/blob/master/sources/osg/Geometry.js#L29

cedricpinson: @Kuranes no because this line https://github.com/cedricpinson/osgjs/blob/cd3737da30daeb63471129edea8009644bc7c60e/sources/osg/State.js#L1084 will never be reach. stats._currentVAO will be null so it will return at the first line

Kuranes: ha, ok.  so it's  implicit  noop. 
(one must not change behaviour of that test in state unaware that it's a noop in nonVAO case.)

cedricpinson: Yeah maybe we will add a comment

PR #897, opened 2017-10-26
Makes PCF an uniform instead of a Define

Kuranes:
Allow dynamic changes without recompilation

### DIFF
--- a/sources/osgShadow/ShadowReceiveAttribute.js
+++ b/sources/osgShadow/ShadowReceiveAttribute.js
@@ -25,8 +25,7 @@ var ShadowReceiveAttribute = function(lightNum, disable) {
     this._precision = 'UNSIGNED_BYTE';
     // kernel size & type for pcf
     this._kernelSizePCF = undefined;
-
-    this._fakePCF = true;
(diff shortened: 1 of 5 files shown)


sources/osgShadow/ShadowReceiveAttribute.js
```diff
@@ -167,6 +165,7 @@ utils.createPrototypeStateAttribute(
 
             uniformMap.normalBias.setFloat(this._normalBias);
             uniformMap.bias.setFloat(this._bias);
+            uniformMap.pcfKernel.setFloat(this._pcfKernel);
```
cedricpinson: Why not using int ?

Kuranes: float allow for 'fuzzy' pcf values like values computed from an algo if need be. (distance to blocker, size of light)  

As it's a uniform and int and float same size, it has no incidence whatsoever and yet more flexible

Kuranes:
I confirm it's tested.
but needs the corresponding PR to be taken advantage of.

PR #898, opened 2017-10-26
Fixes a cache issue in CacheState

cedricpinson:
Here the initial context for Cull StateAttribute
enable = true , mode = BACK

-> state.applyCullAttribute (enable = true, mode = back)
state.enable = true , state.mode = BACK ( buffer.enable = true, buffer.mode = back )
-> state.applyCullAttribute (enable = false, mode = back)
state.enable = true , state.mode = BACK ( buffer.enable = false, buffer.mode = back )
-> state.applyCullAttribute (enable = true, mode = front)
state.enable = true , state.mode = BACK ( buffer.enable = false, buffer.mode = front )

The code to check the update was like that:
```
if (state.enable !== enable) {
  buffer.enable = enable;
  data.changed = true;
}

if (!enable) return;

if (state.mode !== attribute._mode) {
  buffer.mode = attribute._mode;
  data.changed = true;
}
```
The logic was wrong because if new state is not different from last state
applied then buffer.enable was not updated. The problem was if mode was
different then buffer.enable did not have the good value. So the general logic
should be is something change, then update all buffer field to be sure to be up
to date when applying it to gl.

### DIFF
--- a/sources/osg/StateCache.js
+++ b/sources/osg/StateCache.js
@@ -2,6 +2,24 @@ import Depth from 'osg/Depth';
 import BlendFunc from 'osg/BlendFunc';
 import CullFace from 'osg/CullFace';
 
+// To avoid to apply duplicate states we cache them and apply them only
+// when there is a draw commands it means when calling state.drawGeometry
+// called from Geometry.drawImplementation or state.clear
+
+// Here what it does:
+// state.applyCullFaceAttribute(cullFaceDisableAttribute);
+// state.applyCullFaceAttribute(cullFaceEnableAttribute);
+// state.applyCullFaceAttribute(cullFaceDisableAttribute);
+// state.applyCullFaceAttribute(cullFaceEnableAttribute);
+// state.drawGeometry()
+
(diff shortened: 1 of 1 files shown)


sources/osg/StateCache.js
```diff
@@ -316,6 +316,7 @@ StateCache.prototype = {
             state.width !== attribute._width ||
             state.height !== attribute._height
         ) {
+            buffer.enable = enable;
```
Kuranes: not sure why it's needed as first condition line 306 should update it ?

Kuranes:
not easy to understand each stateCache member variable, and the fast it's applied only on draw call
perhaps an exlplanation over an abstract state cache would help
``` State caching, cache state, gl State and only applies if there's a draw
StateCache{
buffer: {} // comments what "state" (gpu, state or stateCache to be applied to GPU) 
state {}, // comments what "state" (gpu, state or stateCache to be applied to GPU) 
enabled : // enabled gpu/ cache / state
}
```

PR #903, opened 2017-11-09
Adds a function to reset the programs variables in GizmoGeometry

rtortajada:
Calling this function avoids problems when creating a new gizmo in a new glContext. The problem is that when a gizmo is created the programs in GizmoGeometry (which are global) are instantiated. After that if a new gizmo is created in a new glContext, the Gizmo Geometry tries to use the programs variables because they aren't undefined but it doesn't work since these programs are associated to the first glContext,

### DIFF
--- a/sources/osgUtil/gizmoGeometry.js
+++ b/sources/osgUtil/gizmoGeometry.js
@@ -124,6 +124,12 @@ var getOrCreateShaderQuadCircle = function() {
     return programQC;
 };
 
+var resetPrograms = function () {
+    program = undefined;
+    program2D = undefined;
+    programQC = undefined;
+};
+
 var createDebugLineGeometry = function() {
     var g = new Geometry();
     g.getAttributes().Vertex = new BufferArray(BufferArray.ARRAY_BUFFER, new Float32Array(4), 2);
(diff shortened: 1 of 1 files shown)


jtorresfabra:
Hi @rtortajada, 
Thanks for your first contribution. 

I feel you are adding a fix that is more a workaround than a fix. From my point of view the gizmo should not use global programs so the correct fix would be to avoid this. 

I suspect these global programs are there for cache, so one idea could be create a registry [context, programs] as when the NodeGizmo is created you have access to the context. If this becomes a problem other idea could be to cache these programs in the NodeGizmo, instead of doing it globally. 

So basically the point would be to avoid global programs. 

Thoughts? 
@cedricpinson @stephomi

cedricpinson:
@jtorresfabra Yeah caching in NodeGizmo could do the trick. But in the overall this questions is the same for other shape.

rtortajada:
Thanks for merging, I was waiting to have time to improve the solution..
I don't know if you want to 'reset' these vars in NodeGizmo and only reuse them for a unique NodeGizmo. Or if there's a way that allows reusing these vars for all the gizmos created until the context is changed.
Anyway, I see that there's a new issue created for that. :)

rtortajada:
Thanks for merging, I was waiting to have time to improve the solution..
I don't know if you want to 'reset' these vars in NodeGizmo and only reuse them for a unique NodeGizmo. Or if there's a way that allows reusing these vars for all the gizmos created until the context is changed.
Anyway, I see that there's a new issue created for that. :)

PR #904, opened 2017-11-09
Fixes shadow atlas for empty scene

Kuranes:
Handles empty scene shadowing using shadow atlas

### DIFF
--- a/sources/osgShadow/ShadowMap.js
+++ b/sources/osgShadow/ShadowMap.js
@@ -43,6 +43,11 @@ utils.createPrototypeObject(
             var near = nv.getComputedNear(),
                 far = nv.getComputedFar();
 
+            if (near === Infinity && far === -Infinity) {
+                this._shadowTechnique.markSceneAsNoShadow();
+                return false;
+            }
+
             mat4.getFrustumPlanes(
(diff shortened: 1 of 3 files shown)


cedricpinson:
It's not clear why we need to do something when we dont draw, could you have a look plz, maybe it's not needed ?

Kuranes:
- documented how/what/why empty scene are handled that way
- added case of empty scene
- renamed few method for clarity
- removed useless uniform change in the no scene case

sources/osgShadow/ShadowMap.js
```diff
@@ -43,14 +43,18 @@ utils.createPrototypeObject(
             var near = nv.getComputedNear(),
                 far = nv.getComputedFar();
 
+            if (near === Infinity && far === -Infinity) {
```
cedricpinson: are you sure it's && and not a ||

Kuranes: yes.
It checks for exact state  'no geom at all did pass the cullvisitor cull'.
So it check the exact state  CullVisitor resets near and far to before doing a cull. if either is not '===' it would mean something has passed the test and a geom is drawn.

PR #907, opened 2017-11-11
Adds Async Shader support

Kuranes:
Idea of the feature: 
----------------------
- Nukes ALL browser **FREEZE** during shader compilation
( see http://toji.github.io/shader-perf/ for basic demo)

How it's done:
----------------
- program apply is done in a 1st step (compile/link  but NO compilestate && attribute/uniform location calls 
- while compilation not yet finished, we used a "basic shader program" (shared for everyone) 
- when timer (0.8s) is done, we do the 2nd stage of Program (compile state/error and attribute/uniform location)
- we then apply the compile progam

Problem
---------
- ALL program are async... not good 

What it needs to be finished:
--------------------------------
- adds a "Async" attribute on stateset you want async.
- adds a "Replace" attribute on stateset where you just changed the shader (which means you can still use the previous shader of the stateset instead of the "Basic Shader" program
- device attributes parameter (time/delay  of the async, pointer to async program, etc ?)

*Is the attribute the good way to do that ?*
(what other way we could do that ?)

Further Todo:
---------------
- fragment/vertex shader compilation could also be async, smaller timer
 (that would make for a 3 step async. VS (0.05s)/FS(0.1s)/LINK(0.7)=> final)

### DIFF
--- a/sources/osg/Program.js
+++ b/sources/osg/Program.js
@@ -3,11 +3,38 @@ import notify from 'osg/notify';
 import GLObject from 'osg/GLObject';
 import StateAttribute from 'osg/StateAttribute';
 import ShaderProcessor from 'osgShader/ShaderProcessor';
+import Shader from 'osg/Shader';
 import Timer from 'osg/Timer';
 
 // singleton
 var sp = new ShaderProcessor();
 var errorCallback;
+
+var uniformRegexp = /uniform\s+\w+\s+(\w+)((\s)?\[(.*?)\])?/g;
+var attributesRegexp = /(in|attribute)\s+\w+\s+(\w+)\s*;/g;
+
+var getAttributeList = function(shaderText) {
+    var attributeMap = {};
+    var r, attr;
+    while ((r = attributesRegexp.exec(shaderText)) !== null) {
+        attr = r[2];
+        attributeMap[attr] = true;
+    }
+
+    return attributeMap;
+};
+
+var getUniformList = function(shaderText) {
+    var uniformMap = {};
+    var r;
+    while ((r = uniformRegexp.exec(shaderText)) !== null) {
+        var uniform = r[1];
+        uniformMap[uniform] = true;
+    }
+    return uniformMap;
+};
+
+var limitSec = 1;
 
 /**
  * Program encapsulate an vertex and fragment shader
@@ -30,37 +57,18 @@ var Program = function(vShader, fShader) {
 
     this._uniformsCache = undefined;
     this._attributesCache = undefined;
+    // state caches
     this._activeUniforms = undefined;
     this._foreignUniforms = undefined;
     this._trackAttributes = undefined;
+    // async compilations
+    this._asyncCompilation = undefined;
 
     if (vShader) this.setVertexShader(vShader);
 
     if (fShader) this.setFragmentShader(fShader);
 
     this._dirty = true;
-};
-
-var attributesRegexp = /(in|attribute)\s+\w+\s+(\w+)\s*;/g;
-var getAttributeList = function(shaderText) {
-    var attributeMap = {};
-    var r, attr;
-    while ((r = attributesRegexp.exec(shaderText)) !== null) {
-        attr = r[2];
-        attributeMap[attr] = true;
-    }
-
-    return attributeMap;
-};
-var uniformRegexp = /uniform\s+\w+\s+(\w+)((\s)?\[(.*?)\])?/g;
-var getUniformList = function(shaderText) {
-    var uniformMap = {};
-    var r;
-    while ((r = uniformRegexp.exec(shaderText)) !== null) {
-        var uniform = r[1];
-        uniformMap[uniform] = true;
-    }
-    return uniformMap;
 };
 
 // static cache of glPrograms flagged for deletion, which will actually
@@ -248,7 +256,7 @@ utils.createPrototypeStateAttribute(
                 this._program.__SPECTOR_rebuildProgram = this._rebuildProgramFromSpector.bind(this);
             },
 
-            _logDebugShader: function(gl, errLink) {
+            _logDebugShaders: function(gl, errLink) {
                 if (errLink !== 'Failed to create D3D shaders.\n') return;
                 // rawgl trick is for webgl inspector
                 var debugShader = gl.rawgl !== undefined ? gl.rawgl : gl;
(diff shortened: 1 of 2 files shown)


sources/osg/Program.js
```diff
@@ -263,105 +271,170 @@ utils.createPrototypeStateAttribute(
...
-                    var attributeArray = window.Object.keys(attributeMap);
-                    var uniformMap = getUniformList(this._vertex.text + '\n' + this._fragment.text);
-                    var uniformArray = window.Object.keys(uniformMap);
+                this._shaderName = shaderName ? shaderName[1] : '';
 
-                    if (compileClean) {
-                        this._program = gl.createProgram();
+                this._start = Timer.instance().tick();
+                this._Accum = 0.0;
+                console.log('Start compile ' + this._shaderName);
```
jtorresfabra: same as above

sources/osg/Program.js
```diff
@@ -3,12 +3,39 @@ import notify from 'osg/notify';
...
+    var uniformMap = {};
+    var r;
+    while ((r = uniformRegexp.exec(shaderText)) !== null) {
+        var uniform = r[1];
+        uniformMap[uniform] = true;
+    }
+    return uniformMap;
+};
+
+var limitSec = 1;
```
jtorresfabra: not a fan of having this global. Also would be nice to have it configurable.

sources/osg/Program.js
```diff
@@ -263,105 +271,170 @@ utils.createPrototypeStateAttribute(
...
+            _glAttachAndCompile: function(gl, programGL, vertexShader, fragmentShader) {
+                gl.attachShader(programGL, vertexShader.shader);
+                gl.attachShader(programGL, fragmentShader.shader);
+                utils.timeStamp('osgjs.metrics:linkShader');
+                gl.linkProgram(programGL);
+            },
 
-                if (!this._gl) {
-                    this.setGraphicContext(state.getGraphicContext());
+            timeAndStart: function(msg) {
```
jtorresfabra: Function name seems a bit weird to me, also avoid console.log

sources/osg/Program.js
```diff
@@ -263,105 +271,170 @@ utils.createPrototypeStateAttribute(
...
-                    var attributeMap = getAttributeList(this._vertex.getText());
-                    var attributeArray = window.Object.keys(attributeMap);
-                    var uniformMap = getUniformList(this._vertex.text + '\n' + this._fragment.text);
-                    var uniformArray = window.Object.keys(uniformMap);
+                this._shaderName = shaderName ? shaderName[1] : '';
 
-                    if (compileClean) {
-                        this._program = gl.createProgram();
+                this._start = Timer.instance().tick();
+                this._Accum = 0.0;
```
jtorresfabra: Better to avoid variable names starting by a capital letter. Also declare it in the constructor.

sources/osg/Program.js
```diff
@@ -263,105 +271,170 @@ utils.createPrototypeStateAttribute(
...
-                    }
+                if (!this._fragment.shader) {
+                    compileClean = this._fragment.compile(gl);
+                }
+
+                this._attributeMap = getAttributeList(this._vertex.getText());
+
+                if (!compileClean) {
+                    // Any error, Any
+                    // Pink must die.
```
jtorresfabra: :see_no_evil: sad...

sources/osg/Program.js
```diff
@@ -263,105 +271,170 @@ utils.createPrototypeStateAttribute(
...
+                this._uniformsCache = {};
+                this._attributesCache = {};
+
+                this.timeAndStart(' end link/check ');
+
+                this.cacheAttributeList(gl, window.Object.keys(this._attributeMap));
+                this.cacheUniformList(gl, window.Object.keys(this._uniformMap));
+
+                this.timeAndStart(' end link location ');
+                console.log('end gl COMPILATION ' + this._shaderName + ': ' + this._Accum + ' ms');
```
jtorresfabra: avoid console

sources/osg/Program.js
```diff
@@ -263,105 +271,170 @@ utils.createPrototypeStateAttribute(
...
+                    // async shader compilation part
+                    this._compileClean = this.compile(state);
+                    // async compilation threshold ?
+                    // strlen of vertex/fragment
+                    // number of uniform
+                    this._asyncCompilation = Timer.instance().tick();
+                    return;
+                } else if (this._asyncCompilation) {
+                    // still waiting
+                    if (Timer.instance().tick() - this._asyncCompilation > 800) {
```
jtorresfabra: Not a fan of hardcoded values. Hard to find later. Also in any case should be 800.0 I guess

sources/osgShader/ShaderGenerator.js
```diff
@@ -237,6 +237,27 @@ ShaderGenerator.prototype = {
...
+        var program = new Program(
+            new Shader(Shader.VERTEX_SHADER, vsDefault),
+            new Shader(Shader.FRAGMENT_SHADER, fsDefault)
+        );
+
+        this.getActiveAttributeList(state, attributes);
+        program.setActiveUniforms(this.getActiveUniforms(state, attributes, textureAttributes));
+        program.generated = true;
+        program.apply(state);
+        program._asyncCompilation = 1000.0;
```
jtorresfabra: Again the think on hardcoded values.

sources/osgShader/ShaderGenerator.js
```diff
@@ -251,9 +272,62 @@ ShaderGenerator.prototype = {
...
-            if (cache !== undefined) return cache;
+            var cachedProgram = this._cache[hash];
+            if (cachedProgram !== undefined && !cachedProgram.getAsyncCompiling()) {
+                return cachedProgram;
+            }
 
+            var waitingProgram;
+            var asyncShader = true;
+            var synchronizedCompilation = !asyncShader;
+            if (!synchronizedCompilation) {
```
jtorresfabra: not sure I get the tree lines above, so !synchronizedCompilation is always true?

sources/osg/Program.js
Kuranes: ha yes, all timing code not intended to be in final PR
( debug/bench code )

sources/osgShader/ShaderGenerator.js
Kuranes: it's a global for debug, comparing both async and sync.

Kuranes:
@jtorresfabra I flagged prototype feedback, so it's less than a review, more of an API / architexture check around the idea, related to PR Comment above.

"prototype"+"need feedback" is more "Idea tested and validated, now must think about serious API" 
(show code that does the thing, some numbers and then ask Help)

My main question is about making the thing APIsable.
see first comment notably
```
- adds a "Async" attribute on stateset you want async.
- adds a "Replace" attribute on stateset where you just changed the shader (which means you can still use the previous shader of the stateset instead of the "Basic Shader" program
- device attributes parameter (time/delay  of the async, pointer to async program, etc ?)
```

PR #908, opened 2017-11-14
Clean up gltf reader and adds beta animation support

cedricpinson:


sources/osgPlugins/ReaderWriterGLTF.js
```diff
@@ -39,15 +45,27 @@ var ReaderWriterGLTF = function() {
...
     this._inputImgReader = undefined;
     this._localPath = '';
 
+    this._extensions = [];
+
+    this._nodeAnimationTypes = undefined;
     this.init();
 };
 
+function base64ToArrayBuffer(base64) {
```
jtorresfabra: could be also static in FileHelper?

sources/osgDB/FileHelper.js
```diff
@@ -1,29 +1,232 @@
...
+
+    promise.finally(function() {
+        window.URL.revokeObjectURL(privateURL);
+    });
+    return promise;
+};
+
+var createArrayBufferFromBlob = function(blob) {
+    return new P(function(resolve, reject) {
+        var fr = new window.FileReader();
```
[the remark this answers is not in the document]
cedricpinson: Yeah agree but it's probably ok

sources/osgPlugins/ReaderWriterGLTF.js
cedricpinson: It was but because it's only used in gltf, I did not want to move it yet

PR #909, opened 2017-11-14
Splits OrbitManipulator limits computation for easier overriding

Nehon:
Separate pitch computation to be able to override it in sub classes.

### DIFF
--- a/sources/osgGA/OrbitManipulator.js
+++ b/sources/osgGA/OrbitManipulator.js
@@ -245,24 +245,26 @@ utils.createPrototypeObject(
             var rightDir = vec3.fromValues(1.0, 0.0, 0.0);
 
             return function(dx, dy) {
-                var pitch = Math.atan(-this._rotation[6] / this._rotation[5]) + dy / 10.0;
-                pitch = Math.min(Math.max(pitch, this._limitPitchDown), this._limitPitchUp);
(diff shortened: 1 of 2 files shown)


sources/osgGA/OrbitManipulator.js
```diff
@@ -245,8 +245,8 @@ utils.createPrototypeObject(
             var rightDir = vec3.fromValues(1.0, 0.0, 0.0);
 
             return function(dx, dy) {
-                var pitch = Math.atan(-this._rotation[6] / this._rotation[5]) + dy / 10.0;
-                pitch = Math.min(Math.max(pitch, this._limitPitchDown), this._limitPitchUp);
+                var prevPitch = Math.atan(-this._rotation[6] / this._rotation[5]);
+                var pitch = this._computePitch(prevPitch, dy);
```
cedricpinson: there is no bind(this) on this function returned

sources/osgGA/OrbitManipulator.js
```diff
@@ -261,6 +261,11 @@ utils.createPrototypeObject(
             };
         })(),
 
+        _computePitch: function(prevPitch, dy) {
```
Kuranes: name could be misleading ?
_computeYaw and _computePitch not in the same level (_computeYaw not called in code above,  not same functionnalities), but exact same name convention ?

PR #913, opened 2017-11-15
Improves how invalid primitives are handled

cedricpinson:
- Warn only the first time that a geometry has invalid primitives
- Discard only invalid primitives from geometry, if there is valid primitives
they will be displayed.


Kuranes:
very nitpick: _generatePrimitive but _getValidPrimitive**s**
(not consistent plural/singular)

cedricpinson:
@Kuranes @jtorresfabra fixed

jtorresfabra:
Good for me

PR #914, opened 2017-11-15
Fixes errors from reports 

Kuranes:
- Refactors so that program compilation can be done async  …
- Fixes exception throwed when just including osgjs but not using it.  …
- fixes manipulator instancied but still not added an eventProxy  …
- Fixes unclean compilation case
- Fixes strange test in webglcaps, adds comment on the extension apply

### DIFF
--- a/sources/osgViewer/eventProxy/StandardMouseKeyboard.js
+++ b/sources/osgViewer/eventProxy/StandardMouseKeyboard.js
@@ -1,4 +1,4 @@
-import {vec2} from 'osg/glMatrix';
+import { vec2 } from 'osg/glMatrix';
 
 var StandardMouseKeyboard = function(viewer) {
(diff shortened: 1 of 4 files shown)


sources/osgViewer/eventProxy/StandardMouseKeyboard.js
```diff
@@ -220,6 +220,7 @@ StandardMouseKeyboard.prototype = {
         }
 
         this.getManipulatorController().setEventProxy(this);
+        this.addEventListeners();
```
cedricpinson: I am not sure you want to call addEventlisteners at each update call. It's called once per frame. You should probably check if it's set or not

Kuranes:
@cedricpinson repushed with flag on eventbinding

PR #920, opened 2017-11-22
Improves box vs sphere comparison

stephomi:


sources/osgGA/Manipulator.js
```diff
@@ -40,37 +40,41 @@ Manipulator.prototype = {
...
-                var boundSphere = node.getBound();
-                if (boundSphere.radius() < bs.radius()) return boundSphere;
+                var bsphere = node.getBoundingSphere();
+                if (bsphere.valid() && bsphere.volume() < box.volume()) {
+                    return bsphere;
+                }
             }
 
-            return bs;
+            // return box
```
Kuranes: strange comment ?
// returns box as a sphere

function name is not that clear that it must return a boundingsphere, perhaps add it either in name (getHomeBoundingSphere) or function header comment.

stephomi: yes as a sphere

stephomi: removed the comments

PR #921, opened 2017-11-26
Improves stats

cedricpinson:
Doing this improvement trigger some issues how events are handled in osgjs. The way events works in osgjs is through eventProxies.
eventProxies are responsible to initialize differents device and get events from them. Then we have different adapters to translate events from different device to manipulators.

On osgStats we need to pick a gl quads and scroll it if it hits. 
The problem is that manipulator is connected to StandardMouseKeyboard proxy and a mouseDown will also start to inititate a rotation on the orbit manipulator. Even if it's possible to disable the StandardMouseKeyboard proxy after the mouse down, we still have a side effect at the mouseup event because orbit manipulator has received the first mousedown. So to fix this we inject the mouseup in StandardMouseKeyboard at the end of the scroll to simulate we released the button (remember the proxy was disable).
To have osgStats working as we would like we would need to have eventProxies more general not only designed for manipulator, and we would need to have like some piority how events are consumed.
For example in this context I would like osgStats to consume events first and be able to discard them for others manipulators if needed

feedbacks ?

### DIFF
--- a/sources/osgStats/Stats.js
+++ b/sources/osgStats/Stats.js
@@ -8,11 +8,47 @@ import Program from 'osg/Program';
 import Shader from 'osg/Shader';
 import Transform from 'osg/Transform';
 import Texture from 'osg/Texture';
-import { mat4 } from 'osg/glMatrix';
+import {mat4} from 'osg/glMatrix';
 import BufferStats from 'osgStats/BufferStats';
 import Counter from 'osgStats/Counter';
 import Graph from 'osgStats/Graph';
 import TextGenerator from 'osgStats/TextGenerator';
+
+// Stats usages:
+// url usable in Options
+//
+// to active the stats: ?stats=1
+//
+// to filter content in the stats: statsFilter=cull;myGroup;webgl
+//
+// to change the fontSize: statsFontSize=40
+//
+// You can also change the configuration before running the viewer to adds counters
+// var config = {
+//     values: {
+//         myCounter: {
+//             caption: 'my uber counter',
+//             average: true,
+//             graph: true,
+//             over: 16
+//         },
+//         myCounter2: {
+//             caption: 'a second counter',
+//             graph: true,
+//             below: 16
+//         }
+//         groups: [
+//             {
+//                 name: myGroup,
+//                 caption: 'blah my group',
+//                 values: [ 'myCounter', 'myCounter2' ]
+//             }
+//         ]
+//     };
+//     getViewerStats().addConfig(config)
+//     getViewerStats().setShowFilter(['cull', 'myGroup']) // will display only cull
+//     and myGroup groups
+//     getViewerStats().setFontSize(50)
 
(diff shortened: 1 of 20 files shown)


sources/osgStats/Stats.js
```diff
@@ -206,28 +274,125 @@ utils.createPrototypeObject(Stats, {
...
+    onMouseUp: function(e) {
+        this._onStats = false;
+        this._eventMouse.setEnable(true);
+        this._eventTouch.setEnable(true);
+        this._eventMouse.mouseup(e);
+    },
+
+    computeNearestIntersection: (function() {
+        var coord = vec2.create();
+        var lsi = new LineSegmentIntersector();
```
jtorresfabra: For stats picking should be good to stop at first intersection if we disable intersection in the rest of the graph. With this PR I think scene is being intersected, am I right?

sources/osgStats/Stats.js
```diff
@@ -206,28 +274,125 @@ utils.createPrototypeObject(Stats, {
...
+        var blendFunc = new BlendFunc('SRC_ALPHA', 'ONE_MINUS_SRC_ALPHA');
+        node.getOrCreateStateSet().setAttributeAndModes(blendFunc);
         node.getOrCreateStateSet().setAttributeAndModes(new Depth(Depth.DISABLE));
         node.getOrCreateStateSet().setTextureAttributeAndModes(0, texture);
+
+        var canvas = this._viewer.getGraphicContext().canvas;
+
+        this._canvas = canvas;
+
+        canvas.addEventListener('mousemove', this.onMouseMove.bind(this));
```
jtorresfabra: We could do something to improve this.

jtorresfabra:
> Even if it's possible to disable the StandardMouseKeyboard proxy after the mouse down, we still have a side effect at the mouseup event because orbit manipulator has received the first mousedown

I think enabling/disabling manipulators should be easier than it is now. Something like viewer.setManipulator(0); could do the trick.

cedricpinson:
> If we manage to disable manipulators we wouldn't need to do this
No you will still have the problem because to know that you want to disable the manipulator you have to get the mousedown event, so the manipulator will also receive it.

jtorresfabra:
Yes, but you then do not need to worry about the mouseup. Also it is very useful to code othe kind of tools. For example measuring, editing vertices etc. So once you detect you picked the stats quad you can disable the manipulator and that's it (ideally).

cedricpinson:
I agree it could be good to disable manipulator easier, but even with that it would not work:
https://github.com/cedricpinson/osgjs/blob/improves-stats/sources/osgGA/OrbitManipulatorStandardMouseKeyboardController.js#L59
on mousedown you change the state of the manipulator so you still have the problem.

jtorresfabra:
Yep, in my opinion disable manipulator should resolve this issue, fixing manipulator state.

PR #924, opened 2017-12-08
Composer fix reusable

stephomi:
reusable was forced to true

### DIFF
--- a/sources/osgDB/readerParser.js
+++ b/sources/osgDB/readerParser.js
@@ -47,7 +47,7 @@ ReaderParser.registry = function() {
 
 ReaderParser.parseSceneGraph = function(node, options) {
     if (node.Version !== undefined && node.Version > 0) {
-        utils.time('osgjs.metric:ReaderParser.parseSceneGraph');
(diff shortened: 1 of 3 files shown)


sources/osgDB/readerParser.js
```diff
@@ -47,8 +47,6 @@ ReaderParser.registry = function() {
 
 ReaderParser.parseSceneGraph = function(node, options) {
     if (node.Version !== undefined && node.Version > 0) {
-        utils.time('osgjs.metric:ReaderParser.parseSceneGraph');
```
cedricpinson: dont remove them, instead replace with 
utils.time('osgjs.metric:ReaderParser.parseSceneGraph', notify.INFO)

stephomi: done

cedricpinson:
@stephomi maybe this line should be fixed too
https://github.com/cedricpinson/osgjs/blob/2ba20789bd2c8b2cad99843d80213d29997be015/sources/osgUtil/ComposerPostProcess.js#L835
I think it's better to have true or false in this variable but not undefined

PR #925, opened 2017-12-13
Fixes touch event handling in Stats

Nehon:


sources/osgStats/Stats.js
```diff
@@ -55,6 +55,12 @@ import BoundingBox from 'osg/BoundingBox';
 //     getViewerStats().setFontSize(50)
 
 var getCanvasCoord = function(vec, e) {
+    if (e.touches && e.touches.length) {
+        var touch = e.touches[0];
+        vec[0] = touch.pageX;
+        vec[1] = touch.pageY;
+        return;
```
cedricpinson: an if else would be more appropriate in this case

PR #929, opened 2017-12-20
Restores the behaviour of First Person Manipulator.

rtortajada:
It was changed when the line `this._buttonup = false;` was added. Before that, this._buttonup was set to true in the call to `this.releaseButton()`. Have you changed this intentionally?


jtorresfabra:
@rtortajada I would say it was intentionally. Maybe you need to call to setEnable(true)?
Anyway @cedricpinson knows better.

rtortajada:
I think it is not about the enable function. Before that when the FirstPersonManipulator was started you needed to drag the mouse to move the camera. But, now the camera is moved when there is a mousemove event, it is not necessary a drag. Also, if you do a mouseout or a mouseclick, the functionality is restored as before. (You can check it in the sound-3d example)
So, I think it is a problem in the init function about the buttonup var.

jtorresfabra:
Indeed, seems a bug. cc: @cedricpinson

Issue #877, opened 2017-10-03
Polytope Intersector fail


jucamal:
Polytope Intersector fails when the polygon vertices are outside the polytope.

To test this issue, modify the createTriangle method of the test in  tests\osgUtil\PolytopeIntersector.js by the next one.

```
var createTriangle = function() {
        var g = new Geometry();
        var vertexes = new MACROUTILS.Float32Array(9);
        vertexes[0] = -400.0;
        vertexes[1] = 600.0;
        vertexes[2] = 0.0;

        vertexes[3] = 0.0;
        vertexes[4] = -300.0;
        vertexes[5] = 0.0;

        vertexes[6] = 400.0;
        vertexes[7] = 600.0;
        vertexes[8] = 0.0;

        var normal = new MACROUTILS.Float32Array(9);
        normal[0] = 0;
        normal[1] = 0;
        normal[2] = 1;

        normal[3] = 0;
        normal[4] = 0;
        normal[5] = 1;

        normal[6] = 0;
        normal[7] = 0;
        normal[8] = 1;

        g.getAttributes().Vertex = new BufferArray(BufferArray.ARRAY_BUFFER, vertexes, 3);
        g.getAttributes().Normal = new BufferArray(BufferArray.ARRAY_BUFFER, normal, 3);

        var primitive = new DrawArrays(primitiveSet.TRIANGLES, 0, vertexes.length / 3);
        g.getPrimitives().push(primitive);
        return g;
    };
```

jtorresfabra:
Hey @jucamal,
Can you test if https://github.com/cedricpinson/osgjs/pull/930 fixes your issues?
Thanks

Issue #858, opened 2017-08-30
Polytope intersection sometimes miss with triangles


stephomi:
I had the issue on a low poly mesh (700 verts)

cedricpinson:
I guess it's fixed now ?

jtorresfabra:
No it is not, in fact it is probably related to #877. I can confirm there is an error.

PR #930, opened 2017-12-29
Fixes polytope intersection when polytope inside triangle

jtorresfabra:


jtorresfabra:
Should fix https://github.com/cedricpinson/osgjs/issues/877 and https://github.com/cedricpinson/osgjs/issues/858

PR #932, opened 2018-01-03
Renames shadowEnable to isContinuousUpdate

Kuranes:
shadow::enable or not => was meant for static or dynamic
Best to rename it shadow::setContinousUpdate
(the "real" disabling shadow is done removing the shadowedscene node)

### DIFF
--- a/sources/osgShadow/ShadowMapAtlas.js
+++ b/sources/osgShadow/ShadowMapAtlas.js
@@ -379,32 +379,32 @@ utils.createPrototypeObject(
 
             for (var i = 0, l = this._shadowMaps.length; i < l; i++) {
                 var shadowMap = this._shadowMaps[i];
-                if (shadowMap.isEnabled() && !shadowMap.isFilledOnce()) {
+                if (shadowMap.isContinuousUpdate() || shadowMap.needRedraw()) {
                     shadowMap.cullShadowCasting(cullVisitor, bbox);
                 }
             }
         },
 
-        isEnabled: function() {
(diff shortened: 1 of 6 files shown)


sources/osgShadow/ShadowMapAtlas.js
```diff
@@ -379,32 +379,32 @@ utils.createPrototypeObject(
...
                 var shadowMap = this._shadowMaps[i];
-                if (shadowMap.isEnabled() && !shadowMap.isFilledOnce()) {
+                if (shadowMap.isContinuousUpdate() || shadowMap.needRedraw()) {
                     shadowMap.cullShadowCasting(cullVisitor, bbox);
                 }
             }
         },
 
-        isEnabled: function() {
+        isContinuousUpdate: function() {
```
jtorresfabra: Probably more correct getContinuousUpdate.

Kuranes: Seems to me osgjs code is more in favor of is... (isDirty, isEnabled, etc)
Can change, but better do it consistently ? Whole bag of cat opening there :)

jtorresfabra: @Kuranes I mean isContinuousUpdate seems weird in english, but I'm no native.

Kuranes: ha well, just applied the pattern consistently instead of natively speaker
"it's a getter?" => "is" + "var name" + ()

PR #935, opened 2018-01-05
Adds a method to RenderBin and RenderStage to clean their PooledResource objects

davehg:
When the canvas is removed the PooledResource objects in RenderBin and RenderStage are not cleaned and they keep references to BufferArrays so the memory is not freed. 

Adding this two methods allows to free that memory.


davehg:
I have added a dispose method in the viewer as @jtorresfabra requested to clean pooled resources.

PR #943, opened 2018-01-11
Filters false positive gpu extension

Kuranes:
Some mobile reuse same driver GPUs with different capabilities
now, we double check if glsl compiler agree

### DIFF
--- a/sources/osg/WebGLCaps.js
+++ b/sources/osg/WebGLCaps.js
@@ -390,6 +390,35 @@ WebGLCaps.prototype = {
         return this._webGLExtensions;
     },
 
+    _checkShader: function(gl, t, text) {
+        var type = t;
+        if (typeof type === 'string') type = gl[t];
+        var shader = gl.createShader(type);
+        gl.shaderSource(shader, text);
+        gl.compileShader(shader);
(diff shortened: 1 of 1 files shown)


sources/osg/WebGLCaps.js
```diff
@@ -390,6 +390,42 @@ WebGLCaps.prototype = {
...
+    _checkShader: function(gl, t, text) {
+        var type = t;
+        if (typeof type === 'string') type = gl[t];
+        var shader = gl.createShader(type);
+
+        gl.shaderSource(shader, text);
+        gl.compileShader(shader);
+
+        if (!gl.getShaderParameter(shader, gl.COMPILE_STATUS) && !gl.isContextLost()) {
+            console.warn("Could't compile shader " + text);
```
jtorresfabra: notify

sources/osg/WebGLCaps.js
```diff
@@ -390,6 +390,42 @@ WebGLCaps.prototype = {
...
+            console.warn(gl.getShaderInfoLog(shader));
+            console.warn('You may want to consider a browser that fully supports WebGL.');
```
jtorresfabra: Maybe the browser is not the culprit

Kuranes: You're right, in fact logs wouldn't help at all in this case.
I will remove them

PR #945, opened 2018-01-18
Fixes exception on old browsers

Kuranes:
Old browser without map, set or uint8, we don't even try, so no need
to do exception.
(osgjs dev users has to filter on his side usage of osgjs)

### DIFF
--- a/sources/osg/polyfill.js
+++ b/sources/osg/polyfill.js
@@ -1,58 +1,58 @@
-// This file contains needed polyfills mainly for IE11
-if (!Math.sign) {
-    Math.sign = function(a) {
-        return a > 0.0 ? 1.0 : a < 0.0 ? -1.0 : 0.0;
-    };
-}
-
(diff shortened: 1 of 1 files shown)


sources/osg/polyfill.js
```diff
@@ -1,57 +1,57 @@
...
-        return a > 0.0 ? 1.0 : a < 0.0 ? -1.0 : 0.0;
-    };
-}
+(function() {
+    // prevent crash on old browser
+    if (!window.Set || !window.Map || !window.Uint8ClampedArray) return;
+    // IE11 does not support Set with constructing arguments. May 2017.
+    var setTest = new window.Set(['test']);
+    var hasConstructorParameterSupport = setTest.has('test');
+    if (hasConstructorParameterSupport) return;
```
jtorresfabra: Not sure, could this be a too early return?

PR #952, opened 2018-02-14
Makes the webvr device polling on demand

Nehon:


sources/osgViewer/input/source/InputSourceWebVR.js
```diff
@@ -20,9 +25,10 @@ var InputSourceWebVR = function() {
     }
     this._nbCallbacks = 0;
     // 3 seconds default poll interval
-    this._pollInterval = 3000;
+    this._pollInterval =
+        options && options.pollInterval !== undefined ? options.pollInterval : 3000;
```
Kuranes: 3000 is quite big ?
maybe a comment explaining the choice

sources/osgViewer/input/source/InputSourceWebVR.js
```diff
@@ -62,57 +68,78 @@ utils.createPrototypeObject(
...
+            }
+        },
+
+        pollHeadset: function() {
+            if (!this._nbCallbacks) {
+                //don't poll if there is no callback registered.
+                return;
+            }
+
             if (!navigator.getVRDisplays) {
```
Kuranes: could be done in the _schedulepolling once and for al ?

sources/osgViewer/input/source/InputSourceWebVR.js
```diff
@@ -62,57 +68,78 @@ utils.createPrototypeObject(
...
-                                var callbacks = self._callbacks['vrdisplaydisconnected'];
-                                if (callbacks) {
-                                    for (i = 0; i < callbacks.length; i++) {
-                                        callback = callbacks[i];
-                                        callback(event);
-                                    }
+            var self = this;
+            navigator.getVRDisplays().then(
+                function(displays) {
+                    if (displays.length > 0) {
```
Kuranes: could be a early return
```if (displays.length=== 0) { 
this.triggerNotFoundEvent();
return;
}```

File: sources/osgViewer/input/source/InputSourceWebVR.js [hunk 4]
```diff
@@ -62,57 +68,78 @@ utils.createPrototypeObject(
...
+                        if (self._hmd !== displays[0]) {
```
Kuranes: mmm perhaps time for a loop over display ?
or an early return :) 
```
if (self._hmd === displays[0]) return;
```

sources/osgViewer/input/source/InputSourceWebVR.js
```diff
@@ -62,57 +68,78 @@ utils.createPrototypeObject(
...
+                    } else {
+                        this.triggerNotFoundEvent();
+                    }
+                }.bind(this)
             );
         },
 
-        setPollInterval: function (interval) {
+        triggerNotFoundEvent: function() {
+            if (this._pollInterval <= 0) {
```
Kuranes: this one is not clear: if _pollInterval is negative, how can we get here ?

otw early return too :)

sources/osgViewer/input/source/InputSourceWebVR.js
```diff
@@ -62,57 +68,78 @@ utils.createPrototypeObject(
...
+                // in case of manual polling we trigger an event when no display was found
+                var callbacks = this._callbacks['vrdisplaynotfound'];
```
Kuranes: could be a func, as it's same exact code as line 112

sources/osgViewer/input/source/InputSourceWebVR.js
```diff
@@ -62,57 +68,78 @@ utils.createPrototypeObject(
...
+                if (callbacks) {
+                    event = this._events['vrdisplaynotfound'];
+                    for (var i = 0; i < callbacks.length; i++) {
+                        var callback = callbacks[i];
+                        callback(event);
+                    }
+                }
```
Kuranes: 130 to 138 could be a function as it's exact same code a 111-117 ?

sources/osgViewer/input/source/InputSourceWebVR.js
Nehon: 3 seconds seemed an ok value. It's the maximum time a user will have to wait when pluging his headset to be recognized.

Kuranes: ha ok, it's the time between putting an headset and getting result ?

When does that arrives ?

once user clicks on the "vr icon" we poll each 3 secs ? or it's at another point ?
(or maybe it's always running ?)

sources/osgViewer/input/source/InputSourceWebVR.js
Nehon: Ok, I changed it.
Now we start polling whenever a callback is added and stop hen there are no more callbacks registered

sources/osgViewer/input/source/InputSourceWebVR.js
Nehon: When poll interval is negative, this means that we don't poll automatically but that the user may poll manually using the pollHeadsetMethod.
When this method is called manually, we need to notify the user that no headset has been found when relevant.
However on automatic polling, we don't want to trigger it. Else if would be triggered every pollInterval when no device is plugged.

Kuranes: ha, didn't see any pollHeadset manual call anywhere, that's why I'm surprised

PR #957, opened 2018-03-06
Fixes crash with IE 11 on Inputmanager Initialization

Nehon:
The joy of IE11.

### DIFF
--- a/sources/osg/polyfill.js
+++ b/sources/osg/polyfill.js
@@ -63,4 +63,20 @@
         }
         return set;
     };
+
+    // IE11 doesn't support Event() constructor
+    // see https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent
+    // Here we define window.Event instead of window.CustomEvent so that it can be instantiated in the InputManager
+    if (typeof window.Event === 'object') {
+        function CustomEvent(event, params) {
+            params = params || { bubbles: false, cancelable: false, detail: undefined };
+            var evt = document.createEvent('CustomEvent');
+            evt.initCustomEvent(event, params.bubbles, params.cancelable, params.detail);
+            return evt;
+        }
+
+        CustomEvent.prototype = window.Event.prototype;
+
+        window.Event = CustomEvent;
+    }
 })();
--- a/sources/osgViewer/input/source/InputSourceKeyboard.js
+++ b/sources/osgViewer/input/source/InputSourceKeyboard.js
@@ -43,6 +43,17 @@ utils.createPrototypeObject(
             if (!parsedEvent.action) {
                 return true;
             }
+            if (!nativeEvent.key) {
+                // Sentry reports errors with the nativeEvent key undefined.
+                // We can't reproduce the issue so this code is supposed to give us more information when that happens.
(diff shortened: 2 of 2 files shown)


sources/osg/polyfill.js
```diff
@@ -63,4 +63,19 @@
...
+        function CustomEvent(event, params) {
+            params = params || { bubbles: false, cancelable: false, detail: undefined };
+            var evt = document.createEvent('CustomEvent');
+            evt.initCustomEvent(event, params.bubbles, params.cancelable, params.detail);
+            return evt;
+        }
+
+        CustomEvent.prototype = window.Event.prototype;
+
+        window.Event = CustomEvent;
```
Kuranes: Better not change Event browser native ?
I thought we would try
- no override of Event 
- usage of "new CustomEvent" instead of "new Event" in rest of the code (for all browser, not just IE)

jtorresfabra: Mostly agree.

Nehon: I don't like the idea of changing working code for all browsers because of a workaround for an old messed up browser.
With this change the input manager code stays the same, only IE11 is affected.

stephomi: I tends to agree with nehon on this point, I find it best if ugly polyfill code is contained in a single file (as much as possible).

Kuranes: rationale is that after reading customEvent is not much different than Event, just better (you can add data, not just an event name)
All code can do "new CustomEvent" without spreading polyfill or doing bad code for others

sources/osgViewer/input/source/InputSourceKeyboard.js
```diff
@@ -43,6 +43,14 @@ utils.createPrototypeObject(
             if (!parsedEvent.action) {
                 return true;
             }
+            if (!nativeEvent.key) {
+                // Sentry reports errors with the nativeEvent key undefined.
+                // We can't reproduce the issue so this code is supposed to give us more information when that happens.
+                var error = new Error('Invalid keyboard event', 'InputSourceKeyboard.js', 47);
```
Kuranes: I would add a JSON.stringify(nativeEvent) concatened with the "inputsource.js"

Kuranes: ```var error = new Error('Invalid keyboard event', 'InputSourceKeyboard.js: ' + JSON.stringify(nativeEvent), 47);```

Nehon: I added it on the cause attribute thinking we could see it on sentry.
But if you prefer like that I'm ok

Kuranes: sentry gets only strings so you need to stringify the object NativeEvent to be able to "debug" from sentry. (you don't get "variable stack", only "function stack" traces in sentry)

sources/osg/polyfill.js
cedricpinson: agree with @Nehon it will override only on ie11 so good for me. If later we really need CustomEvent then we will be able to change for a good reason

PR #968, opened 2018-04-02
Adds support for ANGLE_InstancedArrays extension

jtorresfabra:
Updated the previows PR, but it was better to just do a new one instead of rebasing.. 
Also executed a grunt prettier on the codebase


cedricpinson:
could you splits commits about prettier and the real changes plz ?

sources/osg/Geometry.js
```diff
@@ -120,6 +122,15 @@ utils.createPrototypeNode(
             }
         },
 
+        setAttribArrayDivisor: function(attributeName, divisor) {
+            if (this._attributes[attributeName] === undefined) {
+                notify.error('Attribute' + attributeName + 'does not exist.');
+                return;
+            }
+            if (this._instancedArrayMap === undefined) this._instancedArrayMap = new window.Map();
```
jtorresfabra: Maybe better to use a {}?

jtorresfabra:
@cedricpinson done

sources/osgDB/readerParser.js
```diff
@@ -204,8 +204,8 @@ ReaderParser.parseSceneGraphDeprecated = function(node) {
                 geom.getPrimitiveSetList().push(new DrawElements(mode, array));
             } else {
                 mode = primitiveSet[mode];
-                var first = primitives[p].first;
-                var count = primitives[p].count;
+                var first = primitives[p].first || primitives[p]._first;
+                var count = primitives[p].count || primitives[p]._count;
```
Kuranes: same ?

sources/osg/Node.js
```diff
@@ -36,6 +36,8 @@ var Node = function() {
 
     // it's a tmp object for internal use, do not use
     this._tmpBox = new BoundingBox();
+    this._computeBoundingBoxCallback = undefined;
+    this._computeBoundingSphereCallback = undefined;
```
Kuranes: _computeBoundingSphereCallback never used ?
need to remove it

Kuranes (changes requested):
- inconsistency between primitives member names (with or without '_') and accessor usage (getIndices())
- doesn't work in webgl2
- Why we do not use the current mechanism in webglcaps to create a backport of webgl2 "gl.DrawElementInstanced, gl.DrawArrayInstanced" to webgl1 ? ( see https://github.com/cedricpinson/osgjs/blob/master/sources/osg/WebGLCaps.js#L365-L383 && https://developer.mozilla.org/en-US/docs/Web/API/ANGLE_instanced_arrays for webgl2 names)

Would give Clearer code and would work on webgl1/2 (and allow to remove the "ANGLE" text from everywhere for just "instanced")

PR #973, opened 2018-04-16
Dod experiment localize matrix data

Kuranes:
Todo:  need "better "pre-estimation" for a good pre-alloc value


Kuranes:
https://dev.skfb-local.com/osgjs/examples/performance/?&items=4&deep=3&complexStateSet=true&numTextures=20
Dod: 37ms
Nomal: 45ms

Kuranes:
updated PR with resize and Float64 
=> will have to re-run the test: Was using float32 whereas dod and master is using float64

Kuranes:
Updated with latest and 64bits:
-> desktop hard to see any ms Difference. 
-> mobile from 230 to 220ms.

In Chrome javascript Profiler: the mat4.multiply goes from 6% to 4%.  From 3rd place to 7th

cedricpinson:
I close it for now

PR #984, opened 2018-07-10
Checks with shader that texturecubelodext support is real

Kuranes:
PowerVR gpu compiler has some version where it's just not there...

### DIFF
--- a/sources/osg/WebGLCaps.js
+++ b/sources/osg/WebGLCaps.js
@@ -129,7 +129,7 @@ WebGLCaps.prototype = {
         // order is important
         // to allow webgl extensions filtering
         this.initPlatformSupport();
-        this.initBugDB();
+        this.initBugDB(gl);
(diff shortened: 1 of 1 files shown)


sources/osg/WebGLCaps.js
```diff
@@ -177,6 +177,47 @@ WebGLCaps.prototype = {
...
+    checkShader: function(gl, t, text) {
+        var type = t;
+        if (typeof type === 'string') type = gl[t];
+        var shader = gl.createShader(type);
+
+        gl.shaderSource(shader, text);
+        gl.compileShader(shader);
+
+        if (!gl.getShaderParameter(shader, gl.COMPILE_STATUS) && !gl.isContextLost()) {
+            console.warn(gl.getShaderInfoLog(shader));
```
cedricpinson: notify ?

PR #990, opened 2018-08-02
Fixes automipmapping video

Kuranes:
video with mipmap filter was not getting the mipmaps  updated.

### DIFF
--- a/sources/osg/Texture.js
+++ b/sources/osg/Texture.js
@@ -809,8 +809,9 @@ utils.createPrototypeStateAttribute(
                     // image update like video
                     if (this._image !== undefined && this._image.isDirty()) {
                         this.applyImage(gl);
+                        this.dirtyMipmap();
                     }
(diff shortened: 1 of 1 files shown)


sources/osg/Texture.js
```diff
@@ -809,6 +809,11 @@ utils.createPrototypeStateAttribute(
                     // image update like video
                     if (this._image !== undefined && this._image.isDirty()) {
                         this.applyImage(gl);
+                        // if it's a video
+                        // and filter are set to mipmap
```
cedricpinson: The comment is not great, in a general case, it's more.
if the image is dirty we need to update the texture and mipmap if a mipmap filter is used

Kuranes: Indeed. 
either I remove the comment or I replace "video" by "animated image" ?

cedricpinson: you can remove the comment I think

PR #994, opened 2018-08-22
Changes the way the yaw limit is checked

Nehon:


cedricpinson:
is there a way to disable constraint when set to true ?

